PR #9, opened 2023-11-27
4a add extern test for properties

evomimic:
Added testing code that exercises the holon_api's add_property_value function and compares results. Tests pass, but get_all_holons issue previously reported has not been resolved, so that test code is still commented out.


evomimic:
Self-review

PR #10, opened 2023-11-30
4 introduce and test holonnode

evomimic:
Simple test of create holon, with fetch, get all holons and delete holon works.


nphias (approved):
removal of a curly brace ... 
CI failing from duplicate dependency in root toml:
holochain_integrity_types = "0.2"

PR #13, opened 2023-12-14
5 introduce descriptors bootstrap crate

evomimic:
Initial capacity to bootstrap descriptor definitions as part of sweetest runs is working. Only a small set of descriptor types are actually loaded at this point, but the foundation is here.


nphias:
works for me

PR #18, opened 2024-01-18
16 apply newtype pattern for map value system

dauphin3:
convert type aliases into tuple struct wrapper, apply types to fns


evomimic (approved):
All the changes look good!

PR #19, opened 2024-01-19
17 revert stagedpropertymap to propertymap with holons coordinator zome

dauphin3:
removed Option<BaseValue> and associated iter filters


evomimic (approved):
Looks good

PR #25, opened 2024-01-25
23 implement holonreference access layer

dauphin3:
impl HolonReferenceFns trait for HolonReference and LocalHolonReference


evomimic (approved):
Looks good!

Issue #42, opened 2024-02-22
Introduce StagedReference (replacing HolonReference)


evomimic:
This issue refactors HolonReferences into StagedReferences as part of the overall HolonReference refactoring.

- eliminated the distinction between `LocalHolonReference` and `ExternalHolonReference`
- introduced a different split between immutable holon references (renamed `SmartReference` and introduced by Issue #41) and mutable holon references named `StagedReference` being introduced in this issue.

The CommitManager owns the actual Holons referenced by StagedReference.

- [x] Define StagedReference as a variant of HolonReference enum
- [x] Implement add_related_holons for StagedReference

evomimic:
I'll need to rethink RelationshipTarget now...

evomimic:
merged/closed

Issue #44, opened 2024-02-25
Refactor RelationshipTarget


evomimic:
For now... let's re-factor the `RelationshipTarget` enum, by unifying the representation of relationships without regard to cardinality. In other words, instead of having `One`, `ZeroOrOne` and `Many` variants with different representations for each, all relationships will be represented as collections. 

At the same time, we need to introduce the distinction between Staged (i.e., mutable) and Smart (i.e., immutable) Collections. 

```
pub type RelationshipMap = BTreeMap<RelationshipName, RelationshipTarget>;
pub struct RelationshipName(pub MapString);
pub struct RelationshipTarget {
    pub editable: Option<StagedCollection>, // Mutable collection
    pub cursors: Vec<SmartCollection>, // a set of immutable, access path specific collections
}
```

- [x] Refactor RelationshipTarget definition and affected function implementations
- [x] Confirm app compiles successfully

evomimic:
Resolved and merged into branch #38 by PR#48

Issue #43, opened 2024-02-23
Introduce Request and Response Objects to Pass Staged Holons to/from Client and Guest


evomimic:
If we are going to allow client-side determination of commit boundaries, then the state of staged holons (i.e., holons that have been modified or staged for creation) needs to be kept somewhere. Once we have the ephemeral store, we can stage them there. But until then, the StagingArea will have to be sent back and forth from client to guest.

The easiest way to do this is to introduce Request and Response objects (which we will need for Dances anyway).

PR #48, opened 2024-03-09
implemented issue 42 -- StagedReference -- compiles still has test fa…

evomimic:
Resolves issue #42 and issue #44. The app and sweetest's compile, but sweetest still fails due to issue #43.


dauphin3 (commented):
smart_reference.rs lines 50-60 ensure_rc() could return a bool or the object itself

the if let in line 52 seems unnecessary, as once ensured it shouldn't be None -- and if it is for some reason.. that is a problem that should be checked

dauphin3 (approved):

PR #50, opened 2024-03-10
6 introduce transient holoncache and cachemanager

dauphin3:
```
#[derive(Debug, Clone)]
pub struct HolonCacheManager {
    pub local_cache: HolonCache,
    // pub external_caches: HashMap<HolonSpaceId, HolonCache>,
}

#[derive(Debug, Clone)]
pub struct HolonCache {
    // fetcher: Dance, // the Dance that can be used to fetch records from the backing store for this cache
    pub map: HashMap<HolonId, CachedHolon>,
}

#[derive(Debug, Clone)]
pub struct CachedHolon {
    pub id: HolonId,
    pub til: Timestamp,
    pub holon: Holon,
}

impl HolonCacheManager {
    pub fn new() -> HolonCacheManager {
        HolonCacheManager {
            local_cache: HolonCache {
                map: HashMap::new(),
            },
        }
    }

    pub fn init_cache() -> Cache<HolonId, CachedHolon> {
        Cache::new(99)
    }

    pub fn update_cache(&mut self, holon_id: HolonId, cached_holon: CachedHolon) {
        self.local_cache.map.insert(holon_id, cached_holon);
    }

    pub fn get_rc_holon(
        context: &HolonsContext,
        holon_space_id: Option<HolonId>,
        holon_id: HolonId,
    ) -> Result<Rc<RefCell<Holon>>, HolonError> {
        if let Some(id) = holon_space_id {
            return Holon::fetch_holon(context, id);
        } else {
            let map = context.clone().cache_manager.into_inner().local_cache.map;
            let cached_holon = map.get(&holon_id);
            if let Some(holon) = cached_holon {
                return Ok(Rc::new(RefCell::new(holon.holon.clone())));
            }
            Err(HolonError::HolonNotFound(
                "Invalid HolonId, Holon does not exist".to_string(),
            ))
        }
    }
}
```


dauphin3:
decided to have get_rc_holon return  Result<Rc<RefCell<Holon>>, HolonError> since that's what fetch_holon returns

dauphin3:
for discussion: if holon_space_id is None and holon is not found in local cache (ie localcache.map.get(holon_id) returns None) currently it throws an error since holon should probably exist in either.. although this may be incorrect...

HolonId is a wrapper for ActionHash, so what are the scenarios for holon_space_id.is_none() and the Holon somehow has a commited ActionHash? does the HolonId type need to be extended?

My understanding is that a 'staged' (uncommitted) Holon has key but not an ActionHash. We don't want a HolonReference to have to know the difference, but I'm wondering how can a commited Holon that has an ActionHash not have an associated HolonSpace since I thought every Holon is owned by exactly one HolonSpace

evomimic:
Hey @dauphin3 ... looking forward to discussing this tomorrow.

We should change the branch we're merging into to match the branch you forked branch #6 from. What branch did you fork from?

evomimic:
Code looks good, approving this merge for now... still need to get sweetest working...

PR #59, opened 2024-03-25
58 stub out commit flow

dauphin3:
left commented code for taking mutable reference in commit impls

open question:  should holons field in StagedCollection, which is Vec<HolonReference, be emptied after commit ?


evomimic (approved):
Looks good... approving

PR #86, opened 2024-05-21
Establish a new baseline in main

evomimic:
Before beginning the next round of development, I wanted to clean up old branches and get back to a simpler branching structure.


evomimic:
too many conflicts... I'm going to try a different approach

Issue #79, opened 2024-05-14
Implement AbandonStagedChanges Dance


evomimic:
## Summary
This enhancement adds the _**abandon_staged_changes**_ dance to the Uniform API. This dance marks a staged_holon as `Abandoned` . Staged holons marked `Abandoned` are simply ignored during commit processing -- they are not saved to the persistent store. Once a staged_holon has been abandoned, further attempts to access the holon will return a (newly introduced) `NotAccessible` error. 

## Current State:

There is no native functionality that supports this function. 

However, some existing native functionality needs to modified:

- we cannot simply remove an abandoned holon from the CommitManager's staged_holons vector, because that could affect the position of other holons in the vector and StagedIndex depends upon vector position.
- the current definition of CommitResponse assumes that a holon commit either fails or results in a saved holon. On an 'Ok' result is returned the holon is added to the saved_holons vector. To make the semantics clearer, saves should be distinguished from abandons, even though both will return an Ok Result.
- we need to block attempts to access _abandoned_ staged holons. One design option for this is to have each function assume responsibility for including a  _access guard_. Another option is to move accessibility checks to a separate layer (perhaps as part of a "validation check" that is invoked by the dancer. This provides a more modular design and decreases the possibility of a developer "forgetting" to add a guard to a function. For now, this enhancement request is proposing the first option. We can move to the second option in the future as part of a _dance_validation step_ (perhaps following the _Decorator_ pattern).

### Enhance Native Holons Functionality

_**In the `holon_error.rs` file within the Holons zome:**_

- [x] Add a new `NotAccessible` error to the `HolonError` enum to  report access violations 

```
    #[error("{0} access not allowed while holon is in {1} state")]
    NotAccessible(String, String),

```
_**In the `holon.rs` file within the Holons zome:**_
- [x] Add an `Abandoned` variant to `HolonState` and extend the implementation of the fmt function to cover this variant
- [x] Add an `abandon_staged_changes` method that sets holon's state to `Abandoned`
- [x] Enhance the commit method to support `Abandoned` semantics. 
- [x] Add an `is_accessible(access_type: AccessType)->Result<_,HolonError>` helper function, where 
```
pub enum AccessType {
    Read,
    Write,
}
```

This function should return `Ok` if the requested `access_type` is allowed given the current state of the holon. Otherwise, return a `NotAccessible` error.

     - If requested access_type is `Read`, return true if and only if the holon's state is: `New`, `Fetched`, `Changed`, or `Saved`
[...]

PR #92, opened 2024-05-30
79 implement abandonstagedchanges dance

dauphin3:


crates/coordinator/holons/src/commit_manager.rs
evomimic, line 1: I noticed several comments/questions in the code related to StagedIndex and the `saved_holons` and `abandoned_holons` vectors. The latter vectors are (only) used to convey the results of a single commit() call on the CommitManager back to the caller so the caller can see what happened. They bear no relation  to and, deliberately, have no dependency on the vector of staged_holons that is maintained by the CommitManager (and shipped to client via StagingArea).

evomimic, line 1: I don't think the duplicate checking code on the abandoned_holons and saved_holons vector is necessary, nor can I think of any way to test it. These vectors both start out empty and the method does a single iteration over staged_holons.

crates/coordinator/holons/src/holon_error.rs
evomimic, line 1: What's a GuardError and why do we need it? Guard is a generic term. Guard functions could be implemented to guard against a variety of conditions. If they encounter errors, they should return an indication of the condition that failed.

In this case, we've introduced guard functions to check if the state of a holon supports access in the form requested (read or write). If not, a NotAccessible error is returned. I'm unclear on why we also need a GuardError.

crates/coordinator/dances/tests/shared_test/dance_fixtures.rs
evomimic, line 1: - The simple_abandon_staged_changes_fixture() function is not adding a "add_stage_holon_step" to the test case for person1.
- RelationshipNames should always be all UPPER_CASE -- change "Authored by" to "AUTHORED_BY"
 - The following code seems to be confusing the results of adding a step to the test case with the result of executing that step during test execution:
```
let expected_error = test_case.add_with_properties_step(1, properties);
    if expected_error.is_ok() {
        return Err(HolonError::GuardError("with_property_value".to_string()));
    }
```
If the add_with_properties_step returns an error... just return the error via? Don't map it to a GuardError.
- But what we probably need to do is extend our test_step functions so that they take an additional expected_response parameter.

crates/coordinator/dances/tests/shared_test/test_abandon_staged_changes.rs
evomimic, line 1: I'm seeing now that Issue #79 was ambiguous. In the current implementation, the balance of logic between the test case definition in the fixture and the execute_abandon_staged_changes function  is wrong. Both need  a pretty significant rewrite.

crates/coordinator/dances/tests/dance_tests.rs
evomimic, line 1: The ONLY changes needed to dance_tests are to add the [case::simple_abandon_staged_changes_test(simple_abandon_staged_changes_fixture())] before the rstest_dance_tests function and add the DanceTestStep::AbandonStagedChanges variant to the match step.

evomimic:
I've completed the rewrite and am running tests... ran into an interesting problem that will require more thought:
H1 has AUTHORED_BY relationship to H2 and H3.
when H2 was abandoned, it was not removed from H1's relationship_map. (There is no way for H2 to know that it is referenced in some other staged_holon's relationship_map)

So on the second commit pass (committing relationships), H1 gets an error trying to commit its relationship to H2.

evomimic:
In response to problem noted above, I re-designed the commit processing for relationships... to skip trying to create smartlinks for target references for which a HolonId is not available (e.g., because a StagedReference was `Abandoned` rather than `Saved`). This involved replacing HolonReference's  `commit_smartlink` method with a `get_id` method and then having StagedCollection's  `add_smartlinks_for_collection` method iterate through its vector of HolonReferences and create a smartlink to it ONLY if it has a holon_id.

crates/coordinator/dances/tests/shared_test/test_abandon_staged_changes.rs
evomimic: Complete rewrite

Issue #88, opened 2024-05-23
Implement get_staged_holon_by key and get_holon_by key.


evomimic:
This enhancement adds`get_holon_by_key` and `get_staged_holon_by_key`dances to the MAP's dance card. This will require some enhancements to native holons functionality. 

DESIGN is TBD.

The SmartCollection already includes a `keyed_index`, but doesn't offer a _get_by_key_ method that leverages that index. The CommitManager already offers a get_holon_by_key method for retrieving staged holons.

### Enhance Native Collections functionality

_**In the `smart_collection.rs` file within the Holons zome:**_

- [ ] Implement `SmartCollection::get_holon_by_key(&self, key: MapString)->SmartReference` method.

### Dance Enhancements

_**In the holon_dance_adapter.rs file:**_

- [ ] Define and implement a new `get_staged_holon_by_key` dance adapter function as a Standalone dance with key:MapString passed via the RequestBody and Index returned in the ResponseBode. It should  delegate the call to the CommitManager's existing `get_holon_by_key` method. 
- [ ] Implement `build_get_holon_by_key_dance_request`  as a Standalone request.
- [ ] Define and implement a new `get_holon_by_key` dance adapter function as a Standalone dance with key:MapString passed via the RequestBody and a Holon returned in the ResponseBody. It should  delegate the call to the CommitManager's existing `get_holon_by_key` method. 
- [ ] Implement `build_get_holon_by_key_dance_request`  as a Standalone request.

```
/// *DanceRequest:*
/// - dance_name: "abandon_staged_changes"
/// - dance_type: Command(StagedIndex) -- references the staged holon whose changes are being abandoned
/// - request_body: None
///   
///
/// *ResponseBody:*
/// - an Index into staged_holons that references the updated holon.
///
```

_**In the dancer.rs file:**_
- [ ] Add "abandon_staged_changes" dance to the Dancer's dispatch table
- [ ] Add error message for `NotAccessible` to the `process_dispatch_result` function. 

_**In the dance_response.rs file:**_

- [ ] Map the `NotAccessible` HolonError to `BadRequest` to the `From<HolonError> for ResponseStatusCode`  implementation

### Testing Enhancements

_**In the test_data_types.rs file:**_
- [ ] Add "test_abandon_staged_changes" test step executor
- [ ] Add `AbandonStagedChanges` as a variant in the `DanceTestStep` enum definition
- [ ] add its fmt::Display implementation
[...]

PR #98, opened 2024-06-05
89 unify portablereference into holonreference

dauphin3:
it has been determined that keys cannot be fetched from holon nodes without descriptors and therefore we are continuing forward without including keys in the essential content for matching with expected holons


evomimic:
@dauphin3 : I created Issue #99 to implement the temporary enhancement we discussed and assigned it to you. Please resolve that issue and commit it to this PR before you start on Issue #88.

evomimic:
I changed get_key a bit...

1. When a "key" property is found, attempt to convert its value to MapString before giving up and returning an error.
2. Change the error returned to UnexpectedValueType. InvalidParameter is intended signal cases when one or more input parameters to a function are invalid. In this case, the issue is the ValueType of the key Property, so I added the UnexpectedValueType error to holon_error and mapped it to ServerError in dance_response.

All tests pass... code looks good... approving this PR.

Issue #100, opened 2024-06-07
Tryorama Test Case 1 Bug


evomimic:
This bug came to light in `main` after merging PR #98. Task 5 submits a `stage_new_holons` dance with a `Parameters` dance body, but this dance requires either a `None` or `Holon`(Holon) request body

nphias:
fixed by https://github.com/evomimic/map-holons/pull/102

PR #102, opened 2024-06-11
updated tests with helper classes

nphias:
fixes bug : https://github.com/evomimic/map-holons/issues/100


tests/src/map_holons/holons/common.ts
evomimic, line 1: Keep this clean by just making parameters cell and request:DanceRequestObject

nphias: code was cleaned up and ready to merge with main

evomimic:
- The get_all_holons dance retrieves all holons from the persistent store (i.e., DHT) NOT the staging area.
- I'm proposing we stick with the terminology _stage_new_holon_ in preference to _create_ or _add_ to reinforce the distinction between staging holons and committing them to the persistent store.

I added a commit that updated the test case console messages to more accurately reflect the above.

evomimic:
What this enhancement brought into focus is the need for a deeper discussion and definition of the MAP's client-side architecture. Essentially, we need to implement the _model layer_ of the client-side MVC architecture for the MAP's Uniform API. Since the Uniform API is THE interface for invoking everything in the MAP, it is worth devoting some extra attention to it.

I believe this _model layer_ will look significantly different than the _tests/src/map_holons/holons/common.ts_ definitions included in this PR. But rather than hold up this PR, I'm going to proceed with the merge and document my thoughts elsewhere regarding design of the _model layer_ classes and functions.

PR #134, opened 2024-08-01
112 complete load core schema as holons

evomimic:
Much of the Core Schema Load is working and there have been enough other changes that I think it makes sense to merge this branch to `main` prior to the holochain 0.3 upgrade.


evomimic:
List of changes included in this PR:

1. Introduced core_schema zome and load_core_schema "guest-side" function & dance
2. Introduced lazy_get strategy for loading schema components
3. Added "define_xxx_type" functions for each built-in type to allow agent-defined types to be defined
4. Added (but are not currently using) struct definitions for each of the MAP Descriptor types
5. Introduced "TransientCollections"
6. Added json_adapter for streamlined tracing -- generates a more compact representation of MAP Holons
7. Added a database_print test step to allow a full (but compact) database dump to be written to log
9. Moved definition of HolonGettable Trait from Holon to HolonReference and dropped implementation of HolonGettable from Holon (since holon doesn't have access to "context")
10. Dropped support for get_all_related_holons (for now)
11. Removed need for context in EssentialContent functions
12. Added ability to populate key and relationship maps from Smartlinks on Holon fetch
13. Re-factored is_accessible method in Holon and HolonCollection for greater readability 
14. Relaxed accessibility policies to allow READ access to abandoned holons and updated abandon test case accordingly
15. Relaxed accessibility policies to allow WRITE access to FETCHED holons to allow lazy load of its relationship_map

evomimic:
Not fully functional, but all tests pass. Will work remaining issues on case-by-case basis.

PR #137, opened 2024-08-01
128 invalid utf8 couldnt convert link tag bytes into string

dauphin3:
- encode and decode via bytes instead of string
- add unit test for both functions
- fix write access for fetched state (since it is for cached writing NOT persisted)


evomimic:
Completed code review. All looks good.
All test cases pass sweetests EXCEPT the abandon test case. I'm going to approve the merge anyway and address the issues in branch 112.

PR #141, opened 2024-08-03
133 upgrade to now supported 03x holochain

nphias:
a few warnings left which could be cleaned up

- changes with validation functions - admittedly a bit hacked .. deleteOps etc - suggestion in upgrade docs to scaffold a new 0.3x app to see validation changes 
- changes with how to do getlinks (note i have not optimized for GetOptions::network() or GetOptions::local())
- changes to typescript client and api


evomimic:
Code looks fine. All tests (both sweetest & tryorama) pass.

PR #144, opened 2024-08-15
95 implement removerelatedholons dance

nphias:
draft PR for WiP checkup ..  do not merge


nphias:
all done and squashed into one commit.
note: tryorama tests, have been left for another issue, as more work is needed to update structures from multiple api changes

evomimic:
I had to disable the test sending bad staged index because it caused test code to panic before even building the dance request. Everything else looks good. All sweetests pass.

Issue #118, opened 2024-06-25
Add stage_ new_from_clone and stage_new_version Dances


evomimic:
#### NOTE: This issue has been renamed and its implementation substantially redesigned since it was initially defined.
The renaming was to necessary to make the new dance names more consistent with the existing _stage_new_holon_ dance. The result of all three dances is a staged holon that is ready to be updated or committed.

The redesign is reflected in the proposal section below.

## Current State

The existing _clone_holon_ and _edit_holon_ methods drag the existing holon's property_map and relationship_map up into the reference layer to do the copying. 

## Proposal

1. Allow both existing and staged holons to be cloned. Only existing holons may have new versions staged.
2. The basic behavior of cloning an existing holon should be pushed down to the Shared Objects (Holon) Layer. The resulting clone will have the same relationship map as the original holon. This means that both the DESCRIBED_BY relationship PREDECESSOR will be copied from the original holon.
3. Responsibility for cleaning up the PREDECESSOR relationship and staging the cloned holon with the CommitManager should be implemented in the Reference layer.

### In holon.rs file

- [x] `clone_holon` should create a new() Holon and then copy the `property_map` and `relationship_map` from the original to the clone.
- [x] rename `edit_holon` to `new_version` and change its implementation. It  should simply call `self.clone_holon()` and then set the cloned holon's `state` to `Changed` before returning

### In holon_reference.rs file

- [x] Rename `clone_holon` to `stage_new_from_clone` method.  This method should delegate the call to its variants. Each variant should return the cloned holon. Next reset PREDECESSOR to none, then  invoke `stage_new_holon` on the commit_manager and then return the staged index of the cloned holon.

### In smart_reference.rs file

- [x] Rename `clone_holon` to `stage_new_from_clone` method. This method should use `get_rc_holon` to get it's holon and then call `clone_holon()` on that holon and return the cloned holon.

- [x] Add a `new_version` method. This method should use `get_rc_holon` to get its holon and then call `new_version()` on that holon.  It should then reset the PREDECESSOR relationship on the cloned holon to refer to the original holon. Next, it should invoke `stage_new_holon` on the commit_manager and then return the staged index of the cloned holon.

### In staged_reference.rs file

- [x] Add a `stage_new_from_clone` method. This method should use `get_rc_holon` to get it's holon and then call `clone_holon()` on that holon and return the cloned holon
### In commit_manager.rs file

- [x] Rename `clone_holon` method to `stage_new_from_clone`
[...]

PR #156, opened 2024-10-01
75 implement DeleteHolon dance undescribed

dauphin3:
Notes:
- Holon::is_deletable() always returning Ok until support for get_all_related_holons
- executor designed to be flexible to pass any ResponseStatusCode error (although NotFound is the most logical/obvious and the only one being used in this issue)
- only 1 Holon is being created in the fixture until support for get by key for created_holons is available in the new BTreeMap design (see #118)


crates/coordinator/dances/src/dance_response.rs
evomimic, line 1: I believe 409 Conflict is a better mapping than BadRequest for DeletionNotAllowed

crates/coordinator/dances/tests/shared_test/dance_fixtures.rs
evomimic, line 1: use holochain::core::author_key_is_valid is not needed.

evomimic (approved):
All looks good!

PR #183, opened 2024-12-12
175 stagedreferences from nursery

nphias:
Beginning of a work around the QueryManager  

- QueryManager with HolonFacade (shared access to the SpaceManager)
- added QueryManager to the context
- added traits to confine access to the SpaceManager's staging area / nursery
- moved mutator functions out of public trait in nursery
- renamed get_stage to get_all_holons
- updated flake and packages


evomimic:
Everything compiles and runs. Sweetests pass. 

There is some additional simplification and segmentation I want to do to more clearly distinguish the architectural layers and ensure a better separation of concerns. For example, the HolonStageQuery trait includes methods that return Holons instead of HolonReference. These methods should not be offered by the Reference Layer. As another example, the method:

`fn get_holon_by_index(&self,index: usize)-> Result<Rc<RefCell<Holon>>, HolonError>`

is ONLY intended to be used by the internal implementation of StagedReference. Thus, a trait offering this method should belong (ONLY) to the SharedObjectsLayer and not be visible in the API to the ReferenceLayer.

More details are available in this discussion [Limiting Direct Access to Holons](https://github.com/evomimic/map-holons/discussions/184) discussion.

I'm going to approve and merge this PR and further refinements can be worked in smaller issues.

PR #210, opened 2025-01-30
201 introduce holoncollectionapi add transientstateservice to holon space manager

dauphin3:


crates/holons_core/src/core_shared_objects/holon_collection.rs
evomimic, line 1: The intent here was to MOVE the existing implementations of the trait methods to the HolonCollectionApi impl block, NOT wrap those implementations in another layer.

crates/holons_core/src/core_shared_objects/space_manager.rs
evomimic, line 1: Since we decided to use Rc instead of Arc, replace this comment;
// Arc: Allows safe sharing of transient_state across threads.

with this comment:
// Rc: Allows shared ownership of transient_state

evomimic, line 1: Change:
/// Retrieves a shared, thread-save reference to the transient_state.
to:
/// Retrieves a shared reference to the transient_state

crates/holons_core/src/core_shared_objects/transient_collection.rs
evomimic, line 1: Same comment here as for HolonCollection. The trait implementation simply delegates to the inherent impl methods. While this works, it’s not ideal for maintainability and idiomatic Rust patterns. It introduces redundancy—every method has two definitions that must be kept in sync. It also obscures what parts of the struct's behavior is not covered by ANY trait. 

Suggestion:
Move the method implementations into the impl HolonCollectionApi for TransientCollection block.

crates/holons_core/src/reference_layer/space_manager_behavior.rs
evomimic, line 1: Drop the references to thread-safety from the doc comment preceding transient_state. Replace the comment on Arc with Rc and correct the return type in the doc comment.

zomes/coordinator/dances/tests/shared_test/test_add_related_holon.rs
evomimic, line 1: Drop the superfluous "f" from:
use holochain::sweettest::*;f

evomimic (changes requested):

evomimic (approved):

evomimic:
I going to approve/merge this PR and carry forward a couple of things to a new issue:

- [ ] test_add_related_holons: Drop the superfluous "f" from: `use holochain::sweettest::*;f`
- [ ] space_manager: I don't think this use is correct:` use hdi::prelude::ShardStrategy;`
- [ ] it appears we've added imports of `HolonCollectionApi` in some files that never reference `HolonCollectionApi`

And the next phase of incorporating HolonCollectionApi will carry its use into some new areas so that we have fully encapsulated HolonCollection. This will be elaborated in the next issue.

Issue #224, opened 2025-02-28
Restore get_all_holons Capability


evomimic:
### 1. Summary (Required)

Add support for `get_all_holons` to the `holons_operations_api` and ensure end-to-end implementation for both client-side and guest-side. The existing `get_all_holons_dance` is modified to return a `HolonCollection` instead of a `Vec<Holon>`. This makes this a lighter-weight operation and also means we can leverage standard `HolonsCache` methods to retrieve and provide access to the Holons themselves on demand. Since `HolonCollection` is a new response type, it needs to be added to the set of possible `DanceResponses`. 

- Enhance `holon_service_api` to include: `pub fn get_all_holons() -> Result<HolonCollection, HolonError>` and provide guest-side and client-side implementations of this function

NOTE: The implementation of this enhancement will likely be replaced by Issue #151 once we have support for inverse relationships and HolonSpaces.

---

### 2. Problem Statement (Required)

The Holon type previously supported `pub fn get_all_holons() -> Result<Vec<Holon>, HolonError>`. It's implementation leveraged the `all_holon_nodes` file generated by the Holochain Scaffolding Tool.  This functionality was exposed via the `get_all_holons_dance` and was used often in the _dances_tests_ layer.  With the re-factoring, there is no longer support for this function. This is causing virtually all existing dances test cases to fail.

---

### 3. Dependencies (Required)

- [x] Depends on the DanceCall  and DanceCallService enhancements delivered via Issue #219 
- [x] Client-side implementation depends upon Issue #205 (or a subset delivered via Issue #219)

---

#### In _holons_core/reference_layer_

- [x] Add a `get_all_holons` trait function to `HolonServiceApi`
    `get_all_holons(&self, context: &dyn HolonsContextBehavior)->Result<HolonCollection, HolonError>`

- [x] Add a `get_all_holons` function to the  `holon_operations_api`:
    `pub fn get_all_holons(context: &dyn HolonsContextBehavior) -> Result<HolonCollection, HolonError>`. Note that this returns a collection of references instead of the Holons themselves. The following diagram describes the implementation.

![Image](https://github.com/user-attachments/assets/f4057825-2049-4190-a8c7-e03b0b14c569)

#### in _holons_guest/guest_shared_objects_

![Image](https://github.com/user-attachments/assets/a75dffc4-dfeb-4fe2-8acc-57edbfb8a272)

[...]

PR #225, opened 2025-02-28
219 revert dancetests to prior state

evomimic:
Sweetests now compile, but don't run due to lack of get_all_holons support (used by ensure_db_count and print_database test steps. Merging this now so that Issue #224 can use this as a basis.


crates/holons_client/tests/shared_test/fixtures/abandon_staged_changes_fixture.rs
```diff
@@ -43,19 +43,20 @@ pub fn simple_abandon_staged_changes_fixture() -> Result<DancesTestCase, HolonEr
...
 
-    let mut holons_to_add: Vec<HolonReference> = Vec::new();
+    let holons_to_add: Vec<HolonReference> = Vec::new();
 
     // Use helper function to set up a book holon, 2 persons, a publisher, and an AUTHORED_BY relationship from
     // the book to both persons.
 
-    let relationship_name = setup_book_author_steps_with_context(fixture_context, &mut test_case)?;
+    let relationship_name =
+        setup_book_author_steps_with_context(&*fixture_context, &mut test_case)?;
```
dauphin3, line 52: the `&*` seems like an odd pattern to reference a dereference, maybe it should be Arc::clone or Arc::downgrade
https://doc.rust-lang.org/std/sync/struct.Arc.html#deref-behavior

dauphin3 (approved):
see comment about `&*` pattern

crates/holons_client/tests/shared_test/fixtures/abandon_staged_changes_fixture.rs
evomimic: I agree it is awkward, @dauphin3 .  It stems from the need to retrieve the `context` from within `DanceTestExecutionState` rather than just passing `context` directly as a parameter as we typically do in production code. I've added Issue #227 to hide this bit of awkwardness. Thanks for pointing this out!

PR #239, opened 2025-03-12
upgraded to hc 0.4, fixed UI

nphias:


evomimic:
Got past the compile error

I got thing to compile by replacing:
use holochain::prelude::dependencies::kitsune_p2p_types::dependencies::holochain_trace;

with just
use tracing;

But all six sweetests are running, but failing.

running 6 tests
test rstest_dance_tests::case_2_delete_holon ... FAILED
test rstest_dance_tests::case_4_simple_add_related_holon_test ... FAILED
test rstest_dance_tests::case_5_simple_stage_new_from_clone_test ... FAILED
test rstest_dance_tests::case_1_simple_undescribed_create_holon_test ... FAILED
test rstest_dance_tests::case_6_simple_stage_new_version_test ... FAILED
test rstest_dance_tests::case_3_simple_abandon_staged_changes_test ... FAILED

failures:

---- rstest_dance_tests::case_2_delete_holon stdout ----
Current working directory: "/Users/stevemelville/dev/map-proto/map-dev/map-holons/crates/holons_client"
FATAL PANIC PanicHookInfo {
    payload: Any { .. },
    location: Location {
        file: "crates/holons_client/tests/shared_test/mock_conductor.rs",
        line: 39,
        col: 10,
    },
    can_unwind: true,
    force_no_backtrace: false,
}
thread 'rstest_dance_tests::case_2_delete_holon' panicked at crates/holons_client/tests/shared_test/mock_conductor.rs:39:10:
called `Result::unwrap()` on an `Err` value: ConductorError(RibosomeError(WasmRuntimeError(RuntimeError { source: User(WasmError { file: "/Users/stevemelville/.cargo/registry/src/index.crates.io-6f17d22bba15001f/holochain-0.4.2/src/core/ribosome/real_ribosome.rs", line: 453, error: ModuleBuild("holons_integrity: Error while importing \"__wbindgen_placeholder__\".\"__wbindgen_describe\": unknown import. Expected Function(FunctionType { params: [I32], results: [] })") }), wasm_trace: [] })))
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

evomimic:
# Update from 2025-03-12 at 6:18pm MDT

I was able to get everything to compile by using:
`use holochain_trace;`

in place of:
`use holochain::prelude::dependencies::kitsune_p2p_types::dependencies::holochain_trace;`

#### All sweetests fail.

To simplify debugging, I disabled all tests except one:

 Running tests/dance_tests.rs (target/debug/deps/dance_tests-99f6bdcd2a4ad4a9)

running 1 test
2025-03-13T00:07:48.545575Z  INFO holochain_sqlite::schema: /Users/stevemelville/.cargo/registry/src/index.crates.io-6f17d22bba15001f/holochain_sqlite-0.4.2/src/schema.rs:95: database forward migrated: Conductor from 0 to 1
2025-03-13T00:07:48.546348Z  INFO holochain_sqlite::schema: /Users/stevemelville/.cargo/registry/src/index.crates.io-6f17d22bba15001f/holochain_sqlite-0.4.2/src/schema.rs:95: database forward migrated: Wasm from 0 to 0
2025-03-13T00:07:48.547841Z  INFO kitsune_p2p_types::tx2::mem: /Users/stevemelville/.cargo/registry/src/index.crates.io-6f17d22bba15001f/kitsune_p2p_types-0.4.2/src/tx2/mem.rs:418: bound local endpoint (mem) url=kitsune-mem://1
2025-03-13T00:07:48.547850Z  INFO kitsune_p2p_types::tx2::tx2_restart_adapter: /Users/stevemelville/.cargo/registry/src/index.crates.io-6f17d22bba15001f/kitsune_p2p_types-0.4.2/src/tx2/tx2_restart_adapter.rs:39: tx2_restart_adapter bound Ok(kitsune-mem://1)
2025-03-13T00:07:48.548816Z  INFO holochain::conductor::conductor::builder: /Users/stevemelville/.cargo/registry/src/index.crates.io-6f17d22bba15001f/holochain-0.4.2/src/conductor/conductor/builder.rs:392: Conductor startup: scheduler task started.
2025-03-13T00:07:48.548825Z  INFO holochain::conductor::conductor::builder: /Users/stevemelville/.cargo/registry/src/index.crates.io-6f17d22bba15001f/holochain-0.4.2/src/conductor/conductor/builder.rs:396: Conductor startup: p2p event task started.
2025-03-13T00:07:48.550069Z  INFO holochain::conductor::conductor::startup_shutdown_impls: /Users/stevemelville/.cargo/registry/src/index.crates.io-6f17d22bba15001f/holochain-0.4.2/src/conductor/conductor.rs:390: Conductor startup: DNAs loaded.
2025-03-13T00:07:48.550328Z  INFO holochain_websocket: /Users/stevemelville/.cargo/registry/src/index.crates.io-6f17d22bba15001f/holochain_websocket-0.4.2/src/lib.rs:928: WebsocketListener listening addr=127.0.0.1:53064
2025-03-13T00:07:48.550333Z  INFO holochain_websocket: /Users/stevemelville/.cargo/registry/src/index.crates.io-6f17d22bba15001f/holochain_websocket-0.4.2/src/lib.rs:931: WebsocketListener listening addr=[::1]:53064
2025-03-13T00:07:48.550395Z  INFO holochain::conductor::conductor::startup_shutdown_impls: /Users/stevemelville/.cargo/registry/src/index.crates.io-6f17d22bba15001f/holochain-0.4.2/src/conductor/conductor.rs:404: Conductor startup: admin interface(s) added.
2025-03-13T00:07:48.550497Z  INFO holochain::conductor::conductor::startup_shutdown_impls: /Users/stevemelville/.cargo/registry/src/index.crates.io-6f17d22bba15001f/holochain-0.4.2/src/conductor/conductor.rs:408: Conductor startup: app interfaces started.
2025-03-13T00:07:48.550787Z  INFO holochain::conductor::conductor::startup_shutdown_impls: /Users/stevemelville/.cargo/registry/src/index.crates.io-6f17d22bba15001f/holochain-0.4.2/src/conductor/conductor.rs:415: Conductor startup: apps started.
2025-03-13T00:07:49.315453Z  INFO holochain::conductor::conductor::startup_shutdown_impls: /Users/stevemelville/.cargo/registry/src/index.crates.io-6f17d22bba15001f/holochain-0.4.2/src/conductor/conductor.rs:376: Sending shutdown signal to all managed tasks.
2025-03-13T00:07:49.315651Z  INFO managed task{scope="0.pKlUm"}: holochain_websocket: /Users/stevemelville/.cargo/registry/src/index.crates.io-6f17d22bba15001f/holochain_websocket-0.4.2/src/lib.rs:813: WebsocketListenerDrop
2025-03-13T00:07:49.315679Z  INFO holochain::conductor::conductor: /Users/stevemelville/.cargo/registry/src/index.crates.io-6f17d22bba15001f/holochain-0.4.2/src/conductor/conductor.rs:4077: p2p_event_task has ended
2025-03-13T00:07:49.315850Z  INFO kitsune_p2p_types::tx2::mem: /Users/stevemelville/.cargo/registry/src/index.crates.io-6f17d22bba15001f/kitsune_p2p_types-0.4.2/src/tx2/mem.rs:418: bound local endpoint (mem) url=kitsune-mem://2
2025-03-13T00:07:49.315858Z  INFO kitsune_p2p_types::tx2::tx2_restart_adapter: /Users/stevemelville/.cargo/registry/src/index.crates.io-6f17d22bba15001f/kitsune_p2p_types-0.4.2/src/tx2/tx2_restart_adapter.rs:39: tx2_restart_adapter bound Ok(kitsune-mem://2)
test rstest_dance_tests::case_1_simple_undescribed_create_holon_test ... FAILED

failures:

---- rstest_dance_tests::case_1_simple_undescribed_create_holon_test stdout ----
Initialising log output formatting with option Log
Current working directory: "/Users/stevemelville/dev/map-proto/map-dev/map-holons/crates/holons_client"
FATAL PANIC PanicHookInfo {
    payload: Any { .. },
    location: Location {
        file: "crates/holons_client/tests/shared_test/mock_conductor.rs",
        line: 39,
        col: 10,
    },
    can_unwind: true,
    force_no_backtrace: false,
}
thread 'rstest_dance_tests::case_1_simple_undescribed_create_holon_test' panicked at crates/holons_client/tests/shared_test/mock_conductor.rs:39:10:
called `Result::unwrap()` on an `Err` value: ConductorError(RibosomeError(WasmRuntimeError(RuntimeError { source: User(WasmError { file: "/Users/stevemelville/.cargo/registry/src/index.crates.io-6f17d22bba15001f/holochain-0.4.2/src/core/ribosome/real_ribosome.rs", line: 453, error: ModuleBuild("holons_integrity: Error while importing \"__wbindgen_placeholder__\".\"__wbindgen_describe\": unknown import. Expected Function(FunctionType { params: [I32], results: [] })") }), wasm_trace: [] })))
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    rstest_dance_tests::case_1_simple_undescribed_create_holon_test


I committed and pushed my changes to origin.

nphias:
now only the holons_client has non-wasm compatible dependencies .. however i am still getting the wasm build error.. it seems like its occuring with holons_integrity... so i might test that part now

evomimic:
> now only the holons_client has non-wasm compatible dependencies .. however i am still getting the wasm build error.. it seems like its occuring with holons_integrity... so i might test that part now
> 

Hmm... holons_integrity should only be on the guest-side
> 
> 
> other changes .. only crates that will build wasm need `crate-type = ["cdylib", "rlib"]` every other crate is only `crate-type = ["rlib"]`

nphias:
> > now only the holons_client has non-wasm compatible dependencies .. however i am still getting the wasm build error.. it seems like its occuring with holons_integrity... so i might test that part now
> 
> Hmm... holons_integrity should only be on the guest-side
> 
> > other changes .. only crates that will build wasm need `crate-type = ["cdylib", "rlib"]` every other crate is only `crate-type = ["rlib"]`

exactly and the guest_side  (wasm) is being compiled and executed by the sweetconductor when running holon_client tests

evomimic:
Let's not lose sight of the fact that everything compiles and runs under Holochain 3. So the issues we are seeing are related to changes introduced by holochain 4.

crates/holon_dance_builders/Cargo.toml
evomimic, line 1: Restore this to support both "cdylib" and "rlib" -- the holons_guest crate will need to build dance requests to submit to external HolonSpaces, so holon_dance_builders needs to be WASM compatible.

evomimic:
# Review of Minimalist Upgrade Approach

I've been puzzled by the scope of the changes in this PR and wanted to better understand what motivated them. So I created a separate simple-holochain-4-upgrade branch. 

## Basic Approach

The idea was to adopt a minimalist approach:

1. Confirm sweetests run under Holochain 0.3

2. Do the Holochain 0.4 upgrade

3. Only make changes that are strictly necessary to get holochain 4 to work

4. Carefully document the reasons why such changes were necessary.

## Summary of Findings

- Confirmed sweetests 1 thru 6 all PASS under holochain 0.3 -- exception is *load_core_schema*, but this is a known issue unrelated to the upgrade. 

- I've discovered the reasons behind most of the files changed for this PR

- Identified a one change I recommend we drop

- Identified some additional opportunities for improvement.

## Necessary Changes

### Necessary to Do the Upgrade Itself

1. Edit the flake.nix to change holonix version, packages, Rust toolchain and setup.

    | Aspect                       | Pre-234                                              | 234                                                         |
    | :--------------------------- | :--------------------------------------------------- | :---------------------------------------------------------- |
    | **Holonix version**          | `main-0.3`                                           | ✅ `main-0.4`                                                |
    | **Tooling source**           | Tools listed manually via `inputs'.holonix.packages` | ✅ Uses `inputsFrom = [ inputs'.holonix.devShells.default ]` |
    | **Rust toolchain**           | Explicit `rust` from Holonix packages                | ✅ Comes implicitly from Holonix 0.4 devShell                |
    | **Structure**                | Manual setup of Holochain tools                      | ✅ Cleaner, flake-parts-compliant integration                |
    | **macOS framework optional** | Included in both                                     | ✅ Same behavior                                             |

### Minimal Required Changes

After making the changes for the upgrade. The project no longer compiled. 

1. ***wasmer-related errors***. These errors arose in Cargo.toml files with dev-dependencies on holochain (required for sweetest). It seems Holochain 0.4 upgraded its internal dependency on wasmer from 4.x to 5.x, and that introduced a new feature setup in wasmer. At first, I tried to address the errors by manually including addition wasmer-specific crates and features. But, ultimately I discovered the only required change was to add "wasmer_sys" to the holochain features list:

    ```text
    [target.'cfg(not(target_arch = "wasm32"))'.dev-dependencies]
    holochain = { version = "0.4.2", default-features = false, features = [
        "test_utils",
        "sqlite-encrypted",
        "wasmer_sys"
    ] }
    
    wasmer = { version = "5.0.4", features = ["cranelift"] }
    ```



2. ***Deep holochain::prelude imports***. After getting past the wasmer errors, I ran into these errors caused by trying to access deeply nested internal dependencies via the `holochain::prelude` re-export, which has changed significantly between Holochain 0.3.x → 0.4.x.

    ```text
    error[E0433]: failed to resolve: could not find dependencies in prelude
      --> crates/holons_client/tests/shared_test/fixtures/add_remove_related_fixture.rs:14:25
       |
    14 | use holochain::prelude::dependencies::kitsune_p2p_types::dependencies::lair_keystore_api::config::get_server_pub_key_from_connection_url;
       |                         ^^^^^^^^^^^^ could not find dependencies in prelude
    ```

    Solution:

    - **Delete all imports of** `holochain::prelude::dependencies`

    - Explicitly add holochain_trace to dev-dependencies.

        ```text
        [target.'cfg(not(target_arch = "wasm32"))'.dev-dependencies]
        holochain_trace = "0.4"
        ```

3. ***Sweetests hang***. After correcting the above errors, everything compiles, but all of the sweetests would just hang. They weren't reporting errors, but the following messages were displayed and then... silence.

    ```text

        running 6 tests
        test rstest_dance_tests::case_1_simple_undescribed_create_holon_test has been running for over 60 seconds
        test rstest_dance_tests::case_2_delete_holon has been running for over 60 seconds
        test rstest_dance_tests::case_3_simple_abandon_staged_changes_test has been running for over 60 seconds
        test rstest_dance_tests::case_4_simple_add_related_holon_test has been running for over 60 seconds
        test rstest_dance_tests::case_5_simple_stage_new_from_clone_test has been running for over 60 seconds
        test rstest_dance_tests::case_6_simple_stage_new_version_test has been running for over 60 seconds

    
    ```

    The culprit was in the `conductor_dance_call` function (within the `mock_conductor` module). 

    ```rust
    /// Implements `DanceCaller` for the Sweetest mock conductor.
    ///
    /// This allows `MockConductorConfig` to be used inside `DanceCallService`.
    impl ConductorDanceCaller for MockConductorConfig {
        fn conductor_dance_call(&self, request: DanceRequest) -> DanceResponse {
            futures::executor::block_on(async {
                self.conductor.call(&self.cell.zome("holons"), "dance", request).await
            })
        }
    }
    ```

    According to ChatGPT, this 

    > …was *barely OK* in Holochain 0.3 because Sweettest used async-std internally, and test setups were less sensitive.

    > But in **Holochain 0.4**, the entire test harness is built on top of **Tokio**, and the conductor is run **inside an async runtime**. Using block_on() here tries to **synchronously block on an already-running async executor**, which:

    > 🚫 Causes a deadlock — and your test silently hangs.

    #### Solution:

    1. Make the conductor_dance_call async

        ```rust
        pub async fn conductor_dance_call(&self, request: DanceRequest) -> DanceResponse {
            self.conductor
                .call(&self.cell.zome("holons"), "dance", request)
                .await
        }
        ```



    2. Change all of our test functions to use .await

        ```rust
        let response = self.conductor_dance_call(request).await;
        ```

    ### Summary of Minimal Required Changes

    Fixing the above problems accounts for (and motivates) the majority of changes observed in the PR. 

## Additional Changes Included in this PR

This PR included some additional changes that were not strictly required under a minimalist approach. Each is evaluated below.

### Dropping "cdylib" from Cargo.toml in crates/ directory

 Now that we've begun splitting zome crates from the Rust crates they bundle, the supporting crates (e.g., `holons_core`, `holons_guest`, `shared_types`) can drop "cdylib" (i.e., WASM) as a crate type. This allows them to be  reused in both **WASM** (zomes) and **native** (host/client/test) contexts. **Only crates that are actual zomes need crate-type = ["cdylib", "rlib"]**

This change is not strictly necessary under a minimalist upgrade approach, but it is a sensible change to make. 

***It is worth noting, however, that under Option 3 of our MAP runtime architecture, the*** `holons_client` ***crate will need to be bundled as a WASM library. Hence, it may need to include the "cdylib" crate-type.*** 

**Recommendation**: Keep this change. 

### Removing `holons_client` crate from workspace-level Cargo.toml

This PR drops the `holons_client` crate from the workspace-level Cargo.toml. The `holons_client` crate was introduced to allow important Rust functionality we have implemented around Holon staging and caching to be leveraged on both the guest-side and client-side of our holochain architecture. We re-factored our code into `holons_core` (shared by both sides), `holons_guest` (guest-side only logic) and `holons_client` (client-side only logic). The idea was to bundle `holons_client` as a WASM library that could be linked into our typescript/js client code. The `holons_client` defines an api that wraps the dance calls on the holochain conductor. 

While we are not (yet) testing the higher-level API we envision for the holons_client, we ARE using some utility functions (e.g., a dance_call_service) from the holons_client within our sweetest code.

***Recommendation: Retain the holons_client crate as a workspace member.***

    - This crate includes modules that ARE being used for testing (e.g., dance_call_service). So it is still being built, depended on, and maintained.

        - Keeping holons_client in the workspace gives us:

            - Unified builds: cargo check, cargo test, cargo doc all Just Work™
            - Shared dependency resolution (via workspace = true)
            - Code indexing, IDE support, CI correctness
            - Easy testing (esp. now that sweet_tests depends on it)

      - Removing it:
  
          - Forces us to reference it by path (awkward and error-prone)
          - Breaks workspace-level cargo commands
          - Makes the crate feel like a second-class citizen
          - Risks it going stale



## Additional Suggested Improvements

Finally... going through this exercise led to some additional suggested improvements.

### Re-structure Directory Structure for Tests

  To improve clarity, modularity, and maintainability, we should move our sweetests out of the `holons_client` crate (or ANY of our production-code crates) and into the dedicated top-level `tests` directory: 
  
  ```text
  tests/
  ├── tryorama/      # JS/TS conductor tests using Tryorama
  ├── sweet_tests/   # Rust-based integration tests using Sweetest
  
  ```
  
  *Impacts:*
  
  - `holons_client` and other crates no longer require dev-deps like tokio, holochain, or wasmer
  - Rust tests are now in tests/sweet_tests, with their own Cargo.toml  
  - JS/TS Tryorama tests live in tests/tryorama, with their own package.json (if applicable)

### Update dna.yaml to Manifest Version 2

Our dna.yaml is actually structured according to the Version 2 manifest structure, but declares manifest_version = "1". The fact that it works at all is probably due to **Holochain’s tooling being lenient** during the transition — but officially Version 1 **does not support** the integrity/coordinator split. That is **only valid for manifest_version: "2"**

See [https://docs.holochain.org/reference/happ\_manifest](https://docs.holochain.org/reference/happ\_manifest) for more details

nphias:
> Update dna.yaml to Manifest Version 2
> 
> Our dna.yaml is actually structured according to the Version 2 manifest structure, but declares manifest_version = "1". The fact that it works at all is probably due to Holochain’s tooling being lenient during the transition — but officially Version 1 does not support the integrity/coordinator split. That is only valid for manifest_version: "2"
> 
> See https://docs.holochain.org/reference/happ_manifest for more details

i see no difference in the manifest from a 0.4 or 0.5 scaffold and our manifest.. and the link you provided does not exist

evomimic:
> > Update dna.yaml to Manifest Version 2
> > Our dna.yaml is actually structured according to the Version 2 manifest structure, but declares manifest_version = "1". The fact that it works at all is probably due to Holochain’s tooling being lenient during the transition — but officially Version 1 does not support the integrity/coordinator split. That is only valid for manifest_version: "2"
> > See https://docs.holochain.org/reference/happ_manifest for more details
> 
> i see no difference in the manifest from a 0.4 or 0.5 scaffold and our manifest.. and the link you provided does not exist

OK... It was my understanding that when they introduced the integrity/coordinator split, they bumped the manifest number to "2". What I was trying to highlight was that although the "structure" looked fine, the manifest number itself should be "2". But if the scaffolding tool still generates "1" then we should be fine. I couldn't find anything in the holochain docs anywhere that offers any explanation or guidance on manifest number, so it seems like everything's fine and we can ignore my suggestion here.

evomimic:
In my suggestion to move the sweetests to the top-level tests directory, I also suggested we put the tryorama tests there. Those unifies all the integration test logic under the tests directory. Having a directory with the _tryorama_ name makes its clearer where those tests live. Is there a reason you didn't make that change?

Cargo.toml
evomimic, line 1: For maintainability, we are trying to keep things in alphabetical order.

crates/holons_client/Cargo.toml
```diff
@@ -4,33 +4,15 @@ version = "0.1.0"
...
-
-# Holochain Dependencies
-#hdi = { workspace = true }
+tracing = { workspace = true }
+async-trait = { workspace = true }
 
 # MAP Dependencies
 shared_types_holon = { workspace = true }
 holons_core = { workspace = true }
-holon_dance_builders = { workspace = true }
```
evomimic, line 23: Why were holon_dance_builders dropped from this Cargo.toml?

package.json
```diff
@@ -15,17 +15,16 @@
...
     "sweetest": "npm run sweetest:custom",
-    "sweetest:quiet": "npm run build:happ && export WASM_LOG=$H_LOG,dances=info,holons=info,descriptors=info && export RUST_LOG=$H_LOG,dances=info,holons=info,descriptors=info && cargo test",
-    "sweetest:info": "RUST_LOG=info WASM_LOG=info npm run build:happ && cargo test",
-    "sweetest:debug": "RUST_LOG=debug WASM_LOG=debug npm run build:happ && cargo test",
-    "sweetest:custom": "RUST_LOG=${RUST_LOG:-warn} WASM_LOG=${WASM_LOG:-warn} npm run build:happ && clear && cargo test",
+    "sweetest:quiet": "npm run build:happ && export WASM_LOG=$H_LOG,dances=info,holons=info,descriptors=info && export RUST_LOG=$H_LOG,dances=info,holons=info,descriptors=info && npm run client:test",
+    "sweetest:info": "export RUST_LOG=info,WASM_LOG=info && npm run build:happ && npm run client:test",
+    "sweetest:debug": "export RUST_LOG=info && export WASM_LOG=debug && npm run build:happ && npm run client:test",
+    "sweetest:custom": "export RUST_LOG=warn && export WASM_LOG=warn && npm run build:happ && clear && npm run client:test",
```
evomimic, line 21: The point of the `:custom` variant is to allow specific log levels to be exported BEFORE executing the npm run. This change overrides the exported values -- exactly what we don't. This should be reverted to the previous version which only defaults to warn if for LOG variables not exported.

package.json
```diff
@@ -15,17 +15,16 @@
...
+    "client:test": "cargo test --manifest-path tests/sweetests/Cargo.toml",
```
evomimic, line 22: Why introduce this terminology? These sweetests are NOT testing the client-side code -- calling them client:test is confusing.

package.json
nphias: I dont agree with this. I always design for convention over configuration.. logs should run without having to configure anything / enter env settings. Im happy to save time for other devs

package.json
nphias: changed to sweet:tests

crates/holons_client/Cargo.toml
nphias: sorry added that dependency back.. i was trying to remove unused dependencies .. assume that one will be used in the future

package.json
evomimic: I'm not following how your comment relates to my suggestion. 

My suggestion would not, in any way, _force_ developers to config env settings prior to running tests. But it _would_ restore their ability to, _optionally_, do so in order to give them finer-grained control over tracing levels when needed.

We offer 3 options for initiating sweetest with pre-configured log levels:  `sweetest:quiet,` `sweetest:info`, `sweetest:debug`. These all make it possible for developers to initiate testing without exporting any environment variables. I'm not proposing changing that. 

But these don't cover all the cases, because they all set both client and guest side logging to the same level. What if the developer wants to set client and guest logging differently? There are six more combinations we are not covering. And what if they want to add "trace" level logging on one side or the other? That's 16 total combinations.  We could add (6 to 13) more scripts to cover all the different combinations, but how would we name all those scripts?

Instead, I added one more script as a catch-all: 

`"sweetest:custom": "RUST_LOG=${RUST_LOG:-warn} WASM_LOG=${WASM_LOG:-warn} npm run build:happ && clear && cargo test",`

With no env variables set, this just defaults to `warn` level on both client and guest (so no configuration required by the developer to run the test). But it ALSO allows them to export _either_ or _both_ env variables to precisely what they want and honors that configuration (by not changing it). 

In this PR, the `sweetest:custom` script was changed to force `warn` level on both client and guest, regardless of what the developer has configured. This leaves them  no way to set desired log levels outside of the 3 pre-configured options.

**_I see no reason to restrict developers to ONLY the pre-configured options (nor do I see you proposing this restriction)?_** 

In fact, I have often found it very helpful when debugging issues to be able to run tests with different logging levels on each side.

So, for now, I intend to restore the capability to override log settings, if desired. This avoids regression in previous capabilities.

But there are bigger issues with our test scripts.

-  The transmission of environment variable settings into npm scripts is very problematic and unpredictable. Use of shell scripts gives us more power with less uncertainty.
- We need finer-grained control over tracing level output -- we're getting too much in some areas and not enough in others
- Ability to enable tracing in fixtures
- More ergonomic approach will help -- e.g., we should speak in terms of client tracing, guest tracing rather than WASM_LOG (as an indirect name for guest-side tracing) and RUST_LOG (as indirect name for client-side tracing).

tests/sweetests/tests/shared_test/fixtures/add_remove_related_fixture.rs
```diff
@@ -11,12 +11,12 @@ use holons_core::core_shared_objects::{Holon, HolonCollection, HolonError, Relat
 
 use crate::shared_test::setup_book_author_steps_with_context;
 use crate::shared_test::test_add_related_holon::execute_add_related_holons;
-use holochain::prelude::dependencies::kitsune_p2p_types::dependencies::lair_keystore_api::config::get_server_pub_key_from_connection_url;
+// use holochain::prelude::dependencies::kitsune_p2p_types::dependencies::lair_keystore_api::config::get_server_pub_key_from_connection_url;
```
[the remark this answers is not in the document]
evomimic: removed

tests/sweetests/tests/shared_test/fixtures/add_remove_related_fixture.rs
```diff
@@ -11,12 +11,12 @@ use holons_core::core_shared_objects::{Holon, HolonCollection, HolonError, Relat
...
 
 use crate::shared_test::test_context::init_test_context;
 use crate::shared_test::test_context::TestContextConfigOption::TestFixture;
 use crate::shared_test::test_data_types::BOOK_KEY;
-use holochain::prelude::dependencies::kitsune_p2p_types::dependencies::holochain_trace;
+// use holochain::prelude::dependencies::kitsune_p2p_types::dependencies::holochain_trace;
```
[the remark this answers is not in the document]
evomimic: removed

evomimic:
- [ ] I added on additional commit that resolved remaining issues raised in conversations
- [ ] All sweetests pass except load_core_schema (which is expected... this test has been broken for a while now)
- [ ] Code review completed and conversations resolved
- [ ] Test build is failing, but this is not a show-stopper

PR Approved!

PR #264, opened 2025-06-19
242 refactor holon for improved structure state management and transient support

dauphin3:
getting 
Screenshot 2025-06-17 at 10 40 48 PM

UPDATE:

Rust's borrow checker rules only allow EITHER a single mutable reference or a multiple immutable references to an object... this was affecting Rc<RefCell> when trying borrow_mut() in an attempt to mutate the inner refcell Holon.

This was not a problem before when a Holon was a struct, but migrating to an enum gave challenges for being able to access and subsequently mutate the inner wrapped value.
For example,
specifically in Holon::Staged(StagedHolon) where in get_rc_holon if a StagedHolon was returned the inner value got deeply cloned and a modification was not reflected for the RefCell, but if a Holon enum was returned, doing operations was complicated. The solution turned out to be adding a to_staged() method for Holon which ensures the variant is Staged, then gets the inner StagedHolon. A new mutable variable then needed to be created from borrowing the rc_holon and then cloning it (so as not to break the borrow rules). After the appropriate changes were made, then the rc_holon gets borrowed mutably and updated the inner RefCell value from that variable.


crates/holons_core/src/reference_layer/staged_reference.rs
```diff
@@ -172,15 +204,24 @@ impl HolonWritable for StagedReference {
         context: &dyn HolonsContextBehavior,
```
evomimic, line 204: This implementation works, but it’s less efficient than necessary. Since we’re only updating the `Holon::Staged` variant, we can safely mutate it in place by using `borrow_mut(`) and pattern matching directly on the enum. This avoids both the clone and the extra borrow, which reduces memory usage and simplifies the logic.

I've implemented and tested this change for abandon_staged_changes. 

Please replicate this change for the other mutating methods to avoid the extra cloning.

evomimic (changes requested):
I've completed review of 23 of 113 changed files and I suggest that before I proceed with the review, you resolve the comments I've provided here.

crates/holons_core/src/core_shared_objects/holon/saved.rs
```diff
@@ -0,0 +1,214 @@
...
+                Err(HolonError::NotAccessible(
+                    format!("{:?}", access_type),
+                    "Immutable SavedHolon".to_string(),
+                ))
+            }
+        }
+    }
+
+    // =================
+    //     "MUTATORS"
```
evomimic, line 153: In the `SavedHolon` mutator methods (e.g. `update_property_map`, `increment_version`), the current implementation returns `Ok(())` after calling `is_accessible(...)`. However, `SavedHolon` is immutable by design, and `is_accessible` will always return an error for write-like access types.

Returning `Ok(())` here is misleading — it implies the operation succeeded, even though no mutation occurred. This can obscure bugs and lead to silent failures if `is_accessible` is ever relaxed in the future.

Recommended change:
Replace the final `Ok(())` with an explicit `unreachable!()` call:

```rust
self.is_accessible(AccessType::Write)?;
unreachable!("SavedHolon should never allow write access");
```

This makes the invariant clear and ensures any logic violation will fail loudly during development.

crates/holons_core/src/core_shared_objects/holon/saved.rs
```diff
@@ -0,0 +1,214 @@
...
+        self.is_accessible(AccessType::Write)?;
+        Ok(())
+    }
+
+    // ====================
+    //     DIAGNOSTICS
+    // ====================
+
+    /// Provides structured diagnostic information about the Holon's phase and state.
+    fn debug_info(&self) -> String {
```
evomimic, line 179: Is `debug_info` method really needed (on any variant or overall?).

It is never called and the info it returns is already also returned via `summarize` method.

crates/holons_core/src/core_shared_objects/staged_relationship.rs
```diff
@@ -63,54 +63,90 @@
...
 
-        Ok(Self { map: cloned_relationship_map })
+        Ok(TransientRelationshipMap::new(cloned_relationship_map))
+    }
+
+    // ====================
+    //    DATA ACCESSORS
+    // ====================
+
+    fn get_related_holons(&self, relationship_name: &RelationshipName) -> Rc<HolonCollection> {
```
evomimic, line 121: See TODO on trait: clone required here due to current trait return type.

crates/holons_core/src/core_shared_objects/relationship_behavior.rs
```diff
@@ -0,0 +1,84 @@
...
+    ///
+    /// If the `relationship_name` exists in the Relationship Map, this method returns the
+    /// corresponding collection wrapped in an `Rc`. If the relationship is not found, an empty
+    /// `HolonCollection` wrapped in an `Rc` is returned instead.
+    /// Retrieves the `HolonCollection` for the given relationship name, wrapped in `Rc`.
+    ///
+    /// If the `relationship_name` exists in the Relationship Map, this method returns the
+    /// corresponding collection wrapped in an `Rc`. If the relationship is not found, an empty
+    /// `HolonCollection` wrapped in an `Rc` is returned instead.
+    fn get_related_holons(&self, relationship_name: &RelationshipName) -> Rc<HolonCollection>;
```
evomimic, line 37: TODO(PERF/CLEANUP): This clones the inner HolonCollection, which may be costly in time/space.
Ideally, the trait method would return an associated type (e.g., `type Output: Clone`) to allow
different implementations to return shared ownership types like `Rc<RefCell<_>>`.
For now, we’re preserving this behavior to avoid destabilizing the long-lived feature branch;
revisit after merge for a more efficient and idiomatic design.

crates/holons_core/src/core_shared_objects/transient_relationship.rs
```diff
@@ -0,0 +1,153 @@
...
+        }
+
+        Ok(TransientRelationshipMap::new(cloned_relationship_map))
+    }
+
+    // ====================
+    //    DATA ACCESSORS
+    // ====================
+
+    fn get_related_holons(&self, relationship_name: &RelationshipName) -> Rc<HolonCollection> {
```
evomimic, line 71: See TODO on trait: clone required here due to current trait return type.

crates/holons_core/src/core_shared_objects/holon/saved.rs
dauphin3: deleted

crates/holons_core/src/reference_layer/staged_reference.rs
```diff
@@ -55,7 +56,15 @@ impl StagedReference {
         // Retrieve the holon by its temporaryId
         let rc_holon = nursery_read.get_holon_by_id(&self.id)?;
 
-        // Return a clone of the Rc<RefCell<Holon>>
+        // Confirm it references a StagedHolon and return an Rc<RefCell
```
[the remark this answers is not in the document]
dauphin3: I'm not sure, we had run into this error before.. since this is an impl on a StagedReference specifically we had discussed a necessary check to make sure the TemporaryId is referencing a correct variant (in this case Staged)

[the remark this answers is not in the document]
evomimic: The check has been removed from get_rc_holon.

PR #278, opened 2025-07-09
tests passing

dauphin3:


evomimic:
I added implementations of Added Deref / DerefMut for TransientHolonPool and StagedHolonPool to eliminate need for ".0" nomenclature in Nursery and TransientHolonManager. Reduced number of changes for this PR and sets us up with better ergonomics going forward. After this change, I confirmed all tests passed.

Other than that, everything looked good. Merging/closing this PR.

PR #283, opened 2025-07-25
stub validation

dauphin3:


evomimic:
- I was able to checkout branch 281, compile and test it. Sweetests all passed.
- Reviewed code changes and committed one additional set of changes:
  - Add doc comment to holons_integrity's lib.rs clarifying its responsibilities (including mapping to non-HC dependent types)
  - Moved definition of ValidationError to integrity_core_types and re-enabled HolonError mapping.  Since conventional practice is to use HTTP Status Code 422 to indicate semantic errors (like validation), I added that as a ResponseStatusCode and mapped ValidationError to it.

Confirmed that everything still compiles and tests and checks pass after this commit.

PR #289, opened 2025-07-30
[Enhancement]-introduce type_names crate

dauphin3:
-move RelationShipName to core_integrity_types
-unit tests for enum variant string conversion for each new type name

all tests passing


evomimic:
When we removed sweetest from the workspace, we had to change the package.json's sweet:test script to reference the sweetest crate....  A by-product of that script change is that ONLY the sweetests are being run. I've changed the "sweet:test" script to run both cargo test && cargo test --manifest-path tests/sweetests/Cargo.toml.

When I re-ran the tests, the unit tests failed for both relationship name and property names. 

```
running 4 tests
test holon_names::tests::test_variant_string_conversion ... ok
test value_names::tests::test_variant_string_conversion ... ok
test property_names::tests::test_variant_string_conversion ... FAILED
test relationship_names::tests::test_variant_string_conversion ... FAILED

failures:

---- property_names::tests::test_variant_string_conversion stdout ----

thread 'property_names::tests::test_variant_string_conversion' panicked at crates/type_system/type_names/src/property_names.rs:43:9:
assertion `left == right` failed
  left: PropertyName(MapString("AllowsDuplicates"))
 right: PropertyName(MapString("allows_duplicates"))

---- relationship_names::tests::test_variant_string_conversion stdout ----

thread 'relationship_names::tests::test_variant_string_conversion' panicked at crates/type_system/type_names/src/relationship_names.rs:43:9:
assertion `left == right` failed
  left: RelationshipName(MapString("ComponentOf"))
 right: RelationshipName(MapString("COMPONENT_OF"))
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

```

The problem in both cases is in the unit test, not the code being tested. 

@dauphin3 : 

- [ ] For unit tests... rather than explicitly checking every enum variant... just check a representative subset of them
- [ ] Amend the property name unit tests to expect snake_case
- [ ] Amend the relationship name unit tests to expect SCREAMING_SNAKE_CASE

evomimic:
I decided to slightly re-work our package.json scripts:
* reverted "sweet:test" to just run the sweetests and not cargo test. This allows us to run those tests selectively.
* added a "all-tests" command that runs both cargo test and sweetests.

### We should use `npm run all-tests` as our "definition of done" tests going forward.

Everything else looks good. All tests pass.

PR #294, opened 2025-08-07
Ergonomic Relationship Access and Mutation Using Core Enums

dauphin3:


crates/holon_dance_builders/src/add_related_holons_dance.rs
```diff
@@ -7,7 +7,7 @@ use integrity_core_types::RelationshipName;
 /// Builds a DanceRequest for adding related holons to a source_holon.
 pub fn build_add_related_holons_dance_request(
     staged_reference: StagedReference,
-    relationship_name: RelationshipName,
+    relationship_name: MapString,
```
evomimic, line 10: Why downcast this signature and give up type safety?

crates/type_system/type_names/src/relationship_names.rs
evomimic, line 1: We should also support pass-through of RelationshipName (i.e., don't penalize code that ALREADY has a RelationshipName):

impl ToRelationshipName for RelationshipName {
    #[inline]
    fn to_relationship_name(self) -> RelationshipName { self }
}

impl ToRelationshipName for &RelationshipName {
    #[inline]
    fn to_relationship_name(self) -> RelationshipName { self.clone() }
}

evomimic:
I've confirmed all tests pass. Please make requested code changes and retest.

PR #335, opened 2025-11-03
272 - Holon Loader: two-pass holon import + guest dance adapter + incremental tests

owleyeview:
## Summary

This PR introduces the **Holon Data Loader** end-to-end path and a minimal **guest-side dance adapter** to invoke it from tests/clients. It also adds an **incremental test fixture** that exercises declared links and lays groundwork for inverse mapping.

**Highlights**

* Loader controller two-pass flow: **Pass-1 staging**, **Pass-2 relationship resolution**, **commit-if-clean**.
* Resolver writes declared links and attaches detailed **error holons** when Pass-2 fails.
* Guest-side dance adapter calls `load_bundle()`; TrustChannel path wired for tests.
* Incremental fixture consolidates steps: **empty bundle → nodes-only → declared link** (passing).
* Test helpers for robust diagnostics (response dump + printing `HasLoadError` holons).
* Relationship helper consolidated into a single:

  * `add_loader_relationship_reference(..., Declared|Inverse, ...)`.

## Design notes

* **Pass-1**: map `LoaderHolon` → staged instance; short-circuit if empty bundle.
* **Pass-2**:

  * Resolve endpoints: **Nursery by key** (staged in same bundle).
  * Write declared edges; write inverse edges by mapping to declared name/direction (Tier-1 micro-schema in progress).
  * Accumulate resolver errors; **abort commit** if any.
* **Response**: returns a `HolonLoadResponse` transient with:

  * `response_status_code`, `holons_staged`, `holons_committed`, `links_created`, `error_count`, `dance_summary`
  * `HasLoadError` → `HolonError` holons for concrete diagnostics.

## Testing

* `shared_test/test_load_holons.rs`:

  * Asserts response properties; always dumps response + error holons on mismatch.
* `shared_test/fixtures/load_holons_fixture.rs` (**single incremental fixture**):

  1. Empty bundle ⇒ `UnprocessableEntity`, counts unchanged.
  2. Nodes-only (3) ⇒ `OK`, `LinksCreated = 0`.
  3. Declared link (Book → Person) ⇒ `OK`, `LinksCreated = 1`.
* Constants (sample data) updated for clarity:

  * `BOOK_KEY`, `PERSON_1_KEY`, `PERSON_2_KEY`, `PUBLISHER_KEY`
  * Type/relationship descriptor keys: `BOOK_DESCRIPTOR_KEY`, `PERSON_DESCRIPTOR_KEY`
  * Relationship names: `AuthoredBy`, inverse `Authors`
* Test command I'm using to shop all tracing messages:

```
RUST_LOG=info,holons_loader=trace,holons_core=debug WASM_LOG=info,holons_loader=trace,holons_core=debug cargo test --manifest-path tests/sweetests/Cargo.toml     -- --nocapture
```

## Known limitations / open questions (deferred to follow-ups)

* **Inverse LRR resolution**: Current tests pass with declared links. For inverse, the test environment doesn’t provide stable ids and Nursery is cleared on commit. We’ll stage **both endpoints in the same bundle** for Tier-1 until we decide on saved-by-key fallback.

* **TypeDescriptor micro-schema**: Add a tiny builder to load:

  * `Book.HolonType`, `Person.HolonType`
  * `(Book.HolonType)-[AuthoredBy]->(Person.HolonType)` and inverse `(Person.HolonType)-[Authors]->(Book.HolonType)` wired with `InverseOf`
* **Duplicate link attempts**: We currently dedupe at write edge-key level; document final semantics.
* **Commit atomicity**: Pass-2 errors abort before commit; link writes are all-or-nothing for a clean run.

## Known Issues
* The test_print_database step currently panics and seems like it is trying to call an unimplemented client holon service function.

### DIFF
--- a/crates/holons_loader/src/controller.rs
+++ b/crates/holons_loader/src/controller.rs
@@ -10,7 +10,7 @@
 // This controller keeps only per-call, in-memory state (no cross-call persistence).
 // It is intentionally thin: it wires together Mapper → Resolver → Commit → Response.
 
-use tracing::{debug, info};
+use tracing::{debug, info, warn};
 // use uuid::Uuid;
 
 use holons_prelude::prelude::*;
@@ -58,45 +58,77 @@ impl HolonLoaderController {
         // (If/when keyless or extra targets appear, have the mapper return exact staged_count.)
         let staged_count = mapper_output.staged_count;
 
-        // If Pass 1 produced any errors or the bundle was empty,
-        // build the response now and return (skip Pass 2 & commit).
+        // Extract Pass-1 errors
         let mapper_errors = std::mem::take(&mut mapper_output.errors);
-        let is_empty_bundle =
-            staged_count == 0 && mapper_output.queued_relationship_references.is_empty();
-        if !mapper_errors.is_empty() || is_empty_bundle {
-            debug!(
-                "HolonLoaderController::load_bundle - early return: {}",
-                if !mapper_errors.is_empty() { "pass1 errors" } else { "empty bundle" }
+
+        // ─────────────────────────────────────────────────────────────────────
+        // PASS 1: handle early-exit conditions (errors or empty bundle)
+        // ─────────────────────────────────────────────────────────────────────
+
+        let error_count = mapper_errors.len() as i64;
+        let is_error_case = error_count > 0;
+
+        // ─────────────────────────────────────────────────────────────────────
+        // CASE 1: Pass-1 errors detected
+        // ─────────────────────────────────────────────────────────────────────
+        if is_error_case {
+            warn!(
+            "HolonLoaderController::load_bundle - early return due to Pass 1 errors ({} detected)",
+            error_count
+        );
+
+            // Prefer typed error holons; bubble up if system-level failure
+            let error_holons = make_error_holons_best_effort(context, &mapper_errors)?;
+
+            let summary = format!(
+                "Pass 1 reported {} error(s). Pass 2 and commit were skipped.",
+                error_count
             );
-
-            // Build error holons (prefer typed; fallback to untyped if descriptor missing)
-            let error_holons = make_error_holons_best_effort(context, &mapper_errors)
-                .unwrap_or_else(|e| {
-                    debug!("Failed to build error holons (pass1); proceeding without: {}", e);
-                    Vec::new()
-                });
-
-            let summary = if !mapper_errors.is_empty() {
-                format!(
-                    "Pass 1 reported {} error(s). Pass 2 and commit were skipped.",
-                    error_holons.len()
-                )
-            } else {
-                "Empty bundle: no LoaderHolons found; nothing to process.".into()
-            };
 
             let response_reference = self.build_response(
                 context,
                 run_id,
                 staged_count,
-                0,
-                0,
-                error_holons.len() as i64,
+                0,           // holons_committed
+                0,           // links_created
+                error_count, // always use real error count, not holon count
                 summary,
                 error_holons,
             )?;
 
-            debug!("HolonLoaderController::load_bundle - done (pass1 short-circuit)");
+            warn!("HolonLoaderController::load_bundle - done (aborted after Pass 1)");
+            return Ok(response_reference);
+        }
+
+        // ─────────────────────────────────────────────────────────────────────
+        // CASE 2: Empty bundle (no holons and no relationships)
+        // ─────────────────────────────────────────────────────────────────────
+        // Clarify “empty bundle” semantics for readability and extensibility
+        let no_staged_holons = staged_count == 0;
+        let no_relationships = mapper_output.queued_relationship_references.is_empty();
+        let is_empty_bundle = no_staged_holons && no_relationships;
+
+        if is_empty_bundle {
+            info!(
+            "HolonLoaderController::load_bundle - early return (empty bundle: no holons, no relationships)"
+        );
+
+            let summary =
+                "Empty bundle: no LoaderHolons or relationship references found; nothing to process."
+                    .to_string();
+
+            let response_reference = self.build_response(
+                context,
+                run_id,
+                staged_count,
+                0, // holons_committed
+                0, // links_created
+                0, // errors_encountered
+                summary,
+                Vec::new(), // no error holons
+            )?;
+
+            info!("HolonLoaderController::load_bundle - done (empty bundle short-circuit)");
             return Ok(response_reference);
         }
 
@@ -114,13 +146,16 @@ impl HolonLoaderController {
 
         // If Pass 2 produced any errors, build the response now and return (skip commit).
         if !resolver_errors.is_empty() {
-            debug!(
-                "HolonLoaderController::load_bundle - pass2 errors, short-circuit before commit"
-            );
+            let resolver_error_count = resolver_errors.len() as i64;
+
+            warn!(
+            "HolonLoaderController::load_bundle - pass2 errors ({}), short-circuit before commit",
+            resolver_error_count
+        );
 
             let error_holons = make_error_holons_best_effort(context, &resolver_errors)
                 .unwrap_or_else(|e| {
-                    debug!("Failed to build error holons (pass2); proceeding without: {}", e);
+                    warn!("Failed to build error holons (pass2); proceeding without: {}", e);
                     Vec::new()
                 });
 
@@ -128,17 +163,17 @@ impl HolonLoaderController {
                 context,
                 run_id,
                 staged_count,
-                0,
+                0, // holons_committed
                 links_created,
-                error_holons.len() as i64,
+                resolver_error_count,
                 format!(
                     "Pass 2 reported {} error(s). Commit was skipped. {} holons staged; 0 committed; {} links attempted.",
-                    error_holons.len(), staged_count, links_created
+                    resolver_error_count, staged_count, links_created
                 ),
                 error_holons,
             )?;
 
-            debug!("HolonLoaderController::load_bundle - done (pass2 short-circuit)");
+            warn!("HolonLoaderController::load_bundle - done (pass2 short-circuit)");
             return Ok(response_reference);
         }
 
@@ -166,7 +201,7 @@ impl HolonLoaderController {
             staged_count,
             holons_committed,
             links_created,
-            0,
+            0, // errors_encountered
             if commit_ok {
                 format!(
                     "{} holons staged; {} committed; {} abandoned; {} attempts.",
--- a/crates/holons_loader/src/errors.rs
+++ b/crates/holons_loader/src/errors.rs
@@ -3,7 +3,7 @@
 use holons_prelude::prelude::CorePropertyTypeName::{ErrorMessage, ErrorType};
 use holons_prelude::prelude::*;
 use std::sync::atomic::{AtomicU32, Ordering};
-use tracing::warn;
+use tracing::{error, warn};
 
 // Global counter for generating unique error holon keys
 static ERROR_SEQ: AtomicU32 = AtomicU32::new(1);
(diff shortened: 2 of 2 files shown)


crates/holons_loader/src/controller.rs
```diff
@@ -0,0 +1,315 @@
...
+            info!("HolonLoaderController::load_bundle - done (pass2 short-circuit)");
+            return Ok(response_reference);
+        }
+
+        // ─────────────────────────────────────────────────────────────────────
+        // COMMIT: persist all staged holons (only if both phases succeeded)
+        // ─────────────────────────────────────────────────────────────────────
+        info!("HolonLoaderController::load_bundle - commit");
+
+        let commit_response = commit(context)?;
```
evomimic: just to clarify what "commit" is being called here... precede this line with a comment:
// commit(): provided by HolonOperationsApi via holons_prelude

crates/holons_loader/src/controller.rs
```diff
@@ -0,0 +1,315 @@
...
+use holons_prelude::prelude::*;
+
+use crate::errors::make_error_holons_best_effort;
+use crate::{LoaderHolonMapper, LoaderRefResolver, ResolverOutcome};
+
+pub const CRATE_LINK: &str = "I like loading holons with HolonsLoader!"; // temporary const to link crate to test crate
+
+/// HolonLoaderController: top-level coordinator for the loader pipeline.
+#[derive(Debug, Default)]
+pub struct HolonLoaderController {
```
evomimic: Only staged_count is ever used. committed_count and error_count remain unset and unused.
Suggestion: Remove or wire them up before merge to avoid confusion about controller state.

crates/holons_loader/src/controller.rs
```diff
@@ -0,0 +1,315 @@
...
+        let is_empty_bundle =
+            self.staged_count == 0 && mapper_output.queued_relationship_references.is_empty();
+        if !mapper_errors.is_empty() || is_empty_bundle {
+            info!(
+                "HolonLoaderController::load_bundle - early return: {}",
+                if !mapper_errors.is_empty() { "pass1 errors" } else { "empty bundle" }
+            );
+
+            // Build error holons (prefer typed; fallback to untyped if descriptor missing)
+            let error_holons = make_error_holons_best_effort(context, &mapper_errors)?;
```
evomimic: If constructing typed error holons fails, the loader returns Err(HolonError) and the caller receives no response object.

✅ Suggestion: Catch that failure and fall back to a minimal HolonLoadResponse with a plain text summary, so the guest always receives a usable transient response.

crates/holons_loader/src/controller.rs
```diff
@@ -0,0 +1,315 @@
...
+        // If Pass 2 produced any errors, build the response now and return (skip commit).
+        if !resolver_errors.is_empty() {
+            info!("HolonLoaderController::load_bundle - pass2 errors, short-circuit before commit");
+
+            let error_holons = make_error_holons_best_effort(context, &resolver_errors)?;
+
+            let response_reference = self.build_response(
+                context,
+                run_id,
+                MapString("UnprocessableEntity".into()),
```
evomimic: This function should not deal with ResponseStatusCodes at all. HolonErrors are automatically mapped to the ResponseStatusCode enum by the dancer upon return. The response should return HolonError if there were errors, or response otherwise.

crates/holons_loader/src/loader_holon_mapper.rs
```diff
@@ -0,0 +1,182 @@
...
+        // Direct traversal from LoaderHolon → LoaderRelationshipReference entries.
+        let relationship_name = HasRelationshipReference.as_relationship_name();
+        let collection_handle = loader.related_holons(context, &relationship_name)?;
+
+        // Lock the collection for read, then clone members so we can release the lock early.
+        let member_refs = {
+            let collection_guard = collection_handle.read().map_err(|_| {
+                HolonError::FailedToBorrow("HolonCollection read lock poisoned".into())
+            })?;
+            collection_guard.get_members().clone()
```
evomimic: clone not required... bundle is frozen (read-only), so no problem holding read lock while iterating through the loop.

crates/holons_loader/src/errors.rs
evomimic: ### Improved Error Handling (can be deferred to later PR)

1. The core schema `HolonErrorType's` , `ErrorType` PropertyType should have an Enum value type (instead of StringValue type). Its variants should correspond to the Rust HolonError's enum variants.

2. And the `HolonLoaderSchema` should define a `LoaderError` HolonType that extends `HolonErrorType`. It should declare an `ErrorWith` relationship to `LoaderHolon` that identifies which LoaderHolon caused this error. 

3. `LoaderHolon` should be extended with the following properties:
      - SourceStartByte: MapInteger,
      - SourceEndByte:  MapInteger)

4. `HolonLoaderBundle` should add a `SourceFile: MapString` property type.

Enhancements 2-4 allow correspondence between discovered errors and the position in the source file containing the LoaderHolon definition for which the error was discovered.

tests/sweetests/tests/shared_test/test_load_holons.rs
evomimic: Overall this looks good. However, it exposes a problem with our API surface in that it puts responsibility for getting and dropping locks with callers. This is rife with potential problems. 

We should push lock acquisition down into the TransientHolonManager. For example, it knows that `empty_holon` needs to insert the new holon in its holon pool. So it can lock the pool immediately before the insert and drop it immediately after. Callers shouldn't have to manage the lock semantics.

This should be addressed in a separate Issue/PR... correcting this is out of scope for the current PR.

crates/holons_loader/src/loader_ref_resolver.rs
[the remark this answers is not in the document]
owleyeview: I've addressed points 2-4

crates/holons_loader/src/controller.rs
```diff
@@ -0,0 +1,269 @@
...
+        let mapper_errors = std::mem::take(&mut mapper_output.errors);
+        let is_empty_bundle =
+            staged_count == 0 && mapper_output.queued_relationship_references.is_empty();
+        if !mapper_errors.is_empty() || is_empty_bundle {
+            debug!(
+                "HolonLoaderController::load_bundle - early return: {}",
+                if !mapper_errors.is_empty() { "pass1 errors" } else { "empty bundle" }
+            );
+
+            // Build error holons (prefer typed; fallback to untyped if descriptor missing)
```
evomimic: This could risk returning a  "false pass" if empty error vec is interpreted as meaning... there were no reported errors. A better approach would be for `make_error_holons_best_effort` not to return a `Result` (i.e., make a _better_ best effort!).

crates/holons_loader/src/errors.rs
```diff
@@ -0,0 +1,195 @@
...
+        InvalidHolonReference(_) => "invalid_holon_reference",
+        NotAccessible(_, _) => "not_accessible",
+
+        // Group the rest under a generic bucket to avoid leaking internals
+        _ => "server_error",
+    }
+}
+
+/// Build transient HolonErrorType holons for reporting load errors.
+pub fn make_error_holons_best_effort(
```
evomimic: ### System-level failures should not be masked as “best effort”

The current implementation of `make_error_holons_best_effort` quietly swallows `HolonError`s from the transient layer (e.g., lock failures or `create_empty` errors) by logging a `warn!` and continuing. That makes sense for *domain-level* issues (like a malformed error holon descriptor), but not for *system-level* failures that indicate the loader can no longer function correctly.

Because the only way to create a `TransientReference` is via the `TransientBehaviorService`, any failure there means we’ve lost the ability to represent **any** errors at all. Logging and returning `Ok(Vec::new())` in that case risks producing a “false pass,” where `load_bundle` reports success even though no response could be built.

✅ **Suggested change**
Keep `make_error_holons_best_effort` returning a `Result`, but treat service-level errors as fatal:
- Attempt typed → fallback to untyped → if *that* fails, `return Err(e)` instead of `warn!` and continue.  
- Let `load_bundle` handle the propagated `Err(HolonError)` so that systemic failures abort properly.  
- Base the loader’s error counts on the number of detected domain errors (`mapper_errors.len()`), not on the number of error holons that happened to be materialized.

This preserves the intended separation between recoverable “data” errors and unrecoverable “system” errors, ensuring `load_bundle` never reports a false positive.

crates/holons_loader/src/controller.rs
```diff
@@ -0,0 +1,269 @@
...
+        info!("HolonLoaderController::load_bundle - pass1_stage");
+
+        let mut mapper_output = LoaderHolonMapper::map_bundle(context, bundle)?;
+        // For now we approximate staged_count by the number of staged targets created in Pass 1.
+        // (If/when keyless or extra targets appear, have the mapper return exact staged_count.)
+        let staged_count = mapper_output.staged_count;
+
+        // If Pass 1 produced any errors or the bundle was empty,
+        // build the response now and return (skip Pass 2 & commit).
+        let mapper_errors = std::mem::take(&mut mapper_output.errors);
```
evomimic: ### ⚠️ **Comment: Compute `error_count` from actual mapper errors before materialization**

Right after `let mapper_errors = std::mem::take(&mut mapper_output.errors);`,  
we should capture the domain error count explicitly:

```rust
let error_count = mapper_errors.len() as i64;
```

This preserves the true number of validation or mapping errors detected during Pass 1.  
Later, when building the summary or response, we should pass `error_count` directly instead of computing it from `error_holons.len()`, which only reflects *successful* transient holon creation.

Then, further down (around line 72), allow system-level failures in  
`make_error_holons_best_effort` to propagate normally:

```rust
let error_holons = make_error_holons_best_effort(context, &mapper_errors)?;
```

✅ **Benefits**
- Avoids undercounting errors when holon creation fails.  
- Prevents masking serious transient service failures.  
- Keeps `error_count` tied to actual validation results rather than incidental side effects.

crates/holons_loader/src/errors.rs
```diff
@@ -0,0 +1,195 @@
...
+            }
+        }
+    }
+    Ok(out)
+}
+
+/// Build a transient HolonError holon with {error_type, error_message} **and**
+/// set its descriptor to `HolonErrorType`.
+/// Caller can attach it to the response via REL_HAS_LOAD_ERROR.
+pub fn make_error_holon_typed(
```
evomimic: ### 💡 **Comment: Consolidate typed/untyped variants into a single helper**

`make_error_holon_typed` and `make_error_holon_untyped` differ only by whether a descriptor is applied:

```rust
transient_reference.with_descriptor(context, holon_error_type_descriptor)?;
```

Everything else — creating the transient holon and populating fields — is identical.  
To reduce duplication and future maintenance overhead, consider merging them into one function:

```rust
pub fn make_error_holon(
    descriptor: Option<HolonReference>,
    err: &HolonError,
) -> Result<TransientReference, HolonError> {
    let mut transient_reference = create_empty_error_holon(context)?;
    if let Some(desc) = descriptor {
        transient_reference.with_descriptor(context, desc)?;
    }
    populate_error_fields(context, &mut transient_reference, err)?;
    Ok(transient_reference)
}
```

✅ **Benefits**
- Eliminates redundant code paths and doc comments.
- Makes descriptor application an explicit, optional behavior.
- Keeps all error-holon creation logic in one predictable place.

crates/holons_loader/src/loader_ref_resolver.rs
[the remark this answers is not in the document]
evomimic: Added resolution of #1 to Issue #339

crates/holons_loader/src/loader_holon_mapper.rs
```diff
@@ -0,0 +1,182 @@
...
+        let loader_holon_members = {
+            let loader_holon_collection_handle =
+                bundle.related_holons(context, BundleMembers.as_relationship_name())?;
+
+            let loader_holon_collection_guard =
+                loader_holon_collection_handle.read().map_err(|_| {
+                    HolonError::FailedToBorrow("HolonCollection read lock poisoned".into())
+                })?;
+
+            loader_holon_collection_guard.get_members().clone()
```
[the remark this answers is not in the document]
evomimic: Added this to Issue #339 , so resolving here.

crates/holons_loader/src/loader_holon_mapper.rs
evomimic: will be fixed in Issue #339

evomimic (commented):
Review complete... all prior conversations marked resolved. To accelerate the merge, I fixed the new issues (reported) and tested and committed these changes. Some issues were deferred to Issue #339 for cleanup in a separate PR.

crates/holons_loader/src/errors.rs
evomimic: fixed by reviewer

crates/holons_loader/src/controller.rs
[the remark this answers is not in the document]
evomimic: fixed by reviewer

crates/holons_loader/src/errors.rs
evomimic: will be fixed in Issue #339

Issue #334, opened 2025-11-03
Multi-Bundle Load Holons Dance


evomimic:
### 1. Summary (Required)

**What is the enhancement?**  
Extend the LoadHolons dance to accept and process a collection of HolonLoaderBundles (one bundle per JSON input file) rather than a single bundle for the entire load set. Each bundle records file metadata (e.g., filename, size, optional hash) and enables precise, per-file error reporting (including byte offsets and holon start positions) while preserving the requirement that the *entire multi-file load set* is staged and committed atomically.

---

### 2. Problem Statement (Required)

**Why is this needed?**  
Today, multiple JSON files are parsed and loaded as a single HolonLoaderBundle. This obscures which file (and where within it) triggered an error and makes troubleshooting harder. A 1:1 file→bundle mapping improves ergonomics and debuggability by anchoring loader diagnostics to a specific file and byte location. We must introduce this without changing the atomic, all-or-nothing semantics across the full load set.

---

### 3. Dependencies (Required)

**Does this depend on other issues or features?**  
- ✅ #270 Meta-Descriptor Driven Type Loader  
- ✅ #292 JSON Loader Client  
- ✅ #272 Holon Data Loader  
- ⛳️ #293 LoadHolons Dance (baseline single-bundle version)  
- ⛳️ #333 Holon Loader Launcher + Rust Client Parser (ensures per-file parsing context is available)  

No additional blockers, but this enhancement should build on the single-bundle dance path introduced in #293.

---

### 4. Proposed Solution (Required)

**How would you solve it?**

1. **Request/Response Shape**  
   - **RequestBody**: TransientReference → `HolonLoadSet` with a CONTAINS relationship to multiple `HolonLoaderBundle` holons (one per input file).  
   - **ResponseBody**: HolonReference → HolonLoadResponse holon summarizing load results across all bundles, including:  
     - total_bundles, total_loader_holons, committed_holons  
     - errors: array of HolonLoadError with file-scoped metadata (filename, byte_offset_start, byte_offset_end, loader_holon_key, message, rule_id, severity).

2. **Loader Semantics (unchanged atomicity, expanded scope)**  
   - **Two-pass load remains** (Pass 1 stage properties & queue relationships; Pass 2 resolve & commit).  
   - All bundles are aggregated into a single staging context to allow cross-file references.  
   - **Atomic commit across all bundles**: if any error occurs, no holons commit; otherwise commit all.

3. **Holon Types & Metadata**  
 
[...]

PR #341, opened 2025-11-14
334 multi bundle load holons dance

owleyeview:
## **Draft PR Description — Multi-Bundle Holon Loader**

This PR implements the **multi-bundle HolonLoadSet loader** defined in issue #334.
It replaces the previous single-bundle loader and enables loading *multiple JSON files atomically* with precise, per-file error attribution.

### **What’s Included**

* **HolonLoadSet support**
  The `load_holons` dance now accepts a `HolonLoadSet` that **CONTAINS** one or more `HolonLoaderBundle` holons (1 bundle per input file).
* **File-level provenance**

  * Each bundle carries a required `filename`.
  * Each LoaderHolon may include `start_utf8_byte_offset` to pinpoint error locations.
  * Errors surfaced in the `HolonLoadResponse` include:

    * filename
    * loader_holon_key
    * message
    * byte offset (if provided)
* **Cross-bundle staging + resolution**
  All bundles in the set are staged into a unified Pass-1 context, so `$ref`-style lookups work across files.
* **Atomic commit preserved**
  Any Pass-1 or Pass-2 error halts the load; nothing is committed.
* **Duplicate key detection across bundles**
  The controller detects when multiple files define LoaderHolons with the same key and emits enriched `DuplicateError` holons (via the new `ErrorWithContext` path).
* **Updated HolonLoadResponse**
  Adds `total_bundles` and `total_loader_holons` for visibility when working with larger sets.
* **Revised sweetests and fixture suite**
  The incremental loader fixture now covers:

  * Empty sets
  * Node-only bundles
  * Declared relationship bundles
  * Inverse resolution with inline micro-schema
  * Cross-bundle declared relationships
  * **Multi-bundle duplicate-key failure** (with filename + byte-offset provenance)

### **What’s *Not* Included**

* **Client-side HolonLoadSet builder**
* **JSON Parser/Launcher that emits bundles + offsets**

These will be addressed in a separate issue (requested from the architect).

### **Impact**

* Improves diagnostics, DX, and UI potential without changing loader semantics.
* No backward compatibility needed; callers must now send HolonLoadSet.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,6 +32,7 @@ Cargo.lock
 # OS / IDE cruft
 .DS_Store
 .idea/
+.venv/
 .vscode/
 .running
 settings.json
--- a/crates/holons_loader/src/controller.rs
+++ b/crates/holons_loader/src/controller.rs
@@ -19,8 +19,6 @@ use holons_prelude::prelude::*;
 use crate::errors::{make_error_holons_best_effort, ErrorWithContext};
 use crate::{LoaderHolonMapper, LoaderRefResolver, ResolverOutcome};
 
-pub const CRATE_LINK: &str = "I like loading holons with holons_loader!"; // temporary const to link crate to test crate
-
 /// Local structure to hold LoaderHolon provenance used for error enrichment.
 #[derive(Debug, Clone)]
 pub struct FileProvenance {
--- a/crates/holons_loader/src/lib.rs
+++ b/crates/holons_loader/src/lib.rs
@@ -4,6 +4,5 @@ pub mod loader_holon_mapper;
 pub mod loader_ref_resolver;
 
 pub use controller::HolonLoaderController;
-pub use controller::CRATE_LINK; // temporarily exporting a constant to link to test crate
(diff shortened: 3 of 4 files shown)


evomimic (approved):
✅ Approved — everything looks great!

I went ahead and removed the unused CRATE_LINK constant and the `_build_anchor_holons_loader()` function from the test code.

I'm guessing these were originally added to ensure the holons_loader crate was linked into the test orchestrator binary — likely to force a rebuild of guest code (since the test code doesn’t directly depend on the guest zome, and changes might otherwise be missed).

However:

- Although _build_anchor_holons_loader() referenced CRATE_LINK, the function was never actually called — and in Rust, unused functions (and their dependencies) don’t affect the build output.
- So the CRATE_LINK reference had no impact — the loader crate wasn’t being pulled in at all.
- After removing both, a full clean, build:happ, and sweetest run confirmed that guest code changes still trigger correct recompilation and all tests pass.

Conclusion: the workaround was inert and is no longer necessary — safe to remove for now. We can reintroduce a more explicit trigger (or doc the pattern) if rebuild detection ever becomes flaky again.

Thanks for the solid implementation!

PR #349, opened 2025-11-22
339 cleanup some guest side holonloader issues

owleyeview:
### **Summary**

This PR refactors the transient and staged holon management layers to strengthen our concurrency boundaries, reduce panic paths, and align resolver logic with the holon operations API. These changes improve internal ergonomics, and remove redundant locking responsibilities from call sites.

### **Key Changes**

* Updated relationship-name property constants used in commit routines.
* Updated `HolonSpaceManager` function signatures and revised `load_holons_fixture` call sites to match the new interfaces.
* Refactored `load_holons_fixture` to use the **holon_operations_api** rather than directly accessing transient behavior services.
* Removed all remaining `unwrap()` and `expect()` calls from **Nursery** and **TransientHolonManager**, ensuring that lock failures now surface as `HolonError` instead of panicking.
* Updated `commit` function documentation to correctly describe the new **CommitResponseType holon** returned as a `TransientReference`.
* Adjusted `LoaderRefResolver` to prefer the holon operations API (e.g., `get_staged_holons_by_base_key`) instead of reaching directly into staging services.

### **Follow-Up Work (Future Issue)**

`commit_functions.rs` still contains several `.unwrap()` / direct lock acquisitions inside the guest commit path. These should probably be replaced with proper `HolonError::FailedToAcquireLock` mappings to ensure consistent error handling and to eliminate remaining panic risks. I can open a follow-up issue proposing this cleanup.

### DIFF
--- a/crates/holons_core/src/core_shared_objects/nursery_access_internal.rs
+++ b/crates/holons_core/src/core_shared_objects/nursery_access_internal.rs
@@ -1,10 +1,6 @@
-use std::{
-    any::Any,
-    sync::{Arc, RwLock},
-};
+use std::any::Any;
 
-use super::{holon_pool::SerializableHolonPool, Holon};
-use crate::core_shared_objects::holon_pool::StagedHolonPool;
+use super::holon_pool::SerializableHolonPool;
 use crate::{HolonStagingBehavior, NurseryAccess, StagedReference};
 use base_types::MapString;
 use core_types::{HolonError, TemporaryId};


crates/holons_guest/src/guest_shared_objects/commit_functions.rs
evomimic: ### 🧹 Architectural Cleanup Request: Remove HolonPool Exposure From Commit Path

While reviewing the commit pipeline, I noticed that the current implementation
still exposes the internal `HolonPool` to `commit()` via
`NurseryAccessInternal::get_holon_pool()` and uses pool-level APIs such as
`get_staged_references()` directly on the pool.

This wasn’t explicitly called out in Issue 339, but it does represent a
layering leak that we should clean up now while we’re already touching the
commit path.

**Why this matters**

Exposing the concrete `HolonPool` (and especially its `Arc<RwLock<Holon>>`
elements) from the Nursery:

- leaks internal storage details into the commit logic,
- forces commit() to manage lock boundaries that should belong to Nursery,
- bypasses the reference-layer abstractions (`StagedReference`,
  `HolonReference`, etc.),
- makes future pool refactors more difficult,
- and is inconsistent with the direction of Issue 339: pushing all lock handling
  and pool interactions down into the staging/transient managers.

**Recommended cleanup**

I recommend updating this PR to:

1. **Expose `get_staged_references()` via `NurseryAccessInternal`**,  
   not directly from `HolonPool`.

2. **Stop passing `Arc<RwLock<StagedHolonPool>>` into `commit()`**.  
   The commit function should operate purely on `StagedReference`.

3. **Remove `HolonPool::get_all_holons()` and any remaining uses**,  
   since raw access to staged holons should no longer be needed.

This keeps the commit pipeline aligned with the reference-layer–only boundary
(e.g., `StagedReference`, `TransientReference`) and ensures that all lock
acquisition and internal pool structures remain properly encapsulated inside
the Nursery.

This wasn’t explicitly part of Issue 339, but it is a natural continuation of
the same architectural cleanup and it’s much easier to fix now, while the code
is already open for refactoring.

crates/holons_core/src/reference_layer/holon_operations_api.rs
evomimic: ### Move implementation of stage_new_from_clone and stage_new_version to Nursery

`stage_new_from_clone` and `stage_new_version` can be performed the same on either the client or guest. But currently, neither is offered by HolonStagingBehavior trait, so HolonOperationsApi delegates these calls to the HolonServiceApi. 

- [ ] add `stage_new_from_clone` and `stage_new_version` to HolonStagingBehavior trait
- [ ] move implementation of `stage_new_from_clone` and `stage_new_version` from guest_holon_service to the Nursery
- [ ] Drop `stage_new_from_clone` and `stage_new_version` from the HolonServiceApi trait
- [ ] Change implementation of HolonOperationsApi to delegate these calls to staging service instead of holon service.

evomimic (changes requested):

crates/holons_core/src/core_shared_objects/nursery_access_internal.rs
```diff
@@ -100,7 +92,13 @@ pub trait NurseryAccessInternal: NurseryAccess + HolonStagingBehavior + Send + S
     /// # Returns
     ///
     /// A `Vec<Arc<RwLock<Holon>>>` containing all staged Holons for thread-safe access.
-    fn get_holons_to_commit(&self) -> Vec<Arc<RwLock<Holon>>>;
+    fn get_holons_to_commit(&self) -> Result<Vec<Arc<RwLock<Holon>>>, HolonError>;
```
evomimic: This method should be deleted.

evomimic (approved):
I'll take care of deleting the unused get_holons_to_commit method from NurseryAccessInternal trait and nursery's implementation.

PR #357, opened 2025-12-15
306 dance test refactor

dauphin3:
It was discovered that the minting process for staging should clone the source TransientReference. FixtureHolons index by_key was changed to link a BaseKey to a list of TemporaryIds that each represent a cloned version from the original source and have a lineage of their own of TestReferences.
This allows us to track the expected ResolvedTestReference for a given id to the latest snapshot. In other words, anytime a dance would be creating a new holon at execution time, the fixture creates a new lineage for the base key of the holon it was cloned from.

For ergonomics, test steps and fixture holons are not needed, as the full coverage can only be expressed via the HolonReadable trait, and therefore an expected essential can be manually constructed and compared to an essential_content call on the reference itself without requiring the test harness.

Relationships for essential content are being ignored until a later determination.

### DIFF
--- a/tests/sweetests/src/harness/test_case/dance_test_language.rs
+++ b/tests/sweetests/src/harness/test_case/dance_test_language.rs
@@ -1,4 +1,41 @@
-// sweetests/src/harness/test_case/test_case.rs (excerpt)
+//! # Dance Test Language
+//!
+//! This module defines the **declarative language** used by MAP sweetests to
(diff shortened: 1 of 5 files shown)


evomimic (approved):
No further changes needed. 

For clarity, I renamed `test_case.rs` to `dance-test_language.rs `to better reflect its scope and purpose and added a file-level doc comment to that file.

PR #381, opened 2026-01-27
378 align with revised test framework design

dauphin3:
see disussion #376

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,7 +23,6 @@ workspaces/**/target/**/
 /host/conductora/node_modules/
 
 # Cargo and Rust build artifacts
-Cargo.lock
 .rustc_info.json
 *.rlib
 *.rmeta


tests/sweetests/src/harness/fixtures_support/fixture_holons.rs
evomimic: ### PR Feedback — FixtureHolons API Hardening & Canonical Source Resolution

This file is very close to the intended design, but a few visibility and API-shape issues make it too easy for adders to bypass the lifecycle rules we’ve been converging on—especially around delete semantics and head vs last-live resolution.

The fix is **not behavioral**, but **structural**: we should tighten the public surface so the *only* way adders derive a new source snapshot is via a single, intention-revealing helper.

---

## 1. Introduce a Canonical Public Helper for Adders (Fix for this PR)

### New required helper (public)

```rust
pub fn derive_next_source(
    &self,
    token: &TestReference,
) -> Result<SourceSnapshot, HolonError>
```

**Semantics (normative):**

- Given an input `TestReference`, return the correct `SourceSnapshot` for the *next* adder in the chain.
- Encodes the rule:
  - If the owning FixtureHolon’s head state is `Deleted` → use `last_live_snapshot`
  - Otherwise → use `head_snapshot`
- This is the **only** API adders should use to derive a new source snapshot.

**Rationale:**

- Collapses three currently scattered steps into one:
  - extract `SnapshotId` from token
  - locate owning `FixtureHolon`
  - choose head vs last-live snapshot
- Prevents adders from branching on lifecycle state themselves.
- Centralizes delete/commit semantics in one place.

All adders should replace any of the following patterns with this call:
- `source_token.expected_snapshot().as_source()`
- manual calls to `get_fixture_holon_by_snapshot`
- direct calls to `resolve_snapshot_as_source`

---

## 2. Tighten Visibility to Prevent Misuse (Fix in this PR)

The following APIs encode lifecycle and continuity rules and should **not** be callable by adders directly.

### Make private or `pub(crate)`

#### On `FixtureHolon`

- `resolve_snapshot_as_source`
  - This is an internal policy function.
  - Only `FixtureHolons` should decide how to resolve a source.

#### On `FixtureHolons`

- `get_fixture_holon_by_snapshot`
  - Adders should never obtain a `FixtureHolon`.
  - This enables exactly the kind of bypass logic we’re trying to prevent.

---

## 3. Commit Path Consistency (Fix in this PR)

Inside `commit()`, head mutation currently assigns directly:

- `holon.head_snapshot = expected`

**Use the accessor (advance_head) instead of operating on the struct field directly.**

This bypasses:
- snapshot registration
- last_live tracking
- future invariants

Commit should route through `advance_head`.
While not currently causing problems, we should fix this for this PR. Commit should not be the one place that breaks the rules we just worked so hard to define.


---

## 4. Intended Invariant Going Forward

> **Adders never inspect FixtureHolons.  
> They ask FixtureHolons for the next source snapshot, and nothing more.**

If we enforce this at the API and visibility level, we:
- eliminate head/last-live bugs
- reduce cognitive load for adder authors
- make future lifecycle extensions safer

---

### Summary of Requested Changes

- [ ] Add `derive_next_source(&TestReference)` as the sole public source-derivation API
- [ ] Update all adders to call this helper
- [ ] Restrict visibility of:
  - `FixtureHolon::resolve_snapshot_as_source`
  - `FixtureHolons::get_fixture_holon_by_snapshot`
- [ ] (Route commit head advancement through invariant-preserving logic

This is a structural hardening change that aligns the code with the design we’ve already agreed on and prevents accidental misuse going forward.

tests/sweetests/src/harness/test_case/dance_test_language.rs
evomimic: ### Source Snapshot Derivation in `dances_test_language` (Fix in this PR)

With the introduction of a unified helper on `FixtureHolons` to derive the **next source snapshot**, we should now standardize all adders to use it and eliminate ad-hoc source resolution logic.

#### New Rule (Normative)

> **Adders must never derive a `SourceSnapshot` themselves.**  
> They must obtain it via a single `FixtureHolons` helper (e.g. `derive_next_source(&TestReference)`).

This ensures:
- correct handling of `Deleted` vs live holons
- consistent use of `last_live_snapshot`
- no leakage of fixture-internal invariants into adders

---

### Required Changes in This File

#### ❌ Pattern 1 — `expected_snapshot().as_source()` (Incorrect)

This bypasses last-live resolution and must be replaced.

**Replace:**
```
let new_source = source_token.expected_snapshot().as_source();
```

**With:**
```
let new_source = fixture_holons.derive_next_source(&source_token)?;
```

**Affected adders:**
- `add_abandon_staged_changes_step`
- `add_remove_properties_step`
- `add_with_properties_step`

---

#### ⚠️ Pattern 2 — Manual FixtureHolon Lookup + `resolve_snapshot_as_source()`

This logic is correct but violates encapsulation and should be collapsed into the helper.

**Replace:**
```
let snapshot_id = source_token.expected_id();
let fixture_holon = fixture_holons.get_fixture_holon_by_snapshot(&snapshot_id)?;
let new_source = fixture_holon.resolve_snapshot_as_source();
```

**With:**
```
let new_source = fixture_holons.derive_next_source(&source_token)?;
```

**Affected adders:**
- `add_delete_holon_step`
- `add_stage_new_from_clone_step`
- `add_stage_new_version_step`

---

#### ✅ Pattern 3 — New Logical Holon Creation (Correct)

`add_new_holon_step` intentionally accepts a `TransientReference` and starts a new
TestReference chain.

No changes required.

---

## Why This Matters

- Centralizes delete semantics (head vs last-live) in **one place**
- Prevents subtle fixture-time bugs
- Makes adders mechanically correct and easier to review
- Preserves the invariant that **FixtureHolons owns continuity logic**

---

#### Review Heuristic

> If an adder derives a `SourceSnapshot` without calling `FixtureHolons`, it’s wrong.

Once these changes are applied, all adders will correctly honor delete semantics,
commit behavior, and future invariants without special-casing.

tests/sweetests/src/harness/execution_support/execution_reference.rs
evomimic: ## Simplification Required

The expected execution-time result of test step is fully carried in the fixture-created token for that step. But this file is currently still carrying **fixture intent and validation responsibilities that no longer belong in execution support**. This PR should simplify `execution_reference.rs` to represent **execution outcomes only**, with no fixture semantics or assertion helpers embedded.

Below are the concrete, definitive changes being recommended.

---

### 1. Remove `ResultingReference` and Collapse into `ExecutionReference`

**Current situation**

- `ExecutionReference` wraps:
  - `ExpectedSnapshot` (fixture intent)
  - `ResultingReference` (runtime outcome)
- `ResultingReference` is only used inside this file
- All external call sites conceptually deal with *execution-time resolution*, not “results” as a separate concept

**Recommendation**

- **Eliminate `ResultingReference` entirely**
- Fold its semantics directly into `ExecutionReference`
- Preserve the name **`ExecutionReference`** (preferred and already widely used)

**Why**

- There is only one runtime artifact that matters:   *“What does this fixture token resolve to at execution time?”*
- Delete is not a special case — it’s a valid execution-time outcome.
- This matches existing resolution APIs and mental models

---

### 2. Remove `ExpectedSnapshot` from `ExecutionReference`

**Current situation**
- `ExecutionReference` embeds `ExpectedSnapshot`
- This duplicates fixture intent that already lives in `TestReference`

**Recommendation**
- Remove `expected_snapshot` from `ExecutionReference` entirely

**Why**
- Execution support must be **outcome-only**
- Fixture intent is already available to executors via `TestReference`
- No execution-time resolution requires access to expected snapshot content

---

### 3. Move Validation Responsibility to a Separate Validation Module

Execution reference should record what happened, not judge whether that outcome “makes sense.”

The following methods should be removed:

- `ResultingReference::essential_content`
- `ExecutionReference::assert_essential_content_eq`

**Why**
- Keep the intent of this module pure -- simple lookup of ExecutionReference for a fixture token
- ExecutionReferences no longer contains information about the expected result of a step

At first glance, `essential_content()` appears harmless because it does **not** directly compare against expected state. However, it encodes an invalid semantic assumption that `Deleted` is an error condition rather than a valid execution outcome. `ExecutionReference` should _represent_ deletion, _not reject_ it.

### 4. Introduce Separate Validation Module

Introduce a new `results_validation.rs` module within the test harnesses `execution_support` module that offers a single standalone function: 

```
pub fn assert_expected_eq_actual(
    token: &TestReference,
    actual: &ExecutionReference,
) -> Result<(), HolonError>

```

Passing the TestReference makes validation token-centric rather than snapshot-centric, which matches how expectations are authored.

This function can:

- extract the ExpectedSnapshot from the TestReference, 
- use its expected state to determine whether an ExecutionReference of Deleted is an error or not.
- if both ExpectedSnapshot and ExecutionReference contain Live HolonReferences, extract those references and call the standard ReadableHolon's essential_content method on both.

This module should:

- Handle all lifecycle states correctly:
  - Transient
  - Staged
  - Saved
  - Deleted (as a valid outcome, not an error)
- Centralize:
  - essential content comparison
  - lifecycle state matching
  - future relationship comparisons
- Encode **framework-level invariants**, not step-specific semantics

---

### Why This Is the Right Shape

- Avoids duplicating validation logic across executors
- Keeps execution support purely representational
- Makes validation rules explicit, discoverable, and evolvable
- Allows future extensions (relationships, partial matches, tolerance rules)
  without touching executors or execution storage

evomimic:
## Executor Update Guidance (Applies to All Executors)

Other recommended changes for this PR introduce a **simplified and clarified execution model**. As a result, **all test step executors should be updated to follow the same pattern**. Rather than reviewing each executor individually, this comment defines the **new canonical executor contract**.

These changes are largely **mechanical and uniform**, so your IDE may be leveraged to make these changes across multiple executors at once.

---

## Summary of Changes Executors Must Adopt

### 1. ExecutionReference Is Outcome-Only

Executors must treat `ExecutionReference` as a **pure execution-time outcome**, not a carrier of fixture intent.

**Do not:**
- construct `ExecutionReference` from `ExpectedSnapshot`
- embed fixture semantics in execution support
- call validation methods on `ExecutionReference`

**Do:**
- construct `ExecutionReference` *only* from runtime results:
  - `Live(HolonReference)`
  - or `Deleted`

---

### 2. Validation Is Centralized via a Helper

Validation is no longer a method on `ExecutionReference`.

Executors **must** use the new helper:

```
assert_expected_eq_actual(
    context,
    &source_token,
    &execution_reference,
)?;
```

This helper:
- extracts expected state from the fixture token
- compares essential content when appropriate
- treats `Deleted` as a valid outcome
- encodes framework-level invariants uniformly

Executors should **not** reimplement or bypass this logic.

---

### 3. Canonical Executor Flow (Normative)

All executors should follow this structure:

1. **LOOKUP**
   - Resolve the source reference via `resolve_source_reference`

2. **BUILD / CALL**
   - Build and execute the dance request

3. **STATUS CHECK**
   - Assert `response.status_code == expected_status`

4. **ON SUCCESS**
   - Extract runtime `HolonReference` (if any)
   - Construct `ExecutionReference`
   - Call `assert_expected_eq_actual`
   - Perform any *step-specific* semantic assertions
   
5. **RECORD**
 - Always record using `state.record(&source_token, execution_reference)`
 - The harness derives s`ource_token.expected_id()` internally.
 - Recording is required for ALL steps, including:
   - expected failures
   - deletes
   - delete-after-delete 

- Skipping recording violates execution invariants.

---

### 4. What Executors Should *Not* Do Anymore

- ❌ Call `assert_essential_content_eq`
- ❌ Use `ResultingReference`
- ❌ Inspect or pattern-match on `ExpectedSnapshot`
- ❌ Treat `Deleted` as an error at execution time
- ❌ Embed fixture semantics in execution support

---

## Why This Matters

These changes:

- eliminate duplicated validation logic
- remove subtle deleted-state bugs
- align executors with commit + delete semantics
- make execution support append-only and side-effect free
- reduce executor code size and cognitive load

Most updates can be applied via **IDE-assisted refactoring** or **global search & replace**.

---

## Bottom Line

> Executors describe **what happened**.  
> The harness decides **whether that matches intent**.

All executors should be updated to reflect this separation.

tests/sweetests/tests/execution_steps/delete_holon_executor.rs
evomimic: ## Delete Executor — Additional Corrections Beyond General Guidance

Most changes required in `execute_delete_holon` are already covered by the **general executor guidance** (ExecutionReference simplification, centralized validation, recording via `TestReference`, etc.). However, there is **one delete-specific semantic issue** that should be corrected explicitly.

---

### 1. Recording Must Reflect the *Actual Outcome*, Not the Intent

**Current behavior**
- The executor **always** records:
  ```rust
  ExecutionReference::Deleted
  ```
  regardless of whether the delete request succeeded or failed.

**Why this is incorrect**
- Recording is meant to capture **what actually happened at runtime**
- `Deleted` is only correct if the delete dance returned `OK`
- If the delete request fails (and the failure was expected), the holon is still **live**
- Recording `Deleted` in that case corrupts execution-time resolution and future steps

**Required fix**
- Branch recording logic based on the **dance response**, not just `expected_status`

Conceptually:

- If `response.status_code == OK`
  - Record `ExecutionReference::Deleted`
- Else
  - Record `ExecutionReference::Live(HolonReference)` using the resolved source reference

This ensures:
- delete-after-delete tests remain expressible
- failed deletes preserve liveness correctly
- execution history reflects reality, not intent

---

tests/sweetests/tests/execution_steps/query_relationships_executor.rs
evomimic: ## Executor-Specific Issue: Missing Execution Recording

This executor **never records an execution outcome**, regardless of success or failure.

Unlike many other executors where recording happens conditionally, here there is **no call to `state.record(...)` at all**. This is a concrete omission that is **not just a refactor consequence** and should be fixed explicitly.

### Why This Is a Problem

- The harness requires **every executed step** to produce an execution outcome
- Subsequent steps may depend on this token being resolvable
- Skipping recording breaks append-only execution invariants
- It makes this step invisible to:
  - downstream resolution
  - final database verification
  - execution trace completeness

### Required Fix

After validating the response status, this executor must:

- Construct an `ExecutionReference` that reflects the *actual outcome*:
  - `Live(HolonReference)` if the query returns one
  - or another appropriate outcome type if defined
- Record it via:
  ```
  state.record(&source_token, execution_reference)
  ```

Even if this step is “read-only” or does not conceptually mutate state, **execution recording is still mandatory** so the token participates in the execution graph.

### Note

The existing `// TODO: need to match on expected content` is orthogonal and can be addressed later.  
The absence of execution recording should be corrected in this PR.

---

tests/sweetests/tests/execution_steps/stage_new_holon_executor.rs
evomimic: ## Executor Review — `execute_stage_new_holon`

Most changes needed in this executor are already covered by the **general executor guidance** (ExecutionReference simplification, centralized validation helper, revised `record` signature). However, there are **two executor-specific issues** that should be addressed here.

---

### 1. Executor Enforces a Precondition That Belongs to the Runtime

```rust
// Can only stage Transient
let transient_reference = match source_reference {
    HolonReference::Transient(tr) => tr,
    other => {
        panic!("{}", format!("expected lookup to return TransientReference, got {:?}", other));
    }
};
```

**Issue**

This executor is enforcing a semantic precondition (“can only stage Transient”) **before the dance is executed**.

That logic belongs in the **dance implementation**, not the test executor.

**Why this is a problem**

- Prevents writing tests that assert *runtime rejection* of invalid inputs
- Collapses “expected failure” into a fixture/executor panic
- Violates the rule that executors describe *what happened*, not *what should be allowed*

**Recommendation**

- Remove the explicit `TransientReference` match
- Pass the resolved `HolonReference` directly into the dance builder
- Let the dance return the appropriate failure status when the precondition is violated

This aligns with how delete-after-delete and other invalid-operation tests are intended to work.

---

### 2. Assertion and Recording Occur Even When Failure Is Expected

```rust
// 5. ASSERT the staged holon's content matches
...
resolved_reference.assert_essential_content_eq(context).unwrap();

// 6. RECORD
state.record(source_token.expected_id(), resolved_reference).unwrap();
```

**Issue**

Assertion and recording occur unconditionally, even when `expected_status != OK`.

**Why this matters**

- When a step is expected to fail, there is no valid “staged holon” to compare
- Recording a successful execution outcome in a failure case corrupts execution history
- This contradicts the clarified rule that **recorded outcome must reflect the actual dance result**

**Recommendation**

- Guard assertion and recording behind `if response.status_code == ResponseStatusCode::OK`
- For expected failures:
  - Do **not** assert essential content
  - Do **not** record a live execution reference

This is consistent with the refined rule:
> Recording reflects *what actually happened*, not what was expected.

---

### Summary

Beyond the general executor refactors, this executor specifically needs to:

1. **Remove executor-level enforcement of Transient-only staging**
2. **Condition assertion and recording on successful execution**

Fixing these keeps this executor aligned with:
- runtime-first validation
- correct failure testing
- append-only, truthful execution recording

tests/sweetests/tests/execution_steps/with_properties_command_executor.rs
evomimic: ## Executor Review — `execute_with_properties` (Correction)

Most updates required for this executor are already covered by the **general executor guidance** (ExecutionReference simplification, centralized validation helper, revised recording semantics). There is **one executor-specific correction** needed here, but it is *not* to conditionally skip recording.

---

### Assertion Is Conditional; Recording Is Not

```rust
// 5. ASSERT - essential content matches expected
...
resolved_reference.assert_essential_content_eq(context).unwrap();
...
state.record(source_token.expected_id(), resolved_reference).unwrap();
```

**Issue**

Assertion and recording are currently treated as inseparable, which conflates **validation** with **recording**.

---

### Correct Rule (Reasserted)

- **Assertions** (expected vs actual) must only occur when the dance **succeeds**
- **Recording** must occur **unconditionally**, but must reflect the *actual execution outcome*

---

### Why This Matters

- If `expected_response != OK`, there is no updated holon content to compare
- Performing essential-content assertions on a failed dance is invalid
- However, *not recording* the outcome would break append-only execution invariants
- Recording must reflect **what actually happened**, not what was expected

---

### Required Change

Refactor this executor to:

1. **Always record exactly one execution outcome**
2. **Condition validation only on success**

Illustrative structure:

```rust
let execution_reference = match response.status_code {
    ResponseStatusCode::OK => {
        // build ExecutionReference::Live from response HolonReference
    }
    _ => {
        // build ExecutionReference that reflects no mutation
        // (typically resolves to the prior live holon)
    }
};

// Only assert expected vs actual on success
if response.status_code == ResponseStatusCode::OK {
    assert_expected_eq_actual(context, &source_token, &execution_reference)?;
}

// Always record
state.record(&source_token, execution_reference)?;
```

---

### Summary

This executor must be updated to ensure:

- ❌ Assertions are **not** run on failed dances
- ✅ Recording **always happens**
- ✅ Recorded outcomes reflect **actual runtime behavior**, not expectations

All other required changes are already covered by the shared executor refactor guidance.

tests/sweetests/tests/fixture_cases/ergonomic_add_remove_properties_fixture.rs
evomimic: This test is correct as written, but is not really a "dance test" since it invokes no dances. Consider re-classifying (and moving ) this as a unit test.

tests/sweetests/tests/fixture_cases/stage_new_from_clone_fixture.rs
evomimic: ### 1. Unused Imports (Cleanup Required)

The following imports are not used anywhere in this fixture and should be removed to avoid noise and drift:

```rust
use base_types::{MapInteger};
use holons_core::{
    reference_layer::TransientReference,
    ReadableHolon,
    WritableHolon,
};
use holons_test::TestReference;
```

evomimic (changes requested):
## Overall PR Review Summary

This has been a **substantial and important refactor**, and I can now say  the **major structural changes are in place**. Nice work! Getting that foundation merged was the hard part.

With the core architecture settled, for this review I did a thorough "clean-up" review to get things into a clean final state before we move on to other issues. The goal was to make sure the codebase lands in a **clean, consistent, and maintainable state** before we move on to the next set of features.

As a result, you’ll see a relatively large number of comments. To help orient you:

- **Some comments identify changes that should be addressed in this PR**  
  These are mostly correctness, clarity, or consistency issues introduced or surfaced by the refactor.

- **Other comments explicitly propose follow-up Issues**  
  These are *not* expected to be resolved here. They’re captured so we don’t lose important design or API improvements while momentum is high.

The **most significant change required in this PR** is the **simplification of Execution Holons**, along with the largely mechanical ripple effects that flow from that decision (executor logic, recording rules, and test harness expectations). Once that lands, many of the surrounding concerns naturally fall into place.

Overall: the hard architectural work is done. What remains here is polish, alignment, and clearing the path for the next phase.

evomimic:
## ExecutionReference Simplification Rescinded

tests/sweetests/src/harness/test_case/dance_test_language.rs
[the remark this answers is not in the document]
evomimic: I opened Issue #387 to address this post PR. Am marking this conversation resolved for current PR.

tests/sweetests/tests/fixture_cases/setup_book_and_authors_fixture.rs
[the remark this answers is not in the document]
evomimic: Deferred to Issue #386

evomimic (approved):

evomimic:
# PR Review Complete:

I implemented, tested, and committed the following changes:
  - Refactored ExecutionReference and ExecutionHandle definitions to align with the revised execution model.
  - Replaced ExecutionReference::from_reference_parts with ExecutionReference::from_token_execution and updated all call sites.
  - Simplified ExecutionReference::assert_essential_content_eq to panic on failure instead of returning a Result, matching test-only usage semantics.
  - Updated all executors to use from_token_execution and the revised assert_essential_content_eq.
  - Moved tests/helpers into harness/helpers and resolved all module path and dependency ripple effects.
  - Centralized fixture context creation by moving init_fixture_context into TestCaseInit::new, and updated all fixtures to use the new API.
  - Removed unused imports and resolved compiler and IDE warnings.
  - Verified that all sweetests pass and addressed doc-test issues introduced by the refactor.
  - This PR brings the test harness into alignment with the revised test framework design and simplifies test authorship while preserving existing behavior.

Issue #387, opened 2026-02-06
Dance Test Refactor - Break-up dance_test_language into smaller modules


evomimic:
### 1. Summary

- Decomposes the monolithic `dance_test_language.rs` file into smaller, purpose-focused modules.  
- Renames `source_token` → `step_token` to reflect that a token represents the **entire step contract** (both source and expected), not just the input side.

---

### 2. Problem Statement

The `dance_test_language.rs` file has grown large and multi-focused, combining:
- core test case types,
- step definitions,
- builder/adders,
- and execution-adjacent semantics.

This makes the file harder to navigate, reason about, and safely evolve.  
Additionally, the name `source_token` is misleading, since a `TestReference` always carries **both** source and expected snapshots.

---

### 3. Dependencies

**Does this depend on other issues or features?**

- No hard dependencies.
- Should be done **after** the PR #381  lands to avoid rebasing churn.
- Requires no behavioral changes in execution or fixtures.

---

#### Decompose `dance_test_language.rs`

Split into the following structure:

```
test_case/
    mod.rs
    test_case.rs              // DancesTestCase, TestCaseInit, finalize logic
    test_steps.rs             // DanceTestStep enum + Display impl
    adders.rs                 // add_*_step methods (impl DancesTestCase)
```

Responsibilities remain unchanged; this is a mechanical refactor for clarity and maintainability.

#### Rename `source_token` → `step_token`

- Applies to:
  - `DanceTestStep` fields
  - adder method parameters
  - executor function parameters
- Improves semantic accuracy:
  - the token represents the step’s **full intent**, not just its input source.

---

### 5. Scope and Impact

**Impacted areas**
- Test harness internals (`dance_test_language` only)
- Executor function signatures
- Internal naming consistency

**Not impacted**
- Fixture authoring APIs
- Execution semantics
- Runtime behavior
- Public MAP APIs

This is a refactor + rename only.

---

### 6. Testing Considerations

- Existing test cases should pass unchanged.
- No new test cases required.
- Any failures indicate missed imports, module wiring errors, or incomplete renames.

---

### 7. Definition of Done

- `dance_test_language.rs` fully decomposed into the proposed module structure.
- All references to `source_token` replaced with `step_token`.
- All tests compile and pass without behavior changes.
- No public API changes outside the test harness.

<details>
<summary>Optional Details</summary>

### 8. Alternatives Considered

- Leaving the file monolithic and only renaming fields.  
  Rejected: does not address maintainability concerns.

[...]

Issue #386, opened 2026-02-04
TestCase Safeguard - Prevent mutation after finalization


dauphin3:
### 1. Summary (Required)

**What is the enhancement?**  
Introduce an explicit finalized state on DancesTestCase and enforce it:
Right now, a DancesTestCase can still accept new steps after finalize() has been called. This allows:
- helper-level finalize() calls to silently mask missing finalization in parent fixtures
- accidental mutation of a test case that is already considered complete
- ordering bugs where steps are added too late but still execute

---

### 2. Problem Statement (Required)

**Why is this needed?**  
- Makes finalize() a hard semantic boundary, not just a convention
- Ensures helpers cannot accidentally “complete” a test case
- Forces missing finalization to surface immediately instead of being masked
- Aligns test-case construction with the same immutability guarantees we expect from execution histories

This guard turns a fragile convention into an enforced invariant, which is exactly what this layer should be doing.

---

### 3. Dependencies (Required)

**Does this depend on other issues or features?**  
PR #381 

---

### 4. Proposed Solution (Required)

**How would you solve it?**  
- [ ] finalize():
sets is_finalized = true
errors (or panics in tests) if called more than once

- [ ] All add_*_step(...) methods:
must refuse to add steps if is_finalized == true
ideally return a structured error explaining that the test case is closed

---

### 5. Scope and Impact (Required)

**What does this impact?**  
Test harness

---

### 6. Testing Considerations (Required)

**How will this enhancement be tested?**

- Can it be validated with existing test cases?
- Do new test cases need to be created?
- Are there specific areas in the test ecosystem impacted by this enhancement?

---

### 7. Definition of Done (Required)

**When is this enhancement complete?**  
Provide clear criteria to know when the enhancement has been fully implemented and validated.

---

<details>
<summary>Optional Details (Expand if needed)</summary>

### 8. Alternatives Considered

**What other solutions did you think about?**  
Briefly describe alternatives and why this approach is preferred.

---

### 9. Risks or Concerns

**What could go wrong?**  
Highlight potential challenges, like performance or compatibility issues.

---

### 10. Additional Context

**Any supporting material?**  
Include links, screenshots, logs, or diagrams.

---

</details>

Issue #385, opened 2026-02-04
Dance Test Refactor - Consolidate TestCase description verbosity


dauphin3:
### 1. Summary (Required)

**What is the enhancement?**  
Remove lengthy description steps from TestCaseInit in the fixtures. Instead pass an optional description into the step adders which can be printed via info-level tracing in the rstest file dance_tests.rs for increased verbosity.

---

### 2. Problem Statement (Required)

**Why is this needed?**  
Avoids the need for manual hard coding of steps description and allows for clean verbosity enhancements.

---

### 3. Dependencies (Required)

**Does this depend on other issues or features?**  
PR #381 

---

### 4. Proposed Solution (Required)

**How would you solve it?**  
Each executor takes `description: Option<String>`  which will print a default info level message if None or the given String.
Before running execution, `fn rstest_dance_tests` iterates the steps to print `step.description` for each.

---

### 5. Scope and Impact (Required)

**What does this impact?**  
Fixtures and rstest file.

---

### 6. Testing Considerations (Required)

**How will this enhancement be tested?**
No testing required.

---

### 7. Definition of Done (Required)

**When is this enhancement complete?**  
Each fixture passes a description to every add_step.
Default for None cases prints correctly.
Info level messages are successfully printed during tests.

---

<details>
<summary>Optional Details (Expand if needed)</summary>

### 8. Alternatives Considered

**What other solutions did you think about?**  
Briefly describe alternatives and why this approach is preferred.

---

### 9. Risks or Concerns

**What could go wrong?**  
Highlight potential challenges, like performance or compatibility issues.

---

### 10. Additional Context

**Any supporting material?**  
Include links, screenshots, logs, or diagrams.

---

</details>

Issue #382, opened 2026-02-02
Dance Test Refactor - Add support for relationships in testing


dauphin3:
### 1. Summary (Required)

**What is the enhancement?**  
Ability to test relationships.
It was decided to remove relationships from EssentialHolonContent struct and enhance the assert_essential_content_eq function to recursively chase every related holon. 

---

### 2. Problem Statement (Required)

**Why is this needed?**  
To determine whether two holons are the same requires asserting that their properties are the same and that they are related to the same holons. Right now we are only testing that properties are the same, and in general we can't compare related holons by id because relationships could have targets in different lifecycle stages.

---

### 3. Dependencies (Required)

**Does this depend on other issues or features?**  
PR #381 

---

### 4. Proposed Solution (Required)

**How would you solve it?**  
execution_reference.rs   Enhance assert_essential_content_eq - get_all_related_holons for both expected and actual:  then iterate each relationships to verify the number of target holons is the same and then iterate each collection for that relationship name to exhaustively confirm that for each target holon in the expected collection exists a target in the actial collection whos essential content matches and remove it [need to copy actual collection]. (brute force nested loops approach)

---

### 5. Scope and Impact (Required)

**What does this impact?**  
holons_core/holon_utils.rs

Test harness and relationships tests cases

---

### 6. Testing Considerations (Required)

**How will this enhancement be tested?**

- Can it be validated with existing test cases?
- Do new test cases need to be created?
- Are there specific areas in the test ecosystem impacted by this enhancement?

---

### 7. Definition of Done (Required)

**When is this enhancement complete?**  
Provide clear criteria to know when the enhancement has been fully implemented and validated.

---

<details>
<summary>Optional Details (Expand if needed)</summary>

### 8. Alternatives Considered

**What other solutions did you think about?**  
Briefly describe alternatives and why this approach is preferred.

---

### 9. Risks or Concerns

**What could go wrong?**  
Highlight potential challenges, like performance or compatibility issues.

---

### 10. Additional Context

**Any supporting material?**  
Include links, screenshots, logs, or diagrams.

---

</details>

PR #392, opened 2026-02-10
Test refactor enhancments

dauphin3:
This PR covers:
- [x] #386 
- [x] #385 
- [x] #373 
- [x] #387 
- [x] #382

### DIFF
--- a/shared_crates/holons_core/src/core_shared_objects/transactions/transaction_context.rs
+++ b/shared_crates/holons_core/src/core_shared_objects/transactions/transaction_context.rs
@@ -41,7 +41,7 @@ impl fmt::Debug for TransactionContext {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         f.debug_struct("TransactionContext")
             .field("tx_id", &self.tx_id)
-            .field("is_open", &self.is_open.load(Ordering::Relaxed))
+            .field("is_open", &self.is_open())
             .finish()
     }
 }


tests/sweetests/tests/execution_steps/new_holon_executor.rs
evomimic: ## Not Merge-Ready — Recording Must Be Conditional on Successful Identity Creation

`new_holon` produces a new holon identity that cannot be deterministically known ahead of time. Therefore, recording execution is correct — but it must only occur when identity creation actually succeeds.

Currently, execution recording happens unconditionally:

    let execution_handle = ExecutionHandle::from(response_holon_reference);
    ...
    state.record(&step_token, execution_reference).unwrap();

This assumes that:

- A HolonReference is always returned
- Identity creation always succeeded

That is unsafe.

---

### ❗ Problem

If `expected_status != OK`, then:

- No new holon identity should be recorded
- The token should not become resolvable to a non-existent identity
- Downstream resolution would be corrupted

Even though `assert_eq!(response.status_code, expected_status)` is present, the executor still attempts to extract a `HolonReference` from the body and record it.

This violates the identity principle.

---

### ✅ Required Fix

Move identity extraction and recording inside a success guard:

    if response.status_code == ResponseStatusCode::OK {
        // extract HolonReference
        // apply properties
        // build execution handle
        // assert essential content
        // record execution
    }

If status is not OK:

- Do not attempt to extract a HolonReference
- Do not record execution
- Allow the token to remain unresolved

---

## After This Change

- Identity recording occurs only when a new identity is truly created.
- Negative test cases remain correct.
- Execution state cannot become inconsistent.
- Executor fully aligns with the core execution recording principle.

Once recording is gated on successful identity creation, this executor is merge-ready.

tests/sweetests/tests/fixture_cases/simple_add_remove_properties_fixture.rs
evomimic: ## Not Merge-Ready — Minor Structural & Clarity Fixes Required

This fixture is structurally sound, but there are a few issues that should be corrected before merge.

---

### 1️⃣ Variable Naming Mismatch After Remove Steps

The following variables are misnamed relative to the holon they are modifying:

```rust
let _modified_book_token_after_remove = test_case.add_remove_properties_step(
    &mut fixture_holons,
    modified_example_token,
    ...
```

and later:

```rust
let _modified_example_token_after_remove = test_case.add_remove_properties_step(
    &mut fixture_holons,
    modified_book_token,
    ...
```

The variable names imply the opposite holon of what is actually being modified.

This does not break execution, but it increases cognitive load and makes snapshot lineage harder to reason about.

**Recommended fix:**

Rename them to match the actual holon being mutated:

```rust
let _modified_example_token_after_remove = ...
let _modified_book_token_after_remove = ...
```

Keep token names aligned with the entity they represent.

---

### 2️⃣ Removal Step Uses Misleading PropertyMap Values

You correctly note:

> The value here is arbitrary and could be incorrect, which would go unflagged...

That is a real structural smell.

`remove_properties` semantics are name-based, not value-based, but the DSL requires a `PropertyMap`. This may mislead future readers into thinking the values matter.

**Recommended fix:**

Strengthen the comment to explicitly state:

```rust
// NOTE: Values are ignored by remove_properties; only property names are used.
// The DSL currently requires PropertyMap, but values here are semantically irrelevant.
```

No DSL redesign required in this PR — just clarity.

---

### 3️⃣ Property Name Case Inconsistency (Confirm Intent)

In setup:

```rust
"Title"
```

Later:

```rust
"title"
```

If property names are case-sensitive, this creates a new property rather than modifying the original.

If intentional (testing case sensitivity), document it.

If not intentional, make casing consistent.

---

## Summary

This fixture is otherwise architecturally sound and aligned with the execution model.

After:
- correcting variable names,
- clarifying remove semantics,
- and confirming property-name casing,

it will be merge-ready.

tests/sweetests/tests/fixture_cases/stage_new_from_clone_fixture.rs
evomimic: ## ⚠️ Fixture Cleanup Needed — Transient Pollution in Phase A

This fixture is very close to merge-ready, but there is one structural issue that should be addressed.

---

### ❗ Phase A Introduces Fixture State Pollution

In Phase A, a transient holon is created via:

- `add_new_holon_step(...)` → `ResponseStatusCode::OK`
- Followed by `add_stage_new_from_clone_step(...)` expecting `BadRequest`

The issue:

- The `add_new_holon_step` succeeds.
- A transient holon is now registered in `FixtureHolons`.
- It is never staged, committed, or abandoned.
- It permanently affects fixture state evolution.

This introduces an artificial lineage artifact that is unrelated to the canonical clone scenarios being tested.

While this does not immediately break correctness, it:
- Adds cognitive noise to fixture reasoning.
- Couples negative-path testing to fixture bookkeeping.
- Slightly weakens the invariant that fixture state mirrors meaningful execution evolution.

You even noted this in the TODO comment — that instinct is correct.

---

### ✅ Recommended Adjustment

Avoid minting a fixture-tracked transient solely to provoke a failure.

Cleaner alternatives:

1. Reuse a transient token from setup (before staging).
2. Construct a transient `TestReference` that does not mutate `fixture_holons`.
3. Refactor the DSL so negative-path tests do not require minting tracked identities.

The goal is:

> Negative-path validation should not introduce persistent fixture artifacts.

---

### Everything Else

Phases B and C are well-structured:

- Correct staged vs saved cloning.
- Proper commit round separation.
- Accurate database count assertions.
- Final saved-content verification.
- Correct identity recording expectations.

Those portions are merge-ready.

---

## Verdict

Not fully merge-ready due to transient state pollution in Phase A.

Refactoring Phase A to avoid unnecessary fixture mutation will make this fixture clean, deterministic, and architecturally aligned.

tests/sweetests/tests/fixture_cases/stage_new_version_fixture.rs
evomimic: ## ⚠️ Minor Fixture Consistency Issues — Nearly Merge-Ready

This fixture is structurally sound and aligns well with the execution-recording model. The staged→commit→versioning progression is coherent, and the use of `version_count` to validate staged lookup semantics is correct.

However, there are two small consistency issues worth addressing before merge.

---

## 1️⃣ Description Mismatch on Commit Step

This commit:

```
test_case.add_commit_step(
    &mut fixture_holons,
    ResponseStatusCode::OK,
    Some("With Properties -- first version cloned from book.".to_string()),
)?;
```

The description does not reflect what the step is actually doing.

That description belongs to the previous `with_properties` step, not the commit.

### Recommended Fix

Change to something structurally accurate, for example:

```
Some("Commit --- first version of book".to_string())
```

Descriptions should reflect lifecycle semantics, not duplicate prior step labels.

---

## 2️⃣ Version Count Mutation Timing

You are incrementing `version_count` **after** scheduling each `add_stage_new_version_step`.

That works today because the value is cloned and captured at call time, but it makes reasoning slightly harder.

### Cleaner Pattern

Increment before passing it:

```
version_count.0 += 1;
let _version_2_token = test_case.add_stage_new_version_step(
    ...
    version_count.clone(),
    ...
)?;
```

This makes the expected count explicit at scheduling time and avoids subtle off-by-one mental bookkeeping.

Not required for correctness — but improves clarity.

---

## What Is Correct

- Commit separation between base and versioned state is clear.
- Recording is conditional inside executor (assuming prior PR fix).
- Multiple staged versions before commit are intentionally tested.
- Duplicate base-key semantics are exercised via `expected_failure_code`.
- Finalization occurs cleanly.

---

## Verdict

Functionally correct and structurally aligned.

Recommended: fix commit description and adjust version_count increment order for clarity.

After that, this fixture is merge-ready.

evomimic (changes requested):

dauphin3:
@evomimic I need you to take care of the last 2 unresolved comments.  
-The optionality description comment is mostly done, I need help finishing the impl Debug for DancesTestCase 
- Phase A was something you had recommended and I'm not totally sure what's going on here.. maybe it's so bad to leave it as is, but if you can fix it that would be great

btw this time for my cmake error, `cargo update` did not work. Note I haven't been able to run tests yet.
branch 382 was working fine, so something about my rebase or clean set it off.  I'll troubleshoot if ball gets passed back to me (I think this PR is complete).

tests/sweetests/tests/fixture_cases/stage_new_from_clone_fixture.rs
evomimic: Merging this PR with this issue still unresolved... will solve under a separate issue post-merge.

evomimic (approved):
## PR Review Summary

Approved for merge. This is merge-ready for unblocking downstream work, with one explicitly accepted post-merge cleanup item.

### What’s now in good shape

- Logging model is coherent and aligned with #385 intent:
  - Step narration is owned by the `rstest_dance_tests` driver.
  - Executors no longer carry redundant `description` parameters.
  - `DanceTestStep` `Display` now includes concise, actionable identifiers (token/state/count/status) while keeping narrative text readable.

- Identifier noise is controlled:
  - `TemporaryId` display is centrally shortened (8-char form), improving readability across downstream displays.
  - `TestReference` display benefits from this lower-level formatting and remains structurally informative.

- API/behavior correctness remains intact:
  - Client/guest dance-path and harness fixes compile and execute cleanly.
  - Current runs confirm clean compile and green sweetests, including `--nocapture`.

- Warning cleanup was applied without semantic risk:
  - Removed stale imports/re-exports.
  - Dropped redundant `FixtureHolon.id`.
  - Marked intentional unused parameter as `_state`.

### Noted follow-up (explicitly accepted post-merge)

There is one known fixture-hygiene debt in `stage_new_from_clone_fixture`:
- Phase A mints a tracked transient solely to drive a negative-path (`BadRequest`) assertion.
- This introduces fixture-state pollution (an extra transient lineage artifact), even though it does not currently break correctness.

This is an architectural cleanliness issue, not a correctness blocker for this PR.  
Given the need to unblock further development, this is accepted for post-merge follow-up.

### Outcome

Substantial progress delivered and stable: improved diagnostics, cleaner harness contracts, reduced noise, and no observed regressions.  
Merge now, then follow up with targeted cleanup to model negative-path probes without mutating fixture lineage.

PR #395, opened 2026-02-12
test_case folder restructure

dauphin3:
-dance_test_lanaguage.rs file restructure
-rename: source_token --> step_token
-cleanup imports & warnings

### DIFF
--- a/tests/sweetests/src/harness/test_case/mod.rs
+++ b/tests/sweetests/src/harness/test_case/mod.rs
@@ -9,7 +9,7 @@
 //! by the sweetests harness.
 //!
 //! Test cases constructed using this language are *pure specifications*:
-//! they contain no runtime context, no concrete holon identifiers, and no
+//! they contain no execution-time context, no concrete runtime identifiers, and no
 //! execution logic. Resolution of references, state mutation, and dance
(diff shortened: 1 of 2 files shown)


tests/sweetests/src/harness/test_case/adders.rs
evomimic: ### Suggestion: Extract `ensure_not_finalized()` Helper

Several adder methods repeat the same guard:

```rust
if self.is_finalized == true {
    return Err(HolonError::Misc(
        "DancesTestCase is already finalized, thus closed for any additional steps"
            .to_string(),
    ));
}
```

Since this pattern appears in many methods, it may be worth extracting a small helper to centralize the invariant:

```rust
fn ensure_not_finalized(&self) -> Result<(), HolonError> {
    if self.is_finalized {
        return Err(HolonError::Misc(
            "DancesTestCase is already finalized, thus closed for any additional steps"
                .to_string(),
        ));
    }
    Ok(())
}
```

Then each adder becomes cleaner and less error-prone:

```rust
self.ensure_not_finalized()?;
```

Benefits:

- Removes repetition across methods  
- Prevents future adders from accidentally omitting the guard  
- Centralizes the lifecycle invariant in one place  
- Slightly improves readability of the builder methods

Issue #405, opened 2026-02-23
MAP Commands -- Phase 2.2 — Implement IntegrationHub MAP Command dispatcher


evomimic:
### 1. Summary (Required)

**What is the enhancement?**  
Complete the MAP domain execution boundary established in Phase 2.1 by centralizing policy enforcement in `Runtime`, implementing the in-scope MAP command dispatchers, and integrating `ClientSession` for undo/recovery support as the supporting core PR 418 lands.

This dispatcher is the single domain choke point that:

- binds Wire -> Domain
- resolves transaction context
- enforces lifecycle/descriptor policy
- routes to scope-specific domain dispatchers
- converts domain results/errors back to wire form

This phase builds the new domain path in parallel with legacy paths (no cutover removal yet).

---

### 2. Problem Statement (Required)

**Why is this needed?**  
Even with a structural command contract, correctness remains fragile unless enforcement is centralized. Today, binding/policy/routing concerns are distributed across ingress/receptor-era paths and some lifecycle and commit-like ingress policy still lives in receptor-era code paths, even though the intended architecture is for Runtime / command dispatch to own that policy.

We need one auditable place where:

- wire binding happens
- tx selection and reference binding are enforced consistently
- lifecycle and commit guard policy are applied deterministically
- domain execution routing is structural (not string authority)
- error/result shaping is standardized

Without this, policy leaks and architectural drift continue.

---

### 3. Dependencies (Required)

**Does this depend on other issues or features?**

Reference docs: [MAP Commands Spec](https://memetic-activation-platform.github.io/map-dev-docs/core/commands) and [MAP Commands Cheat Sheet](https://memetic-activation-platform.github.io/map-dev-docs/core/commands-cheat-sheet/) define baseline vocabulary and command inventory for v0 alignment.

- Phase 2.1 contract types and boundaries are defined via Issue #404 .
- Core lifecycle primitives and tx-bound reference binders are available
- Legacy paths remain active during this phase for parity/testing

---

### 4. Proposed Solution (Required)

**How would you solve it?**

#### 4.1 Implement Scope-Specific Domain Dispatchers

Implement full command bodies inside each scope dispatcher. Dispatchers already exist and route correctly; the work is filling in the `NotImplemented` stubs.

These methods operate on domain types only (no wire types).

Scope branching is structural routing only; policy decisions remain descriptor-driven.

`Query` will remain `NotImplemented` as its core capability is not yet implemented.
---

[...]

PR #430, opened 2026-03-27
426 enable multi transaction and command dispatch testing in sweetest harness via runtime

### LINKED ISSUE #426: Enable Multi-Transaction and Command Dispatch Testing in Sweetest Harness via Runtime
### 1. Summary

**What is the enhancement?**

Refactor the sweetest test harness to use `Runtime` as its transaction execution boundary, so transaction-scoped test steps run through the same `Runtime::execute_command()` layer that production uses after IPC binding.

This migration has three goals:

- Exercise transaction lifecycle operations through `RuntimeSession` and `Runtime::execute_command()`
- Make client-side transaction lifecycle authoritative in tests (`Open` → `Committed`, then reject further transaction operations)
- Enable runtime-backed multi-transaction scenarios in the harness, starting with “commit, then begin a new transaction”

This change is a harness migration, not a second parallel harness. There should remain one set of `DanceTestStep` variants and one execution model for transaction-scoped test steps.

This issue is intentionally scoped to the runtime execution boundary, not full IPC transport coverage. `MapCommandWire` / `MapIpcRequest` binding remains a separate test concern.

---

### 2. Problem Statement

**Why is this needed?**

The sweetest harness is still centered on a single long-lived `TransactionContext`, and most execution steps call `context.initiate_dance()` directly.

That creates three problems:

**1. Transaction lifecycle is not being tested at the runtime boundary.**  
The harness does not currently drive transaction operations through `Runtime::execute_command()`, so the command descriptor layer, lifecycle gating, commit guard behavior, and session-driven transaction ownership are not covered by integration tests.

**2. The harness assumes a single transaction remains usable for the whole test.**  
That assumption no longer matches the intended lifecycle policy. Once a transaction is committed, subsequent transaction-scoped lookups and mutations should fail with `TransactionAlreadyCommitted`. Existing tests like `stage_new_version_test` still assume they can commit and continue using the same transaction.

**3. Multi-transaction scenarios are not supported by the harness.**  
The runtime session can own multiple transactions, but the test harness only tracks one context and has no notion of beginning a fresh transaction mid-test.

Routing transaction-scoped steps through `Runtime` addresses these problems by making the harness use the same lifecycle enforcement and transaction ownership model as the real runtime layer.

---

### 3. Dependencies

- **`RuntimeSession` (already finished)** — transaction ownership layer in `host/crates/map_commands_runtime/src/runtime_session.rs`
- **`Runtime::execute_command` pipeline (already finished)** — lifecycle enforcement and scope-based dispatch in `host/crates/map_commands_runtime/src/runtime.rs`
- **`Runtime::session()` accessor (already exists)** — `runtime.rs:31`
- **Transaction lifecycle state (already merged)** — `TransactionContext` lifecycle enforcement (`Open` → `Committed`), `assert_allowed()` policy gate, and `host_commit_in_progress` guard
- **`map_commands` crate split (already finished)** — `map_commands_contract` (domain types), `map_commands_wire` (IPC envelope types), `map_commands_runtime` (execution)
- **No blockers** — all prerequisite infrastructure is in place in #405 

---

### 4. Proposed Solution

**How would you solve it?**

#### Design principle: one execution path, matching production

All test step executors are migrated to build `MapCommand`s and dispatch through `Runtime::execute_command()`. This is the same execution boundary that Conductora uses in production (see `host/conductora/src/runtime/dispatch_map_command.rs`). Wire/IPC envelope binding (`MapCommandWire` → `MapCommand`) lives in Conductora, not in `map_commands_runtime` — the test harness operates at the domain command level, below the IPC envelope.

There is no `Option<Runtime>` or dual-path design. Every test uses `Runtime`.

#### 4.1 Add `map_commands_runtime` and `map_commands_contract` as dependencies of sweetests

`Runtime` and `RuntimeSession` live in `host/crates/map_commands_runtime/`. The command and result types (`MapCommand`, `MapResult`, `TransactionAction`, `SpaceCommand`, etc.) live in `host/crates/map_commands_contract/`. The sweetest crate (`holons_test`) already depends on `holons_client` from the host workspace (`tests/sweetests/Cargo.toml:46`), so this doesn't introduce a new cross-boundary pattern.

#### 4.2 Replace `TestExecutionState` with Runtime-based state

```rust
pub struct TestExecutionState {
    pub runtime: Runtime,
    pub active_tx_id: TxId,
    pub execution_holons: ExecutionHolons,
}
```

`Runtime` is required, not optional. `active_tx_id` tracks which transaction is currently targeted by test steps. The `RuntimeSession` inside `Runtime` is the single source of truth for all transaction state — there is no cached `Arc<TransactionContext>`.

Accessor and helper methods:

```rust
impl TestExecutionState {
    pub fn new(runtime: Runtime, active_tx_id: TxId) -> Self {
        TestExecutionState {
            runtime,
            active_tx_id,
            execution_holons: ExecutionHolons::default(),
        }
    }

    /// Look up the active transaction context on demand from the session.
    pub fn context(&self) -> Result<Arc<TransactionContext>, HolonError> {
        self.runtime.session().get_transaction(&self.active_tx_id)
    }

    /// Begin a new transaction via SpaceCommand::BeginTransaction and make it active.
    pub async fn begin_transaction(&mut self) -> Result<TxId, HolonError> {
        let command = MapCommand::Space(SpaceCommand::BeginTransaction);
        let result = self.runtime.execute_command(command).await?;
        match result {
            MapResult::TransactionCreated { tx_id } => {
                self.active_tx_id = tx_id;
                Ok(tx_id)
            }
            other => Err(HolonError::InvalidParameter(
                format!("Expected TransactionCreated, got {:?}", other)
            )),
        }
    }
}
```

#### 4.3 Replace `init_test_context()` with `init_test_runtime()`

```rust
pub async fn init_test_runtime(
    test_case: &mut DancesTestCase,
) -> (Runtime, TxId) {
    init_tracing();

    let holon_service: Arc<dyn HolonServiceApi> = Arc::new(ClientHolonService);
    let dance_initiator = create_test_dance_initiator().await;

    let space_manager = Arc::new(HolonSpaceManager::new_with_managers(
        Some(dance_initiator),
        holon_service,
        None,
        ServiceRoutingPolicy::Combined,
    ));

    let session = Arc::new(RuntimeSession::new(Arc::clone(&space_manager)));
    let runtime = Runtime::new(session);

    // Begin first transaction through the real command path
    let result = runtime
        .execute_command(MapCommand::Space(SpaceCommand::BeginTransaction))
        .await
        .expect("failed to begin initial transaction");

    let tx_id = match result {
        MapResult::TransactionCreated { tx_id } => tx_id,
        _ => panic!("expected TransactionCreated"),
    };

    // Import transient holons from fixture phase
    let context = runtime.session().get_transaction(&tx_id)
        .expect("failed to get initial transaction");
    let bound = test_case.test_session_state
        .get_transient_holons()
        .clone()
        .bind(&context)
        .expect("failed to bind transient holons");
    context.import_transient_holons(bound)
        .expect("failed to import transient holons");

    (runtime, tx_id)
}
```

`init_test_context()` is removed. `init_fixture_context()` stays unchanged (fixture phase doesn't execute commands).

#### 4.4 Update `dance_tests.rs` initialization

```rust
let (runtime, tx_id) = init_test_runtime(&mut test_case).await;
let mut state = TestExecutionState::new(runtime, tx_id);
```

#### 4.5 Migrate executors to the runtime boundary

Transaction-scoped executors should be rewritten to dispatch through `Runtime::execute_command()`.

Preferred execution order:

- Use native `TransactionAction` / `HolonAction` commands where available
- Use `TransactionAction::Dance(...)` only as a temporary fallback where native command coverage is incomplete

Examples of steps that should become native runtime-backed commands immediately:

- `Commit`
- `BeginTransaction`
- staged/transient lookup helpers
- `NewHolon`
- `StageNewHolon`
- `StageNewVersion`
- `StageNewFromClone`
- `DeleteHolon`

Assertion-only helpers do not need to be forced through `MapCommand` if direct bound-reference reads are simpler and the command boundary itself is not what is under test.

#### 4.6 Add `BeginTransaction` first, leave room for `SwitchTransaction`

As the first multi-transaction capability, add a `BeginTransaction` test step that opens a new transaction and makes it active.

This is enough to repair tests that need to:

1. commit a transaction
2. begin a fresh transaction
3. continue with more transaction-scoped work

Support for switching among multiple still-open transactions can be added in a follow-up issue if needed. If implementation naturally supports a `SwitchTransaction` step cleanly, that can be included; otherwise it should remain explicitly out of scope for this issue.

#### 4.7 Fixture phase — no changes

The fixture phase creates transient holons for snapshot expectations. It doesn't execute commands or commit. It continues using a bare `TransactionContext` via `init_fixture_context()`.

---

### 5. Scope and Impact

**What does this impact?**

| Area | Impact |
|------|--------|
| `tests/sweetests/Cargo.toml` | Add `map_commands_runtime` and `map_commands_contract` |
| `execution_state.rs` | Replace single-context state with `runtime + active_tx_id` |
| `test_context.rs` | Replace `init_test_context()` with `init_test_runtime()` |
| `dance_tests.rs` | Update initialization to create runtime-backed execution state |
| executor files | Migrate transaction-scoped execution through `Runtime::execute_command()` |
| `test_steps.rs` | Add `BeginTransaction` |
| adders | Add `add_begin_transaction_step()` |

**What stays unchanged:**

- Fixture authoring model
- Existing `DanceTestStep` vocabulary, except for additive lifecycle steps like `BeginTransaction`
- Fixture-phase initialization
- `TestReference`, `FixtureHolons`, `FixtureBindings`
- Transport-layer IPC binding types (`MapIpcRequest`, `MapCommandWire`)

---

### 6. Testing Considerations

- All existing sweetest cases should continue passing unless they were implicitly relying on reuse of a committed transaction.
- `stage_new_version_test` should be updated to stop assuming post-commit reuse of the same transaction.
- At minimum, add a runtime-backed lifecycle test that proves:
  - a transaction can be committed
  - a subsequent transaction-scoped operation on that same transaction is rejected
  - a fresh `BeginTransaction` allows work to continue
- If scope permits, add a follow-up or additional test for multiple live transactions and switching between them.
- Native-command migration should be incremental where needed. If some existing steps still rely on `TransactionAction::Dance(...)`, that is acceptable within this issue as long as the harness itself is runtime-backed and lifecycle-correct.

---

### 7. Definition of Done

- [x] `pub fn session()` accessor exists on `Runtime` (`runtime.rs:31`)
- [x] `map_commands_runtime` and `map_commands_contract` added to `holons_test`
- [x] `TestExecutionState` stores `Runtime` and `active_tx_id` instead of a single public `TransactionContext`
- [x] `init_test_runtime()` implemented and wired into the test harness
- [x] `BeginTransaction` step added with adder and executor
- [x] `Commit` step dispatched through `Runtime::execute_command()`
- [x] Transaction-scoped executors migrated to the runtime boundary
- [x] Native command actions are used where already supported by the contract/runtime
- [x] Any remaining dance-backed executor paths are explicitly documented as temporary fallback
- [x] `stage_new_version_test` updated to reflect committed-transaction lifecycle rules
- [x] Existing sweetest tests pass, except for cases intentionally deferred due to missing native command coverage
- [x] At least one lifecycle test demonstrates: commit, rejection on committed tx, begin new tx, continue work

owleyeview:
### Summary

This PR includes work from two issues.

**#405** introduced the runtime execution infrastructure this harness migration depends on:
- Split `map_commands` into `map_commands_contract` (domain types), `map_commands_wire` (IPC envelope), and `map_commands_runtime` (execution), keeping transport types out of execution signatures
- Enforced committed-transaction lifecycle: `TransactionAlreadyCommitted` is returned when commands target a committed transaction, with `CommandDescriptor` gating read-only vs. mutation access

**#426** migrated the sweetest harness to use that infrastructure:
- Migrates test step executors from direct `context.initiate_dance()` calls to `Runtime::execute_command(MapCommand)` dispatch, matching the production execution boundary
- Adds `BeginTransaction` test step for multi-transaction scenarios
- Adds `transaction_lifecycle_test` proving: commit → rejection on committed tx → begin new tx → continue work
- Introduces `HolonErrorKind` enum for asserting expected error variants in tests without matching exact string contents
- Adds `rebind()` on wire types (`SerializableHolonPool`, `HolonReferenceWire`, etc.) for re-importing serialized fixture data into newly opened transactions, intentionally bypassing the tx_id validation that `bind()` performs (see doc comments)
- Adds minimal `ResponseStatusCode` → `HolonError` conversion in `ClientHolonService` to preserve error variants like `HolonNotFound` rather than flattening everything to `Misc`

### Open threads

1. **Guest-side error flattening:** Most guest-side `HolonError` variants are collapsed into a handful of `ResponseStatusCode` values in the dance response and re-wrapped as `HolonError::Misc(String)` on the client side. The minimal conversion added here recovers a few variants (`NotFound`, `NotAuthorized`), but a proper solution would make the original `HolonError` available to the client directly — either via a serialized error payload in the dance response or a dedicated error wire type.

2. **`QueryRelationships` test step:** This step still falls back to `context.initiate_dance()` because there is no native `TransactionAction::Query` command yet. Worth considering whether to remove it as a step entirely and convert it to a `RelatedHolons` assertion helper (similar to `MatchSavedContent`). Relatedly: should `AbandonStagedChanges` be promoted to a native command in the contract?

3. **Step contract stores live references:** `DanceTestStep` variants (e.g., `LoadHolons`) currently store raw `TransientReference`/`HolonReference` values carrying a fixture-time `TransactionContextHandle`, which forces executors to rebind them at execution time. A cleaner design would store identity data (`TemporaryId`, `HolonId`) instead, making rebinding unnecessary. Harness-internal refactor, no production impact.

### Test plan

- [x] All existing sweetest cases pass (`npm run sweetest`)
- [x] `transaction_lifecycle_test` covers commit → reject → begin new tx → continue
- [x] Verify no regressions in CI after merge
- [ ] Future: convert remaining dance-fallback executor paths to native commands as contract coverage expands


evomimic (approved):
Approve.

This looks merge-ready to me for `405-map-commands----phase-22-implement-integrationhub-map-command-dispatcher`, assuming the planned stack order stays the same (`429` and `418` first, then this branch into `405`).

What I like most here is the architectural direction: the sweetest harness is being moved onto `Runtime::execute_command(...)`, which already returns `Result<MapResult, HolonError>`. That helps decouple test execution from the older `DanceResponse` / `ResponseStatusCode` shape and moves the harness toward the IPC-native future where `HolonError` can be returned directly without HTTP-style status-code translation.

I also like the addition of `HolonErrorKind`. That gives the tests a stable way to assert expected error variants without depending on exact string payloads or transport-shaped status codes. In other words, this PR helps the transition to direct `HolonError` return by teaching the harness to think in terms of domain errors rather than `ResponseStatusCode`s.

Two non-blocking caveats I want to note:

1. The client-side error-kind preservation is narrower than the PR summary suggests. The new `map_dance_error(...)` helper is only wired into `delete_holon_internal`; the other dance-backed client methods still flatten non-OK responses to `HolonError::Misc`. That’s consistent with the open thread, but it means this PR does not broadly solve guest-side error flattening yet.

2. The remaining dance-fallback executors (`QueryRelationships` and `AbandonStagedChanges`) do not fully enforce the new `HolonErrorKind` contract on non-OK `DanceResponse`s. They currently assert “not OK” rather than a specific expected variant, so those paths are still a bit weaker than the fully runtime-backed steps.

None of that feels blocking for this branch. My read is that this PR is a good transitional step: it materially advances the runtime-backed command path and aligns the tests with direct `HolonError` handling, while intentionally not attempting to fully deliver the end-state of returning original `HolonError`s across every remaining dance-backed boundary.

PR #448, opened 2026-04-13
441 post pr440 hardening complete recovery wiring and receptor runtime finalization

nphias:
```mermaid
flowchart LR
    A[Runtime] --> B[RuntimeSession]

    B --> C[active_sessions: HashMap<TxId, Arc<ClientSession>>]
    B --> D[archived_sessions: HashMap<TxId, Arc<ClientSession>>]
    B --> E[HolonSpaceManager]
    B -- Restore Tx session --> I[LocalRecoveryReceptor]

    C --> G[ClientSession]
    D --> G

    G --> H[TransactionContext]
    G --> F[LocalRecoveryReceptor]
    
    I --> J[list_open_sessions]

    F --> K[recover_latest tx_id]
    F --> L[persist context, description, disable_undo]
    F --> M[cleanup tx_id]

```

- wiring complete
  - runtime-init does orchestration
  - runtime delegates to runtime-session
  - runtime session , takes care of recovery and manages ClientSessions
  
- tests added
-  execution policy:
runtime.rs
- added ExecutionPolicy { snapshot_after }
- kept execute_command(...) as the default wrapper
- added execute_command_with_policy(...)
- after successful dispatch, it now calls session.persist_success(...) when:
  - snapshot_after is true
  - the command is not ReadOnly
  - the command has a transaction context
  - added a small command_label(...) helper for snapshot descriptions

lib.rs
 - now exports ExecutionPolicy
 - dispatch_map_command.rs
 - now passes RequestOptions.snapshot_after into Runtime::execute_command_with_policy(...)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4092,9 +4092,11 @@ version = "0.1.0"
 dependencies = [
  "base_types",
  "core_types",
+ "holons_client",
  "holons_core",
  "holons_loader_client",
  "map_commands_contract",
+ "recovery_receptor",
  "serde_json",
  "tokio",
 ]
--- a/host/Cargo.lock
+++ b/host/Cargo.lock
@@ -5220,9 +5220,11 @@ version = "0.1.0"
 dependencies = [
  "base_types",
  "core_types",
+ "holons_client",
  "holons_core",
  "holons_loader_client",
  "map_commands_contract",
+ "recovery_receptor",
  "serde_json",
  "tokio",
 ]
--- a/host/conductora/src/map_commands/root_space.rs
+++ b/host/conductora/src/map_commands/root_space.rs
@@ -7,11 +7,12 @@ use tauri::{command, State};
 pub async fn root_space (
     _receptor_factory: State<'_, ReceptorFactory>,
 ) -> Result<SpaceInfo, String> {
-    unimplemented!("This command is currently a placeholder and needs to be implemented to fetch the root space information from the appropriate receptor.");
-
+    tracing::warn!("[TAURI COMMAND] 'root_space' invoked but not implemented; returning typed error");
+    Err("root_space command is not implemented. Use dispatch_map_command instead or see issue #441 for follow-up.".to_string())
+}
     //tracing::debug!("[TAURI COMMAND] 'root_space' command invoked");
     //let spaces = receptor_factory.all_spaces_by_type(&ReceptorType::Local)
      //   .await
       //  .map_err(|e| format!("receptor service error: {:?}", e))?;
     //Ok(spaces)
-}
+
--- a/host/conductora/src/runtime/dispatch_map_command.rs
+++ b/host/conductora/src/runtime/dispatch_map_command.rs
@@ -2,7 +2,7 @@ use std::sync::RwLock;
 
 use core_types::HolonError;
 use map_commands_contract::MapCommand;
-use map_commands_runtime::{ExecutionPolicy, Runtime};
+use map_commands_runtime::{Runtime}; //ExecutionPolicy
 use map_commands_wire::{
     MapCommandWire, MapIpcRequest, MapIpcResponse, MapResultWire,
 };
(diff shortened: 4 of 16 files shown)


evomimic (changes requested):
Reviewing against Issue #441 only. The larger Experience Unit / marker / undo-redo model alignment work has been intentionally deferred to Issue #450, which now tracks alignment with the authoritative v1.2 spec:

- Issue #450: Align MAP Undo/Redo Runtime and SDK with Experience Unit Model
- Spec: [MAP Core — Structural Pre-Commit Transaction Editing Model -- v1.2](https://memetic-activation-platform.github.io/map-dev-docs/core/commands-and-runtime/struct-exp-undo-redo/)

## Findings

1. [P1] `snapshot_after` is being applied to `Commit`, which is semantically invalid and can recreate recovery state after cleanup  
`host/crates/map_commands_runtime/src/runtime.rs`

`snapshot_after` is now clarified as an experience-layer boundary signal for eligible mutable editing operations that leave the transaction open for continued editing. `Commit` is not such an operation: it closes the transaction boundary, destroys transaction-scoped undo/recovery state, and must not create a new checkpoint afterward.

In the current flow, commit cleanup runs first, but the generic post-success `snapshot_after` hook can still persist a fresh recovery snapshot for that same transaction immediately afterward. That is both a logic bug and a semantic mismatch:

- logic bug: cleanup is undone by a later persist
- semantic bug: `Commit` should never honor `snapshot_after`

This should be fixed in this PR by ensuring `snapshot_after` is applied only to eligible in-transaction editing commands, never to `Commit`.

2. [P1] `root_space` is still an invoke-exposed panic path  
`host/conductora/src/map_commands/root_space.rs`

Issue #441 explicitly called for removing the placeholder panic path, but `root_space` still unconditionally uses `unimplemented!()`. Since this command remains registered, the host can still panic on invocation.

3. [P2] Local provider behavior without `"recovery"` remains implicit  
`host/conductora/src/setup/providers/local/setup.rs`

The branch still silently does nothing when a local provider is enabled without the `"recovery"` feature. Issue #441 required explicit behavior here: either register a non-recovery local receptor path or fail fast with a typed configuration error.

4. [P2] Runtime ingress must remain unified  
`host/crates/map_commands_runtime/src/runtime.rs`

This PR introduces request metadata for `snapshot_after`, but it should not split the runtime into separate execution entrypoints. The runtime should expose one authoritative execution method and apply request metadata there. Since this PR is the one introducing the split, unifying `execute_command` and `execute_command_with_policy` belongs in this PR, not a later cleanup.

## Note on Scope

By factoring the Experience Unit / marker-aware undo-redo model work out to Issue #450 , we can keep the scope of this PR tightly focused on just fixing the problems noted here.

nphias:
four fixes made as requested.
execution_policy argument logic deferred to issue 450 - Experience unit model

evomimic (approved):
Approving for merge with explicit defer of the remaining `snapshot_after` / Experience Unit boundary wiring work to Issue #450.

I reviewed the latest branch state against the current intended scope and the previously identified issues. The fixes for the concrete hardening items look sound:

- `Commit` no longer recreates recovery state after cleanup
- `root_space` no longer panics on invocation and now returns a typed error
- local provider setup without `"recovery"` now fails fast explicitly
- runtime ingress has been re-unified

I also confirmed that current validation remains green:

- happ build passes
- host build passes
- all sweetests pass
- app behavior is good under both `HC_DEV_MODE=1` and `HC_DEV_MODE=0`

One important note on scope: the `snapshot_after` path is not being treated as complete in this PR. We got somewhat ahead of ourselves by beginning recovery persistence / undo-redo integration before the broader Experience Unit model and surrounding capabilities were fully in place. Given that this path is not currently blocking downstream work, and given that Issue #450 now exists specifically to align runtime/SDK behavior with the authoritative v1.2 undo/redo spec, I’m comfortable deferring that wiring there rather than forcing an interim implementation here that will be rewritten.

Follow-up is explicitly tracked in:

- Issue #450: Align MAP Undo/Redo Runtime and SDK with Experience Unit Model
- Spec: [MAP Core — Structural Pre-Commit Transaction Editing Model -- v1.2](https://memetic-activation-platform.github.io/map-dev-docs/core/commands-and-runtime/struct-exp-undo-redo/)

So this approval is for PR 448 as a hardening/integration step, with the understanding that Experience Unit boundary semantics and the final `snapshot_after` wiring are intentionally deferred to #450.

PR #487, opened 2026-05-05
479 make loader inverse relationship resolution schema aware

### LINKED ISSUE #479: Make loader inverse relationship resolution schema-aware
### 1. Summary (Required)

**What is the enhancement?**
Make inverse-relationship handling correct and robust across the loader pipeline. Two related bugs surfaced while implementing #463:

1. The loader client had deferred dealing with inverse relationships properly.  It asserts `is_declared = true` for every relationship in `LoaderRelationshipReference`, but the client cannot reliably resolve authored relationship names to declared-vs-inverse RTDs from the current import payload alone, because the needed relationship descriptors may be staged in the same load set or already saved. Inverse relationships present in schema JSON (e.g. `VariantOf`) are persisted as-is instead of being rewritten to their declared counterpart (`Variants`).
2. The guest resolver's inverse Relationship TypeDescriptor(RTD) lookup uses exact concrete endpoint descriptor keys, so it fails to find relationship type descriptors that are declared on abstract supertypes when the actual endpoints are concrete subtypes.

This issue moves declaredness resolution to the guest (where the type-system authority lives) and generalizes the inverse RTD lookup to walk `Extends` on endpoint descriptors.

---

**Why is this needed?**
The loader currently fails to reliably convert inverse relationships from import JSON into their declared form before saving. This breaks descriptor semantics that expect declared-side relationships to exist after schema load.

Concrete example: schema JSON uses `VariantOf` from enum variant types back to their enum value type, but runtime descriptor logic expects the declared `Variants` relationship from enum value type to enum variant type. Today, because the client unconditionally tags every relationship `is_declared = true` (`builder.rs:209`), the inverse-side relationship is persisted directly and `EnumValueDescriptor` cannot discover its declared variants.

The client should not fix this on its own. The JSON states the authored relationship name and direction as written on the source holon, but resolving that authored name to a declared or inverse relationship type requires the active type graph. Only the guest, which already runs a fixed-point resolution pass over the staged load set and the saved type graph, has the information needed.

A second, independent bug compounds the first: even when the guest knows it is looking at an inverse relationship, `declared_name_for_inverse()` (`loader_ref_resolver.rs:464`) constructs the canonical RTD key from the concrete endpoint descriptors and looks it up exactly. Relationships declared on abstract supertypes (such as `(EnumVariantValueType)-[VariantOf]->(EnumValueType)`) are not found when concrete endpoints extend those abstract descriptors.

---

Discovered while implementing #463; tracked separately as a loader-correctness enhancement. Depends on the current loader schema, core schema relationship descriptors, and the existing guest loader inverse-resolution path. No external blockers.

---

**How would you solve it?**

Make the guest the single authority for relationship declaredness, and generalize inverse RTD lookup to walk `Extends`.

**Client-side (`host/crates/holons_loader_client`)**
- Remove `IsDeclared` from `LoaderRelationshipReference` — both the wire/transient representation and the parser/builder code that emits it. The field is currently a stub (always `true`); deleting it prevents future code from mistakenly trusting it.
- Parser stops attempting to classify relationships as declared vs inverse. It emits each relationship exactly as written in JSON, with its source/target/name.

**Guest-side (`happ/crates/holons_loader`)**
- During relationship resolution, use the authored JSON relationship name plus the resolved source/target endpoint descriptor lineages to find the applicable RTD in the staged and saved type graph.
- If the RTD is declared, persist the relationship as authored. If the RTD is inverse, follow InverseOf and persist the canonical declared-side relationship with endpoints swapped. This is what makes `VariantOf` load as `Variants`.
- Generalize `declared_name_for_inverse()` to walk `Extends` on both source and target endpoint descriptors when looking up the canonical RTD key, so RTDs declared on abstract supertypes are found.
- Preserve existing pass ordering: `DescribedBy` (pass 2a) and `InverseOf` (pass 2b) continue to be resolved before the general fixed-point pass, so the type graph needed for declaredness resolution is in place when general relationships are processed.

**Guardrail**
- The resolver must not invent or substitute a relationship name before resolving the authored JSON name. The authored name is the lookup key; the graph determines whether that authored name is declared or inverse.

**Out of scope (deferred)**
A future optimization could reintroduce an optional client-side hint (`Option<bool>`) on `LoaderRelationshipReference` that the client populates only when it can prove declaredness from local information, and which the guest treats as a non-authoritative hint it verifies. This is **not** part of this issue. Guest-side resolution is correct and sufficient; a hint layer should only be added if profiling shows guest resolution is a hot-path bottleneck.

---

**What does this impact?**

- `host/crates/holons_loader_client` — parser, builder, and types: remove `IsDeclared`.
- `happ/crates/holons_loader` — `loader_ref_resolver.rs` (declaredness resolution via `Extends`, inverse-to-declared rewrite, generalized inverse RTD lookup) and any callers in `loader_holon_mapper.rs` / `controller.rs` that consume the removed field.
- Loader schema JSON for `LoaderRelationshipReference` — remove the `IsDeclared` property.
- Core schema loading (used in sweetests and bootstrap): no JSON changes; `VariantOf` and similar inverse relationships start being correctly rewritten to declared counterparts at load.
- Descriptor semantics that traverse declared relationships, especially enum variant discovery via `EnumValueDescriptor`.

Because this project is pre-production, this issue does not require preserving compatibility with older loader-client payloads.

---

New sweetest coverage:
- Loading an inverse relationship from JSON and verifying the saved graph exposes the declared counterpart, not the inverse.
- Inverse relationship lookup where endpoint descriptors are concrete subtypes but the RTD is declared on abstract ancestors, exercising the `Extends` walk.
- Enum variant loading via `VariantOf`, verifying that the saved graph exposes `Variants`.
- End-to-end core schema load verification for `OperatorCategory` variants discoverable through `EnumValueDescriptor`.

Existing coverage to preserve:
- The Book/Person inverse relationship fixture continues to pass.
- Existing `DescribedBy` and `InverseOf` bootstrap behavior continues to pass.

Unit-level coverage on the guest resolver for the new `Extends`-walking lookup helper.

---

- `IsDeclared` is removed from `LoaderRelationshipReference` (wire, transient, parser, builder, schema JSON).
- Guest loader determines declaredness by walking the relationship type descriptor's `Extends` chain.
- Guest loader rewrites inverse relationships to their declared counterparts before staging.
- Guest inverse RTD lookup walks `Extends` on both endpoint descriptors and finds RTDs declared on abstract supertypes.
- `VariantOf` in schema JSON loads as a declared `Variants` relationship.
- Existing inverse relationship behavior (Book/Person, `DescribedBy`, `InverseOf`) remains intact.
- Sweetest coverage proves enum variants are discoverable through declared-side relationships after core schema load, including `OperatorCategory`.
- No schema JSON changes are required to work around loader behavior.
- No remaining references to a client-asserted declaredness flag in the codebase.

owleyeview:
### Summary

- **Removes `IsDeclared` from `LoaderRelationshipReference`** — the client-side stub that unconditionally tagged every relationship as declared, causing inverse relationships (e.g. `VariantOf`) to be persisted as-is instead of being rewritten to their canonical declared counterpart (`Variants`).
- **Moves declaredness resolution to the guest**, where the active type graph is authoritative. The resolver now classifies each authored relationship name as declared or inverse by walking the matching RTD's `Extends` chain, then rewrites inverse relationships (flipped endpoints, declared name) before staging.
- **Generalizes inverse RTD lookup to walk `Extends` on endpoint descriptors**, so RTDs declared on abstract supertypes are found when concrete subtypes are the actual endpoints. Adds `find_relationship_type_for_endpoints` with `optional_descriptor_key` so ancestors lacking a `Key` property are skipped rather than aborting the search.
- **Adds `RelationshipDirection` classifier** (`classify_relationship_direction`) in `holons_core::descriptors` alongside the existing `validate_extends_chain_reaches` pattern; both now delegate to the shared `search_extends_chain` helper, eliminating the pre-existing duplication.
- **Adds `Extends` as a bootstrap pass (Pass-2b)** between `DescribedBy` (2a) and `InverseOf` (2c), so descriptor ancestry is queryable before the general classification pass (2d) runs.
- **Corrects core schema JSON** — fixed incorrect `InstanceProperties`/`InstanceRelationships` references and added missing `UsesKeyRule`/inverse entries discovered during end-to-end testing.

### Test plan

- [x] `npm run test:unit` — unit tests for `classify_relationship_direction` (direct, multi-step, malformed), `find_relationship_type_for_endpoints` (concrete hit, abstract-ancestor hit, no-hit), and `try_resolve_by_type_graph` (no partial writes on classification failure)
- [x] `npm run sweetest` — all existing fixtures pass: `load_holons_internal_test` (inverse LRR maps to declared edge, link count verified), `load_book_person_inverse_schema_test`, `load_core_schema_test`
- [x] Core schema verification asserts `DeletionSemantic --Variants--> [Allow, Block, Cascade]` and `OperatorCategory --Variants--> [Equality, Ordering]` after load, proving `VariantOf` in JSON is correctly rewritten to the declared side
- [x] No remaining references to `IsDeclared` anywhere in the codebase


evomimic (changes requested):
## Review Findings

[P2] `LessThanOperator` is assigned to the wrong category

In `host/import_files/map-schema/core-schema/MAP Schema Types-map-core-schema-operator-schema.json` around lines 421-432, `LessThanOperator` is declared with `operator_category = "OperatorCategory.Equality"`, but its own description says it is an ordering operator (`lhs < rhs`). That misclassifies the core schema data loaded by this PR. The new enum-variant assertions do not catch this because they only verify that `OperatorCategory` has the expected variants, not that operators reference the correct category.


Everything else looks good.

PR #505, opened 2026-05-20
holochain 0.6 upgrade

### LINKED ISSUE #504: holochain-0.6-upgrade

nphias:
-followed official upgrade guide 0.5-0.6

updated flake.nix ( be sure to do nix flake update before you do nix develop)
removed sbd server, webrtc.  p2p messaging now with iroh
updated plugin api for 0.6.1
updated hdk guest for 0.6x hdk api (get_links, delete_link)
manifest updates for 0.61 all yaml files
hack in happ files to allow uuid and getrandom to work with wasm
fixed sweetest build pinned dependencies

dev_mode build with warm cache now down to 3 seconds
cold cache 7 seconds
non- dev_mode build 17 seconds

future changes:  there is no build isolation for the tauri-holochain-plugin.. this means that crate dependencies are affected in other packages..   ideally the plugins should be in a separate workspace... for now we only have one tauri plugin that is a core part of the system ..so its not of great concern


evomimic:
The CI failure looks like a build-script/config mismatch in the Holochain 0.6 upgrade.

CI is failing in `npm test` when it reaches:

    npm run test:unit:no-run -w map-happ

which runs:

    cargo test --manifest-path Cargo.toml --workspace --target wasm32-unknown-unknown --no-run

That path does not set the `getrandom` backend cfg. The PR currently sets it only in `happ/package.json` for `build:zomes`:

    RUSTFLAGS='--cfg getrandom_backend="custom"'

So `npm run build:happ` can compile the zomes, but the wasm unit-test no-run path compiles `getrandom 0.3.4` for `wasm32-unknown-unknown` without a backend and hits the CI error.

Recommended fix: make the backend cfg part of Cargo config instead of one npm script.

Add this at the repo root:

    .cargo/config.toml

with:

    [target.wasm32-unknown-unknown]
    rustflags = ['--cfg', 'getrandom_backend="custom"']

Then either remove the inline `RUSTFLAGS=...` from `happ/package.json` or leave it only temporarily, but the checked-in Cargo config should be the source of truth so `build:zomes`, `check`, and `test:unit:no-run` all compile with the same wasm target settings.

One small caveat: I would not put this under `happ/.cargo/config.toml` unless `happ/package.json` is also changed, because `clean:rust:deep` currently deletes `happ/.cargo`:

    "clean:rust:deep": "npx rimraf target .cargo && cargo clean"

Validation after the change:

    nix develop --command bash -c "npm run test:unit:no-run -w map-happ"
    nix develop --command bash -c "npm run build:happ"

This matches the Holochain 0.6 setup: `hdk 0.6.1` defines `__getrandom_v03_custom` for wasm and routes it through HDK `random_bytes`, so `getrandom_backend="custom"` is the right backend. It just needs to be applied consistently across all wasm Cargo invocations.

nphias:
unfortunately the only way i could get it to work without touching the package.json is under .cargo

evomimic:
### Approved. The MAP is now successfully upgraded to Holochain 0.6 and this PR looks ready to merge.

I did push one tweak to `happ/package.json` file so that the deep clean script no longer deletes `happ/.cargo/config.toml`.  Without that tweak, builds failed after clean.

II re-validated the Holochain 0.6 upgrade after the follow-up fix to `happ/package.json`. 

Verified after a clean environment cycle:

- `npm run clean`
- exited Holonix
- fresh `nix develop`
- `npm run build:happ`
- `npm run build:host`
- `npm start`
- `npm test`

All passed locally, and GitHub CI is green.

Issue #546, opened 2026-06-12
Compare saved sweettest holon graphs by definitional relationships


evomimic:
### 1. Summary (Required)

**What is the enhancement?**  
Introduce a sweettest-side utility for deriving a holon's definitional relationship graph and use it in saved-content comparison, so `MatchSavedContent` compares expected vs. persisted holons by essential content plus definitional relationships rather than exact equality across all saved edges.

---

### 2. Problem Statement (Required)

**Why is this needed?**  
Issue #442 makes commit Pass 2 materialize inverse `SmartLink`s during relationship persistence. That means persisted target holons can now carry additional inverse edges that the fixture phase never staged.

The current saved-content assertion path does not distinguish between:
- definitional relationships that should participate in structural equality
- non-definitional relationships that may appear as commit-generated persistence artifacts

Today `execute_match_db_content()` rebinds saved `SmartReference`s into a fresh assertion context and then delegates to `ExecutionReference::assert_essential_content_eq()`, which calls `compare_holon_graph_eq(...)`. That helper:
- compares `EssentialHolonContent`
- fetches `all_related_holons()` on both sides
- requires exact relationship-name and member-count equality across all non-empty relationships
- recursively matches the full relationship graph

This exact-match behavior now fails for expected saved snapshots when the persisted graph includes additional non-definitional inverse edges such as:
- inverse author/book edges materialized from declared relationships
- successor/predecessor-adjacent bookkeeping edges introduced during version commits

Repository-grounded context:
- saved-content executor: `tests/sweetests/tests/execution_steps/match_db_content_executor.rs`
- shared harness comparator: `tests/sweetests/src/harness/execution_support/execution_reference.rs`
- descriptor-side definitional accessor already exists via `RelationshipDescriptor::is_definitional()`
- inverse relationships are constrained by schema and are not expected to be definitional

Without a narrower comparison policy, sweettest saved-content assertions will continue to conflate expected structural identity with additional persisted navigational edges.

---

### 3. Dependencies (Required)

**Does this depend on other issues or features?**  
- Adjacent to Issue #442 Auto-Commit Inverse SmartLinks During Relationship Persistence (Commit Pass 2)
- Depends on existing descriptor-backed relationship metadata being available through current descriptor APIs
- No separate DevDocs track artifact is being used for this issue; this issue is grounded directly in current repository behavior and the adjacent #442 change context

---

### 4. Proposed Solution (Required)

**How would you solve it?**  
[...]

PR #547, opened 2026-06-12
442 auto commit inverse smartlinks during relationship persistence commit pass 2

### LINKED ISSUE #442: Auto-Commit Inverse SmartLinks During Relationship Persistence (Commit Pass 2)
### 1. Summary

Extend Pass 2 of `commit_functions::commit()` so that when a declared relationship is persisted as a `SmartLink`, the commit logic also resolves and persists the corresponding inverse `SmartLink` automatically.

This enforces the MAP schema contract that all declared relationships are **bidirectional and materialized**, and ensures graph traversal works correctly from both directions.

---

### 2. Problem Statement

The MAP type system defines `DeclaredRelationshipType` and `InverseRelationshipType` as a **paired schema contract**:

- A declared relationship links to its inverse via `HasInverse`
- The inverse links back via `InverseOf`
- This guarantees bidirectional traversal of relationships

However, Pass 2 of `commit_functions::commit()` currently:

- Persists only forward `SmartLink`s from the declaring side
- Does not resolve or persist inverse relationships
- Does not enforce the schema requirement that inverses must exist

As a result:

- Graph traversal from the target side may fail silently
- Schema guarantees are not enforced at persistence time
- Relationship consistency depends on caller behavior (which is incorrect)

Importantly:

> Inverse relationships are **not directly writable via the MAP API**

Therefore:

- Callers cannot and should not stage inverse relationships
- Commit-time population is the **only valid mechanism** for materializing inverse links
- Missing inverses represent a **schema contract violation**

---

### 3. Scope

This issue focuses exclusively on:

- Resolving inverse relationship names from the type system
- Persisting inverse `SmartLink`s during Pass 2
- Enforcing declaring-side-only mutation semantics

This issue does **not** include:

- Changes to the `StagedHolon` state machine
- Relationship source anchoring refinements (see Issue 2)
- Duplicate detection or idempotency behavior (see Issue 3)

---

#### 4.1 Commit-Side Inverse Resolution

For each declared relationship collection being committed:

1. Determine the source `LocalId` used for forward `SmartLink` persistence  
2. Follow `DescribedBy` to the source holon’s `TypeDescriptor`  
3. Follow `InstanceRelationships` to candidate `DeclaredRelationshipType` descriptors  
4. Match the relationship name against the descriptor’s `type_name`  
5. Follow `HasInverse` to the corresponding `InverseRelationshipType`  
6. Read the inverse descriptor’s `type_name` to determine the inverse relationship name  
7. For each forward relationship member:
   - Persist the forward `SmartLink`
   - Persist the inverse `SmartLink` with endpoints reversed  

---

#### 4.2 Declaring-Side-Only Mutation Rule

- Relationship mutations originate only from the **declaring (forward) side**
- Inverse relationships are **never staged or directly written by callers**
- Commit is responsible for inverse materialization

---

#### 4.3 Source Selection (Current Behavior)

This issue reuses the existing logic for determining the source `LocalId` of forward `SmartLink`s.

⚠️ Note:

- This behavior will be **refined in Issue 2**, which introduces explicit staged states (`ForUpdateNewVersion`, `ForUpdateGraphOnly`) and a formal relationship anchoring rule
- When Issue 2 lands, both forward and inverse persistence must derive their source from that rule rather than assuming the staged holon is always the source

---

#### 4.4 Bootstrap Safety

Inverse resolution must work when schema holons are:

- staged (during bootstrap), or  
- already persisted  

Traversal should:

- read staged relationship maps when available  
- fall back to persisted graph traversal otherwise  

The invariant is:

> Schema closure must be resolvable from staged or saved state

---

#### 4.5 Schema Invariant

All `DeclaredRelationshipType` descriptors are expected to have an inverse.

If:

- a declared relationship is matched, and  
- no `HasInverse` target can be resolved  

then:

- this must be treated as a **schema error**, not silently ignored

---

#### 4.6 External Target Constraint

Inverse persistence is only valid when the target holon is local.

If a relationship target resolves to a non-local holon:

- forward `SmartLink` may still be written (if supported by current logic)  
- inverse `SmartLink` must **not** be written locally  

This defines a boundary of schema contract enforcement.

---

### 5. Scope and Impact

| Component | File(s) | Impact |
|-----------|---------|--------|
| Commit Pass 2 | `commit_functions.rs` | Primary implementation site |
| SmartLink persistence | `smartlink_adapter.rs` | Likely unchanged |
| Type graph traversal | descriptor resolution logic | Extended for inverse lookup |

Out of scope:

- StagedHolon state model changes (Issue 2)  
- Duplicate detection and idempotency (Issue 3)  
- Inverse deletion semantics  
- Cardinality enforcement  
- External/proxied inverse propagation  

---

### 6. Testing Considerations

- Verify forward + inverse traversal for committed relationships  
- Verify inverse resolution works with staged schema (bootstrap)  
- Verify inverse resolution works with persisted schema  
- Verify missing `HasInverse` produces a schema error  
- Verify external targets do not cause commit failure  

---

### 7. Definition of Done

- [ ] Declared relationships automatically produce inverse `SmartLink`s at commit  
- [ ] Inverse relationships are never staged or directly written by callers  
- [ ] Inverse resolution works against both staged and persisted schema  
- [ ] Missing `HasInverse` is treated as a schema error  
- [ ] External targets do not trigger invalid inverse writes  
- [ ] Existing tests pass without regression  
- [ ] At least one end-to-end test validates bidirectional traversal after commit  

---

### 8. Key Design Principles

- Inverse relationships are **schema-enforced, not caller-supplied**  
- All mutations originate from the **declaring side**  
- Commit enforces **schema contracts**, not just mechanical persistence  
- Inverse persistence must mirror forward persistence using the same resolved source

owleyeview:
# Auto-commit inverse SmartLinks during relationship persistence (Commit Pass 2)

Closes #442 (Issue 1 of the three-issue relationship persistence split).

## Summary

Commit Pass 2 now persists **both directions** of every committed declared relationship: the forward SmartLink as before, plus the corresponding inverse SmartLink materialized automatically on the target. Inverse relationships remain schema-enforced, not caller-suppliable — callers still stage only declared relationships, and commit resolves the inverse through the descriptor layer.

This makes relationship persistence **strict**: a staged relationship that cannot be resolved against the source holon's effective schema surface causes the commit to report `CommitRequestStatus = Incomplete` rather than silently persisting a one-way edge. Freeform/undescribed relationships are still fine in-memory (transient/staged ergonomics are unchanged) — they are rejected at commit time only.

## Production changes

**Commit Pass 2** (`happ/crates/holons_guest/.../commit_functions.rs`)
- `RelationshipCommitSource` centralizes forward source selection so forward and inverse persistence share one anchor — Issue 2 (version anchoring) can later retarget both in a single place.
- For each committed relationship collection: resolve the inverse relationship name (outside the collection read lock, avoiding re-entrant locking on `DescribedBy`), persist the forward SmartLink, then persist the local inverse SmartLink with endpoints reversed and source-key smart properties for saved-relationship reconstruction.
- `require_local_target` guard: external (multi-space) targets fail with `NotImplemented` before either direction is written. This issue guarantees **local** inverse materialization only; the guard is the single multi-space seam.
- Any inverse resolution/persistence failure flows through the existing first-error path → `Incomplete`.

**Descriptor layer** (`shared_crates/holons_core/src/descriptors/`)
- `inverse_resolution.rs` — `resolve_inverse_relationship_name`: prefers materialized `HasInverse`; falls back to scanning transaction-local staged `InverseRelationshipType` descriptors whose `InverseOf` points back at the declared descriptor (bootstrap-safe for the core-schema load, where `HasInverse` doesn't exist yet because commit itself materializes it). Requires exactly one match; missing or ambiguous resolution is a schema error. 12 unit tests.
- `relationship_surface.rs` — `effective_relationship_declaration`: resolves writable relationship declarations over the holon's full effective descriptor lineage (MAP Type System v1.2 `DescribedBy` + `Extends` semantics), not just the direct descriptor — required for descriptor holons whose structural relationships are licensed through their own `Extends` chain. 4 unit tests.

**Core schema** (`host/import_files/map-schema/core-schema/`)
- Added missing `InstanceRelationships` declarations so the meta-tower fully self-describes under strict resolution (commit of descriptor holons' structural relationships requires reachability from the effective surface).
- Added `Predecessor`/`Successor` relationship type pair, declared at `MetaHolonType` so any described holon resolves them.

## Test harness changes (reviewers: please note)

1. **Some fixtures now expect failure.** Strict Pass 2 means committing freeform relationships on undescribed holons is an error by design. `simple_add_remove_related_holons_fixture` is the canonical rejection case: its commit asserts `CommitRequestStatus = Incomplete`. Pass 1 still saves the holons (DB-count assertions remain valid); no relationship SmartLinks persist.

2. **Fixtures that commit relationships must now load the core schema first.** Successful relationship persistence requires the source holon's effective schema surface to declare the relationship, so schema-governed commit tests load core schema (and the Book/Person test schema where relevant) before staging. `stage_new_version_fixture` is the main migration: it loads both schemas, describes the Book via the loaded `Book.HolonType` (whose `Extends` chain reaches `MetaHolonType`, where `Predecessor` is declared), and derives DB counts from the load-metrics constants. In-memory-only ergonomics fixtures stay schema-free.

3. **`MatchSavedContent` can no longer guarantee an exact match between nursery content and saved content.** Commit adds inverse SmartLinks on *target* holons that fixtures never staged (e.g. Person gains `Authors`, a prior Book version gains `Successor`, `Book.HolonType` gains `Instances`). The graph comparison was relaxed to superset semantics (every expected edge must exist in the saved graph; extra saved edges are tolerated). This is an **interim policy**: #546 (already filed) will replace it with definitional-relationship comparison, branched off this branch and PR'd back in. Inverse correctness is asserted by dedicated traversal steps instead (see below), not by the general matcher.

Supporting harness additions:
- **Commit/load status assertions**: `add_commit_step` takes an `ExpectedCommitStatus`; loader steps/executors assert `LoadCommitStatus`. Previously an `Incomplete` commit was an `Ok` dispatch and passed silently — these close that gap suite-wide.
- **`LookupSavedHolonByKey` step** (`TestHolonState::SavedLookup`): fixtures can resolve loader-saved holons (e.g. `Book.HolonType`) by key at execution time and use them as relationship targets. Expected snapshots for these are key-only stubs (marker property), matched by key without recursion and excluded from fixture counts.

## Test restructuring

- `load_holons_internal_fixture` narrowed to **loader-controller behavior** (empty bundle, nodes-only, untyped-relationship rejection, duplicate-key failure, metrics/error holons, `LoadCommitStatus`). The inline mini-schema typed/inverse success cases were removed — a tiny inline schema cannot satisfy strict Pass 2, and growing it would clone core schema.
- Typed/inverse **success** coverage moved to schema-backed loader-client flow: `load_book_person_inverse_schema_fixture` now also commits described Book/Person instances with `AuthoredBy` and asserts **bidirectional traversal** post-commit (forward `AuthoredBy`/`DescribedBy`; commit-materialized `Authors`/`Instances`).
- **Bootstrap coverage**: after core-schema load, representative materialized `HasInverse` edges are asserted (`InstanceProperties→InstancePropertyFor`, `Predecessor→Successor`, and the self-referential `InverseOf→HasInverse` pair), proving the staged-`InverseOf` fallback resolved the bootstrap circularity. The Book/Person verification likewise asserts `AuthoredBy.has_inverse() → Authors`.
- Loader `LinksCreated` semantics unchanged: it counts loader-resolver-staged relationships and does **not** double-count commit-materialized inverses.

## Validation

- All 13 dance-test cases pass (`npm run sweetest`), including the re-enabled full suite.
- `npm run fmt:check`, `npm run check`, `npm run test:unit` (241 TS tests + Rust unit tests) clean.

## Out of scope / follow-ups

- **#546** — definitional-relationship saved-content comparison replaces the interim superset matching (branch off this branch, PR back). Includes a targeted Predecessor/Successor traversal assertion for the version fixture, and depends on correcting `IsDefinitional` flags in the core schema (several spec-listed definitional relationships are currently `False`).
- Issue 2 of the split — action-oriented staged states / version anchoring (`RelationshipCommitSource` is the single retarget point).
- Issue 3 of the split — duplicate detection and idempotent SmartLink persistence.
- Cross-space inverse propagation (`require_local_target` is the seam).


evomimic (dismissed):
## Review Findings

Thanks for the substantial rework here. I think the PR is close, and I'm good with the saved-content comparison change being handled separately under #546, so I am not treating that as a blocker for #547.

I do still see one merge blocker for Issue #442 itself:

1. `tests/sweetests/tests/dance_tests.rs`: the new `DanceTestStep::AssertRelatedHolons` variant is not handled in the main test-step dispatcher. This currently breaks `npm test` locally with a non-exhaustive match compile error, so the branch is not yet green on the full local validation path.

## Policy Clarification

The code as delivered rejects external targets with `HolonError::NotImplemented` before persisting a one-sided relationship.  This is at odds with what Issue 442 specified.  Issue #442 text described non-local targets as a case where the forward write may proceed while inverse materialization is skipped.

However, I think the code is correct and the Issue spec is wrong and I have updated the descriptors design spec to v1.3 and clarified the  relationship persistence policy:

- inverse relationships are part of the semantic meaning of the committed relationship occurrence
- the runtime should not silently persist only the forward side when inverse materialization is required
- unresolved non-local inverse materialization should be surfaced as deferred semantic work with detection and remediation, not treated as successful steady-state one-sided persistence

Given that clarification, the code is in line with the policy.

## Notes

- `shared_crates/holons_core/src/descriptors/inverse_resolution.rs`: the staged `InverseOf` fallback appears architecturally justified here. Pass 2 operates on `StagedHolon`s that have already moved to `Committed(local_id)` in their internal `staged_state`, while still retaining their staged relationship maps. That means descriptor holons and their targets may already be committed and locally addressable even when descriptor relationships such as `HasInverse` or `InverseOf` have not yet been materialized as persisted SmartLinks due to unordered Pass 2 relationship commit timing. In other words, consulting staged `InverseOf` can be the correct same-transaction source of truth for schema closure. The remaining question is narrower: whether the fallback is appropriately scoped to the intended same-transaction descriptor closure, rather than whether staged fallback should exist at all.
- I am explicitly not counting the broader `MatchSavedContent` semantics change as a blocker for this PR because that follow-up is already being handled in child work under #546, which seems like the right place to finish that part cleanly.

## Validation

Manual verification reported on this branch:
- `npm run clean`: passed
- `build:happ`: passed
- host build: passed
- `npm test`: failed with the missing `AssertRelatedHolons` match arm compile error above
- `npm start`: passed

CI is currently green on the PR head, but given the local `npm test` failure, I do not think the PR is merge-ready yet.

evomimic:
Looks good to me.

This lands the core Issue #442 behavior on the actual PR branch: Pass 2 now auto-materializes inverse SmartLinks for declared relationships, enforces schema-backed resolution at commit time, and adds the descriptor-side bootstrap fallback needed to establish `HasInverse` during schema closure without duplicating descriptor traversal logic in `commit_functions.rs`.

I also think the surrounding test and fixture changes are directionally right for this issue:
- undescribed/freeform relationship fixtures now fail at commit instead of silently persisting one-way edges
- schema-backed success coverage moved to the right place
- bootstrap behavior and bidirectional traversal are both exercised explicitly
- the broader saved-content matcher semantics are clearly called out as follow-up work in #546 rather than being hidden inside this PR

CI is green on the latest PR head, and the manual verification reported so far shows `build:happ`, `build:host`, and `npm start` succeeding. With the local-branch divergence corrected, this now appears aligned with the actual GitHub branch under review.

Non-blocking note: the external-target semantics in the older Issue #442 wording now seem superseded by the clarified descriptor policy, so I would review this PR against the stricter “no successful one-sided semantic relationship write” rule rather than the earlier forward-only wording.

PR #553, opened 2026-06-16
550 dance pr4b core schema alignment for dance input parameter holons

### LINKED ISSUE #550: Dance PR4b Core Schema Alignment For Dance Input Parameter Holons
### 1. Summary (Required)

**What is the enhancement?**  
Align the MAP Core Schema with the new-world dance input model by defining explicit schema support for request-parameter holons used by canonical `DanceInvocation` execution.

This follow-up captures schema work deferred during Dance PR4 implementation. The current implementation proved that request-driven dances such as `DeleteHolon` need a clearer core-schema contract for input parameters than the present generic `DanceType.RequestType -> HolonType` surface provides.

The main schema-level outcomes should be:

- define a clear parameter-holon pattern for request-driven dances
- allow reusable parameter descriptor types such as a `HolonId` projection-style type
- expose input-parameter semantics more explicitly than the current `RequestType` name suggests
- align `HolonId` parameter encoding with an appropriate value family rather than relying on a temporary string-serialized bridge in runtime code

---

**Why is this needed?**  
Dance PR4 landed a working `DanceV2` path for host-local core dances, including `DeleteHolon`, without changing canonical core-schema import files. That was the right implementation posture for PR4, but it required a temporary runtime bridge because the schema is not yet expressing dance input parameters as clearly or as directly as the execution model now expects.

Implementation work surfaced several specific gaps:

- request-driven dances are semantically different from targeted dances
- `DeleteHolon` should receive a `HolonId` as an input parameter, not require a resolved target holon at ingress
- the current `DanceType.RequestType` relationship name does not clearly communicate that it is serving as the descriptor for invocation input parameters
- the current schema does not yet provide a clean reusable projection-style parameter descriptor for cases like “a request containing only HolonId”
- the imported schema models `HolonId.PropertyType` against a value family that is not yet aligned with the runtime representation used in the landed PR4 bridge

The result is that the current runtime must do something intentionally provisional:

- build a transient request holon for `DeleteHolon`
- store `HolonId` in a string-serialized form
- recover typed `HolonId` from that bridge inside a typed request wrapper

That is good enough for PR4 cutover, but it is not the intended long-term contract between schema and runtime.

Without this enhancement:

- request-parameter modeling remains under-specified at schema level
- core dances will continue to need ad hoc bridging logic for parameter holons
- reusable request descriptor patterns will remain unclear
- future guest-only and extension-dance work will inherit avoidable ambiguity around dance inputs

---

Yes.

This issue is downstream of the Dance PR4 implementation work that surfaced the schema gap during real cutover of core dances to `DanceV2`.

Repository-grounded dependency assessment:

- Dance PR4 command ingress and static execution alignment is now sufficiently real to expose the schema mismatch concretely
- this issue should coordinate with the authoritative schema source and regeneration flow rather than patching generated/imported JSON directly in `map-holons`
- if there is separate value-family/runtime alignment work for bytes-like or id-like values, that work may need to land alongside or immediately before the final runtime cleanup for this issue

This issue should not absorb:

- broader query/navigation dance work from Dance PR5
- guest-only dance execution pattern design
- broad runtime removal of transitional compatibility surfaces unrelated to dance input parameters
- unrelated descriptor-surface work

---

**How would you solve it?**  
Update the authoritative MAP Core Schema definition to make request-parameter holons explicit and reusable, then regenerate downstream schema imports and remove the temporary runtime bridge used by `DeleteHolon`.

Recommended implementation shape:

- Rename the dance input relationship family from `RequestType` / `RequestTypeFor` to `InputParameters` / `InputParametersFor`.
  - Replace the declared relationship `(DanceType.HolonType)-[RequestType]->(HolonType)` with `(DanceType.HolonType)-[InputParameters]->(Projection.HolonType)`.
  - Replace the inverse relationship `(HolonType)-[RequestTypeFor]->(DanceType.HolonType)` with `(Projection.HolonType)-[InputParametersFor]->(DanceType.HolonType)`.
  - Update `DanceType.HolonType.InstanceRelationships` to reference `InputParameters` instead of `RequestType`.
  - Update descriptions and downstream accessor/runtime names to treat this relationship as “descriptor of the invocation’s input-parameter holon”.

- Tighten the role of `Projection.HolonType` as the schema-owned base for parameter holons.
  - Mark `Projection.HolonType` as an abstract base type.
  - Update its description to explicitly state that projection-shaped holons can serve as dance input-parameter bodies in addition to response and projected-result shapes.
  - Keep concrete parameter shapes modeled as projection subtypes that declare their structure through `InstanceProperties`, following the existing `CommitResponse.Projection` pattern.

- Add a reusable parameter projection for holon-id input.
  - Add `HolonId.Projection` as a concrete subtype of `Projection.HolonType`.
  - Define its shape through `InstanceProperties` with exactly one required property: `HolonId.PropertyType`.
  - Treat this as the reusable parameter descriptor for any dance that takes a single holon id as input, rather than creating a `DeleteHolon`-specific property contract.

- Rebind `DeleteHolon.DanceType` to the reusable parameter projection.
  - Add an `InputParameters` relationship from `DeleteHolon.DanceType` to `HolonId.Projection`.
  - Model `DeleteHolon` as request-driven rather than target-driven.
  - The intended schema contract for `DeleteHolon` becomes: one request holon conforming to `HolonId.Projection`, no required `Target` holon.

- Introduce a dedicated canonical value type for serialized holon ids.
  - Add `HolonIdValueType` as a concrete subtype of `MapBytesValueType`.
  - Retarget `HolonId.PropertyType` from `MapBytesValueType` to `HolonIdValueType`.
  - Keep `HolonId.PropertyType` reusable and canonical, but make its value contract specific rather than treating it as generic bytes.

- Eliminate the provisional PR4 string bridge.
  - After schema regeneration, update runtime so `DeleteHolon` request holons use the canonical `HolonId.Projection` plus `HolonIdValueType` contract.
  - Remove the temporary JSON-string serialization/deserialization path currently used to carry `HolonId` through a string property.
  - Update typed parameter accessors in `holons_core` to read the canonical representation directly.

A concrete first target is `DeleteHolon`:

- its invocation should remain request-driven
- its request holon should be described by `HolonId.Projection`
- that descriptor should mean exactly “one required HolonId input parameter”
- `DeleteHolon.DanceType` should relate to that descriptor through `InputParameters`
- runtime should no longer need to serialize and deserialize `HolonId` through a string workaround

Repository-grounded areas likely to be affected after schema regeneration:

- authoritative schema source outside generated import JSON
- `host/import_files/map-schema/core-schema/` regenerated artifacts
- `shared_crates/holons_core/src/descriptors/dance_descriptor.rs`
- `shared_crates/holons_core/src/dances/contract.rs`
- `shared_crates/holons_core/src/dances/implementations/delete_holon.rs`
- any descriptor/runtime accessors currently named around `RequestType`
- tests that currently assume the transitional request bridge

---

**What does this impact?**  
This enhancement impacts:

- MAP Core Schema definition for dance input parameter modeling
- schema regeneration/import flow
- descriptor semantics for request-driven dances
- runtime construction and decoding of dance request holons
- future core and extension dance implementations that need parameter holons
- cleanup of transitional PR4 runtime bridging code

It should improve:

- clarity of the dance input contract
- reuse of parameter descriptor types across dances
- alignment between schema intent and runtime implementation
- extensibility for non-core dances whose request shapes must be modeled descriptor-first
- readability of the distinction between targeted dances and request-driven dances

It should not in this issue:

- redesign the entire guest-side dance surface
- introduce query/navigation dance semantics
- force migration of every existing legacy dance in one step
- directly edit generated core-schema JSON by hand as the source of truth

---

- Can it be validated with existing test cases?
  - Partly.
  - Existing descriptor tests and dance runtime tests provide a starting point for validating request-type lookup and invocation construction.
- Do new test cases need to be created?
  - Yes.
  - Add schema/descriptor tests proving the new parameter-holon descriptor shape is discoverable and structurally correct.
  - Add runtime tests proving `DeleteHolon` request construction uses the canonical schema-aligned parameter holon.
  - Add runtime tests proving `HolonId` is read from the canonical representation, not from the transitional string bridge.
  - Add regression tests for request-driven dances that have a request holon and no target holon.
- Are there specific areas in the test ecosystem impacted by this enhancement?
  - Yes.
  - `shared_crates/holons_core/src/descriptors`
  - `shared_crates/holons_core/src/dances`
  - `host/crates/map_commands_runtime`
  - schema contract tests and any regeneration validation around imported core schema artifacts

Testing boundary note:

- this issue should explicitly prove schema-to-runtime alignment for request-parameter holons
- once the canonical schema contract lands, the temporary JSON-string bridge should be removed and no longer treated as acceptable coverage

---

This enhancement is complete when all of the following are true:

- the authoritative core schema explicitly supports the intended request-parameter holon pattern for request-driven dances
- the dance input relationship semantics are clarified in the schema and accompanying design docs
- a reusable parameter descriptor exists for the `DeleteHolon` input shape
- `DeleteHolon` is described in schema as request-driven rather than target-driven
- `Projection` or an equivalent schema-owned pattern can explicitly describe parameter properties where needed
- the `HolonId` input parameter uses a canonical schema-aligned value representation
- regenerated/imported schema artifacts are updated in `map-holons`
- runtime `DeleteHolon` request handling no longer relies on the transitional string-serialization bridge
- tests cover descriptor shape, runtime request construction, and runtime request decoding for the canonical representation

---

<details>

- Keep the current PR4 bridge indefinitely.
  - Rejected because it encodes schema mismatch into runtime and leaves request-parameter semantics unclear.
- Model `DeleteHolon` as a targeted dance.
  - Rejected because the desired input is a `HolonId` parameter, not a pre-resolved affording holon.
- Add one bespoke request descriptor per dance without a reusable parameter pattern.
  - Rejected because common parameter shapes such as `HolonId` should be reusable across dances.
- Solve this only in runtime code without schema changes.
  - Rejected because the gap is fundamentally a schema-contract problem revealed by runtime implementation.

---

### 9. Risks or Concerns

**What could go wrong?**  

- renaming or reframing `RequestType` could ripple into more code and docs than expected
- value-family alignment for `HolonId` may depend on broader runtime support not yet fully landed
- if projection-style parameter types are introduced too narrowly, they may solve only `DeleteHolon` and not generalize well
- if the schema remains ambiguous about targeted versus request-driven dances, future implementations may drift back into mixed patterns

---

### 10. Additional Context

**Any supporting material?**  

This issue was discovered during implementation of Dance PR4 client-ingress `DanceV2` cutover work.

Repository-grounded implementation findings:

- `DeleteHolon` now executes correctly through `DanceV2`
- its landed implementation uses a transient request holon and a typed runtime wrapper
- because schema changes were explicitly deferred, the current runtime temporarily encodes `HolonId` as JSON text in a string property
- that bridge is intentionally provisional and should be removed once the core schema expresses the intended input-parameter contract directly

Recommended sequencing:

- treat this as the schema-alignment follow-up immediately after PR4 cutover
- use `DeleteHolon` as the first exemplar
- then generalize the resulting pattern for future request-driven core and extension dances

---

</details>

owleyeview:
## Summary

Aligns the generated MAP Core Schema imports and runtime descriptor expectations with the new dance input-parameter model.

Key changes:
- Renames the dance input descriptor relationship from `RequestType` to `InputParameters`.
- Retargets dance input descriptors from generic `HolonType` to `Projection.HolonType`.
- Makes `Projection.HolonType` the abstract base for projection-shaped request/input and response/result bodies.
- Adds `HolonId.Projection` as the reusable parameter projection for single-`HolonId` inputs.
- Adds `HolonIdValueType` and retargets `HolonId.PropertyType` to it.
- Updates descriptor/runtime tests and schema verification expectations to use `InputParameters`.

## Notes

This does not convert command types into dance types. `DeleteHolon.CommandType` remains a command type in this issue; the schema work only establishes the reusable input-parameter descriptor pattern needed by request-driven dances.

## Validation

- `cargo test -p map_commands_runtime --manifest-path host/Cargo.toml`
- Generated schema JSON validates with `jq`.
- `npm run test` has all tests passing


evomimic (changes requested):
## Review Findings

1. `shared_crates/holons_core/src/dances/contract.rs:331-374` and `shared_crates/holons_core/src/dances/contract.rs:556-569` do not satisfy the core runtime cleanup required by issue #550. The issue’s contract says this PR should bind `DeleteHolon` to the reusable schema-owned `HolonId.Projection` descriptor and remove the transitional string bridge for `HolonId`. Instead, the runtime still synthesizes a local `DeleteHolonParameters` descriptor, writes `HolonId` via `serde_json::to_string(&holon_id)` into a string property, and reads it back only from `BaseValue::StringValue`. This leaves the old provisional runtime path in place even though the schema now defines `HolonId.Projection` and `HolonIdValueType`. Smallest practical fix: make `build_delete_holon` and `DeleteHolonParameters` consume the canonical schema-backed input-parameter descriptor and value representation instead of the ad hoc runtime-only contract.

2. `tests/sweetests/tests/dance_tests.rs:104-116` weakens committed validation by commenting out most sweettest cases, including `delete_holon`, transaction lifecycle, and loader-oriented coverage. For a PR whose purpose is schema/runtime alignment, reducing committed end-to-end coverage makes it harder to trust that the rename and descriptor retargeting did not introduce regressions outside the one remaining core-schema test. Smallest practical fix: re-enable the disabled cases, or replace them with committed equivalent coverage that demonstrates why the removed tests are no longer needed.

CI note: GitHub checks were green on the latest PR head SHA `187a2f69b5c1609301b0fbbfa762f7c7f4766e80`.

Manual verification note: `build:happ`, `build:host`, and `npm test` were reported passing; `npm start` was still in progress at review time.

owleyeview:
@evomimic 

I see that the runtime still uses the provisional DeleteHolon parameter bridge.

Before I change it: do you want this PR to add a canonical `DeleteHolon.DanceType` descriptor to core schema now, or should it keep `DeleteHolon` represented as a command and only update the existing host-local DanceV2 DeleteHolon path to use the schema-owned `HolonId.Projection` input descriptor?

Related: `HolonIdValueType` is bytes-like in schema, but `BaseValue` currently has no bytes variant. Should removing the string bridge include adding runtime bytes value support now, or should this PR stop at binding to `HolonId.Projection` and leave value-family runtime support to a follow-up?

evomimic:
### DeleteHolon Parameter bridge

> I see that the runtime still uses the provisional DeleteHolon parameter bridge.
> 
> Before I change it: do you want this PR to add a canonical `DeleteHolon.DanceType` descriptor to core schema now, or should it keep `DeleteHolon` represented as a command and only update the existing host-local DanceV2 DeleteHolon path to use the schema-owned `HolonId.Projection` input descriptor?

Let's just keep `DeleteHolon` as a command for now. I do not want this PR to introduce a canonical `DeleteHolon.DanceType` descriptor in core schema yet, since that pushes the command surface into a broader schema-dance redesign than this issue is meant to cover.

For the runtime cleanup, please focus on the host-local `DeleteHolon` `DanceV2` path only:
- bind it to the schema-owned `HolonId.Projection` input descriptor
- remove the provisional string-based parameter bridge on that host path

### HolonIdValueType

> Related: `HolonIdValueType` is bytes-like in schema, but `BaseValue` currently has no bytes variant. Should removing the string bridge include adding runtime bytes value support now, or should this PR stop at binding to `HolonId.Projection` and leave value-family runtime support to a follow-up?

Since we're adding HolonId as a property type, it feels like we should provide full support for its correct ValueType at the same time. We already have the foundation:

- MapBytes already exists in base_types.rs
- bytes value kinds already exist in type_kinds.rs
- the loader has a commented note explicitly waiting for BaseValue::BytesValue in loader_ref_resolver.rs (line 767)

So I think the scope is small enough that we can pull this change into this PR: 

1. Add BaseValue::BytesValue(MapBytes) in [base_types.rs](/Users/stevemelville/dev/map-proto/map-dev/map-holons/shared_crates/type_system/base_types/src/base_types.rs).
2. Upgrade MapBytes to derive at least Debug, Serialize, and Deserialize there too, because BaseValue is already serialized over IPC.
3. Teach runtime value validation about bytes in [value_descriptor.rs](/Users/stevemelville/dev/map-proto/map-dev/map-holons/shared_crates/holons_core/src/descriptors/value_descriptor.rs) and [helpers.rs](/Users/stevemelville/dev/map-proto/map-dev/map-holons/shared_crates/holons_core/src/descriptors/value_descriptor_subtypes/helpers.rs).
4. Extend TS wire types and guards for BaseValue in [references.ts](/Users/stevemelville/dev/map-proto/map-dev/map-holons/host/map-sdk/src/internal/wire-types/references.ts), since MapResultWire::Value(BaseValue) crosses IPC via [result_wire.rs](/Users/stevemelville/dev/map-proto/map-dev/map-holons/host/crates/map_commands_wire/src/result_wire.rs).
5. Add explicit HolonId <-> MapBytes conversion at the host-local DeleteHolon seam, because I don’t see that conversion implemented yet.

evomimic (changes requested):
## Review Finding

Not merge ready.

The latest commit improves the implementation. However, I still classify it as not merge-ready because of the descriptor-identity gap described below. The `BaseValue::BytesValue` work looks appropriately scoped, the tests that were previously commented out are back.

1. Right now the PR defines `HolonId.Projection` twice. Once in the imported Core Schema  and again on the fly at runtime by minting a fresh Projection + HolonId descriptor pair. Instead of creating the descriptor on the fly, we should consume the existing definition. 

Why that matters: issue 550 is about schema/runtime alignment. If runtime rebuilds a lookalike descriptor, we still have two sources of truth. If runtime binds to imported HolonId.Projection, then the schema definition is the source of truth and the host path is actually aligned to it. 

I’m not asking to move all descriptor definition to schema all at once. Just this one host-local DeleteHolon input-contract seam, because this PR already introduced that schema type specifically for this purpose.

owleyeview:
Per conversation with @evomimic, we've decided to relax the need to lookup the specific `HolonId.Projection` descriptor for now and go back to defining a 'transient' descriptor on the fly that matches the shape of the projection. 

**Problem To Document**

Calling `get_all_holons()` from `build_delete_holon` made invocation construction depend on the readability of every saved holon.

In the DeleteHolon sweetest case:

1. First delete succeeds.
2. Second delete is intended to fail with `HolonNotFound`.
3. But before delete execution, `build_delete_holon` calls `get_all_holons()` to find `HolonId.Projection`.
4. `get_all_holons()` tries to build/index a full `HolonCollection`.
5. Collection indexing calls `key()` on smart references.
6. Some smart references lack cached `Key`, so they fetch the full holon from storage/cache.
7. After delete, at least one referenced holon is missing, so global lookup fails with `NotFound`.
8. That gets wrapped as `Misc("GetAllHolons dance failed...")`.
9. The real delete path never runs, so the expected `HolonNotFound` is masked.

The underlying issue: `get_all_holons()` is too broad and too fragile for descriptor resolution. It can fail because of unrelated deleted/missing holons while trying to resolve one schema descriptor.

Future fix should wait for narrower lookup support: key-populated SmartLinks, Expand + Filter query steps, and eventually `GetHolonByKey` for a `HolonType`.

evomimic:
## Correction to Review Findings


Dance input validation should validate against the dance’s declared input contract, not require that the supplied input holon be described by the exact same TypeDescriptor holon ID.

A transient input holon may include a descriptor reference when available, but absence of that descriptor is not invalid if the receiving dance’s declared contract can be satisfied.

For this issue, it means that the dance builder needs to construct the HolonId.Projection and populate it with the HolonId. But it does not actually have to even create a descriptor, much less reference the previously created HolonId.Projection descriptor.

evomimic (approved):
Provisionally approved.

This PR is ready to merge for its original scoped goal: landing the schema-alignment rename from `RequestType` to `InputParameters`, introducing `HolonId.Projection` and `HolonIdValueType`, and adding the runtime / SDK / wire support needed for canonical `HolonId` bytes values. CI is green, and manual verification was reported successful for `build:happ`, `build:host`, `npm test`, and `npm start`.

## Cautionary Provisions

I do want to explicitly note that this approval is provisional in the architectural sense, not blocked on this PR:

- We uncovered broader `dance_v2` design questions that appear to go beyond the intended scope of issue #550 and this follow-up cleanup PR.
- In particular, we still need a clearer answer for how a submitted `DanceInvocation` identifies its dance contract in an environment-independent way, and where host-side contract resolution should occur.
- We also surfaced a deeper modeling question about whether dance input should always be `Projection`-shaped, or whether a dance may legitimately accept a reference to any `HolonType` as input.
- There is also a larger ingress question about the long-term relationship between command ingress and dance ingress, which should not be decided implicitly here.

Given the original purpose of this PR, I do not think it is fair to hold it open for those larger refactor questions. My recommendation is to merge this PR for the concrete schema and value-type improvements it delivers, and track the broader `dance_v2` contract-resolution and input-model questions in follow-up work.

PR #554, opened 2026-06-17
546 compare saved sweettest holon graphs by definitional relationships

### LINKED ISSUE #546: Compare saved sweettest holon graphs by definitional relationships

owleyeview:
## Summary

Implements Issue #546: saved sweettest graph comparison now uses descriptor-backed definitional relationship semantics instead of exact all-edge equality or broad superset matching.

This change makes `MatchSavedContent` compare:

- essential holon content
- exact definitional relationship presence/member saved identity
- relationship members by saved holon identity rather than recursively revalidating every nested member graph

Non-definitional persisted edges, including commit-materialized inverse SmartLinks from #442, no longer cause saved-content comparison failures solely because fixture snapshots did not stage those edges.

### Saved Graph Comparison

`ExecutionReference::assert_saved_content_eq()` now drives saved DB comparison for `MatchSavedContent`.

For each saved holon root, the comparator:

1. compares expected vs actual `EssentialHolonContent`
2. resolves the actual holon’s descriptor
3. derives effective instance relationships from the descriptor
4. filters those relationships to `IsDefinitional == true`
5. compares only those definitional relationship maps
6. requires exact relationship-name/member-count agreement for definitional edges
7. matches definitional relationship members by saved holon identity

Each saved fixture holon is still compared independently as a root, so member content remains covered without requiring every nested relationship occurrence to recursively reproduce the full latest graph.

This avoids brittle failures where a relationship member snapshot is slightly stale in one nested position even though that member is validated correctly as its own saved root.

### Existing In-Memory Assertions Preserved

`assert_essential_content_eq()` keeps the existing expected-subset relationship behavior for ordinary step-level assertions.

That means transient/staged command-step checks still validate expected relationships without requiring descriptor-backed definitional filtering. The new definitional semantics are scoped to saved DB content comparison.

### Fixture Expected-Target Resolution

Relationship adders now resolve expected relationship targets through `FixtureHolons` instead of embedding `token.expected_reference()` directly into the expected snapshot graph.

A new helper was added:

    FixtureHolons::resolve_expected_relationship_target(token: &TestReference) -> Result<HolonReference, HolonError>

This centralizes the rule that expected relationship graphs should embed the current expected head for the logical fixture holon identified by a target token. Execution-time target resolution remains separate: executors still resolve the tokens supplied to the step when invoking the runtime operation.

### Book/Person Fixture Fix

`load_book_person_inverse_schema_fixture` now captures the updated `person_token` after adding:

    Person --DescribedBy--> Person.HolonType

and uses that updated token as the target for:

    Book --AuthoredBy--> Person

This keeps the fixture explicit and aligned with the stricter definitional comparison semantics.

## Why This Is Needed

Issue #442 changed commit behavior so declared relationships automatically materialize inverse SmartLinks. As a result, saved holons can legitimately contain extra non-definitional navigational edges that were never present in fixture snapshots.

The prior interim saved comparison used broad superset semantics. That avoided false failures from inverse edges, but it was too permissive: it could tolerate unexpected extra definitional structure.

This PR narrows the comparison policy to the intended semantic boundary:

- definitional relationships participate in saved structural equality
- non-definitional relationships do not
- inverse SmartLinks remain validated by targeted traversal assertions, not by whole-graph equality

## Harness Design Implications

This PR exposed and partially addresses an important sweettest harness authoring issue: fixture tokens are immutable snapshots, but logical holons evolve through head advancement.

Previously, relationship adders embedded target snapshots directly from the token passed by the fixture author:

    token.expected_reference()

That made the expected relationship graph sensitive to stale tokens. If a fixture mutated a target holon and later used an older token for that target in a relationship, the expected graph could embed old target content.

With definitional comparison, this became visible. For example:

1. create/stage `Person`
2. add `Person --DescribedBy--> Person.HolonType`
3. accidentally use the older pre-`DescribedBy` `person_token` as the `AuthoredBy` target
4. saved comparison fails because `DescribedBy` is now definitional

The harness now mitigates this for relationship adders by resolving target tokens through `FixtureHolons` current head. This keeps caller intent stable: “use this logical holon as the relationship target,” not “use this exact historical snapshot as the relationship target.”

## Documentation Follow-Up Needed

The dance-test framework documentation and skill guidance should be updated to reflect this clarified model.

Recommended documentation updates:

- State that `TestReference` values are immutable snapshot tokens, but many adders interpret them as handles to a logical fixture holon through `FixtureHolons`.
- Clarify that mutating adders still return a new token and authors should continue to chain returned tokens for readability and correctness.
- Clarify that relationship adders resolve target tokens through `FixtureHolons::resolve_expected_relationship_target`, so they embed the target logical holon’s current expected head rather than the literal historical token snapshot.
- Distinguish source resolution from relationship-target expected resolution:
  - source adders clone from the appropriate source snapshot for the operation
  - relationship adders embed the current expected target snapshot for graph expectations
  - execution-time resolution remains separate
- Explain that `MatchSavedContent` compares saved roots by essential content plus definitional relationships, and matches definitional relationship members by saved holon identity.
- Note that nested relationship member content is not recursively revalidated from every edge because each saved fixture holon is compared independently as a root.
- Document that non-definitional persisted edges, including commit-generated inverse SmartLinks, are intentionally ignored by saved-content equality and should be verified through targeted traversal/assertion steps when relevant.
- Update any language implying that old tokens automatically “follow the head” in every context. The more precise rule is that harness APIs decide how a token is resolved based on intent; callers should not inspect tokens or choose head vs last-live themselves.

## Validation

- `npm run fmt:check` passed.
- `cargo check --manifest-path tests/sweetests/Cargo.toml` passed before the final helper rename.
- `npm run test:unit` passed.
- Full sweettest suite passed after these harness changes.


evomimic (changes requested):
## Review Findings

1. [P1] Definitional relationship member matching is currently key-only, so `MatchSavedContent` can miss wrong-version targets on versioned holons.

The new saved-content path in `tests/sweetests/src/harness/execution_support/execution_reference.rs` says it compares exact definitional relationship member identity, but `match_by_identity()` currently delegates to `compare_identity()`, which only compares `key()` values.

Why this is a blocker:

- MAP allows multiple saved versions of a holon to share the same key.
- This PR also updates fixture relationship adders to resolve targets to the current fixture head, which means the exact target version now matters.
- As written, the comparator can still accept the wrong saved version as long as the key matches.

Concrete examples:

- Expected: `Book --DescribedBy--> Book.HolonType v2`
- Actual: `Book --DescribedBy--> Book.HolonType v1`
- Result today: passes if both versions share the same key

- Expected: `Person.HolonType --ComponentOf--> Schema v3`
- Actual: `Person.HolonType --ComponentOf--> Schema v2`
- Result today: passes if both schema versions share the same key

That is a false positive on a definitional relationship, which is exactly the class of mismatch Issue #546 is supposed to catch.

Why the surrounding changes make this more apparent:

- `FixtureHolons::resolve_expected_relationship_target(...)` was added so relationship adders embed the current expected head rather than a stale token snapshot.
- `load_book_person_inverse_schema_fixture` was updated to capture the newer `person_token` before using it as the `AuthoredBy` target.
- Those changes only fully pay off if saved-content comparison can distinguish same-key different-version targets.

Suggested fix options:

1. Preferred: compare exact saved identity for definitional relationship members.
Use the execution-time saved reference or another version-distinguishing saved identity, rather than only `key()`.

2. Safe fallback: keep the new definitional-relationship filtering, but use recursive graph comparison for definitional members instead of key-only matching.
This is more expensive, but it preserves correctness.

3. Acceptable only if proven reliable in this codepath: compare a richer identity tuple.
For example, key plus a version-distinguishing field such as saved id or predecessor/original id where available.

Recommended regression test:

- Add a saved-content test where a definitional relationship target advances to a new version while keeping the same key.
- Assert that `MatchSavedContent` fails when actual points at the older saved version and expected points at the newer head.

Until that is fixed, I would not call PR #554 merge-ready.

evomimic:
## A Possible Formal Definition for Definitional Equivalence

Two holons are _**definitionally equivalent**_ if:

1. Their essential properties are equal.
2. They contain the same definitional relationship sets.
3. Corresponding definitional relationships target either:
    * the same saved holon identity, or
    * holons that are themselves definitionally equivalent.

 ## Relationship Identity Comparison Cases

### Case 1: Both relationship targets are saved holons

If both relationship targets are saved holons, compare their saved identities directly.

    target_a.id == target_b.id

If the IDs match, the comparison can stop at that target.

The saved HolonId acts as a stable identity anchor. Anything behind that saved holon is assumed to be the same object because both relationships point to the same persisted holon.

### Case 2: One or both relationship targets are not saved holons

If either target is transient or staged, its reference identity is not stable enough to compare directly.

In this case, compare the target holons by essential content instead.

    essential_content(target_a) == essential_content(target_b)

This comparison must recurse through the target's own definitional relationships.

The recursive walk continues until either:

1. A saved-holon identity anchor is reached, or
2. All definitional relationship targets have been compared by essential content.

### Summary Rule

For each definitional relationship target pair:

    if both targets are saved:
        compare saved HolonIds
    else:
        recursively compare essential content

This lets saved holons terminate the comparison, while staged and transient holons are compared structurally.

evomimic:
## Revised Review Findings

1. [P1] PR #554 implements definitional comparison with the wrong semantic rule, and the correct rule belongs on `ReadableHolon` rather than in the sweettest harness.

After further review, I think the finding should be stated more directly.

The new saved-content comparison path in `tests/sweetests/src/harness/execution_support/execution_reference.rs` currently treats key equality as the identity rule for definitional relationship members. That is not a valid basis for definitional equivalence.

For definitional equivalence, corresponding definitional relationship targets should compare by this rule:

- if both targets are saved holons, compare saved holon identity directly
- otherwise, compare them structurally by essential content plus their own definitional relationships
- maintain a visited-pair set so cyclic graphs terminate safely
- never use key equality as the success condition

So the issue is not that the comparator is merely “too weak.” It is using the wrong comparison model.

Why this matters:

- different saved versions of a holon can share the same key
- key equality can therefore produce false positives on definitional relationships
- that lets `MatchSavedContent` accept the wrong definitional target version even when structural identity differs

Example:

- Expected: `Book --DescribedBy--> Book.HolonType v2`
- Actual: `Book --DescribedBy--> Book.HolonType v1`
- Same key: `Book.HolonType`
- Different saved holon identity
- Current PR can pass
- Correct definitional equivalence must fail

Second example:

- Expected: `PersonType --ComponentOf--> Schema v3`
- Actual: `PersonType --ComponentOf--> Schema v2`
- Same key: `Schema`
- Different saved holon identity
- Current PR can pass
- Correct definitional equivalence must fail

I also think this should not remain sweettest-local logic.

`ReadableHolon` already defines `essential_content()`, and definitional equivalence is a phase-spanning reference-layer concern over `HolonReference`, not a test-only concern. It should be available across transient, staged, and saved references through a shared runtime API.

Suggested design direction:

- extend `ReadableHolon` with something like `is_definitionally_equivalent(&self, compare_to_holon: HolonReference) -> Result<bool, HolonError>`
- implement it in `holons_core` using:
  - essential-content equality
  - exact definitional relationship-set comparison
  - saved-id comparison when both relationship targets are saved
  - recursive structural comparison otherwise
  - visited-pair tracking for cycle safety
- have the sweettest harness call that shared API instead of defining its own key-based comparison semantics inside `ExecutionReference`

There is already a code-level hint that this logic belongs lower in the stack: `shared_crates/holons_core/src/core_shared_objects/holon/holon_utils.rs` contains a TODO about descriptor-backed definitional relationship support, and `EssentialRelationshipMap` already lives in `holons_core`.

Because PR #554 currently encodes definitional equivalence incorrectly and in the wrong layer, I do not think it is merge-ready yet.

owleyeview:
Thanks for the revised finding @evomimic. A couple updates from the current branch state:

- The saved-content definitional member matcher no longer uses key equality as the success condition.
- For definitional relationship members, `MatchSavedContent` now resolves the expected fixture snapshot through `ExecutionHolons` and compares the resulting saved `HolonId` against the actual saved target `HolonId`.
- Key-only matching remains only for saved-lookup stubs, which are harness placeholders for holons created outside the fixture ledger.

I agree with the formal equivalence rule as the right MAP-level semantic direction:

1. compare essential content
2. compare exact definitional relationship sets
3. for relationship members:
   - if both targets are saved, compare saved `HolonId`
   - otherwise recurse structurally through essential content plus definitional relationships
4. track visited pairs for cycles

I’m less sure that `ReadableHolon` can replace the sweettest harness implementation directly for this issue. The fixture expected side is not always a normal runtime saved reference. `ExpectedSnapshot`s are stored as transient fixture snapshots even when their expected state is `Saved`; the harness needs `ExecutionHolons` to map those fixture snapshot IDs to the actual committed `HolonId`s produced during execution. A plain `ReadableHolon` API would not have access to that fixture ledger. The harness also has saved-lookup stubs, which are key-only expected placeholders for schema-loaded holons outside the fixture ledger; those are test artifacts, not runtime semantics.

That said, I agree the reference layer should probably grow a first-class definitional equivalence API. A follow-up could add something like:

- `ReadableHolon::is_definitionally_equivalent(...)` or a richer diagnostic-returning comparator
- descriptor-backed definitional relationship filtering in `holons_core`
- saved-target `HolonId` anchoring
- recursive structural comparison for transient/staged targets
- visited-pair cycle protection
- tests for saved-vs-saved identity, staged/transient structural recursion, wrong-version saved targets, missing/extra definitional members, and cycles

Then `MatchSavedContent` could be migrated to use that shared comparator where both sides are ordinary runtime references, while keeping a thin harness adapter for fixture snapshots and saved-lookup stubs.

Do you think that reference-layer implementation should be pulled into #546, or should #546 stay scoped to the sweettest saved-content comparison and we open a follow-up for the runtime `ReadableHolon` equivalence API?

evomimic:
> I’m less sure that ReadableHolon can replace the sweettest harness implementation directly for this issue. The fixture expected side is not always a normal runtime saved reference. ExpectedSnapshots are stored as transient fixture snapshots even when their expected state is Saved; the harness needs ExecutionHolons to map those fixture snapshot IDs to the actual committed HolonIds produced during execution. A plain ReadableHolon API would not have access to that fixture ledger. The harness also has saved-lookup stubs, which are key-only expected placeholders for schema-loaded holons outside the fixture ledger; those are test artifacts, not runtime semantics.

I think we need to be precise here. It's true that an `ExecutionReference` is a harness-only wrapper, and I agree the reference layer should not depend on it or on `ExecutionHolons`. The part I want to distinguish more carefully is where the comparison logic actually begins.

`assert_essential_content_eq()` and `assert_saved_content_eq()` are harness entrypoints, but both resolve down to runtime `HolonReference`s before the actual graph comparison runs. From `compare_holon_graph_eq(expected: &HolonReference, actual: &HolonReference, ...)` downward, the logic is no longer inherently about fixture wrappers; it is comparing runtime references and their definitional structure.

That means the harness-specific concerns are narrow:

- mapping fixture expected snapshots to realized saved identities through `ExecutionHolons`
- handling saved-lookup stubs, which are test-only placeholders

But the core definitional-equivalence algorithm itself looks reference-layer shaped:

- compare essential content
- compare exact definitional relationship sets
- for relationship members:
  - if both targets are saved, compare saved `HolonId`
  - otherwise recurse structurally
- track visited pairs to handle cycles

So I think the right architectural boundary is:

- keep `ExecutionReference` as the harness adapter
- keep fixture-ledger lookup and saved-lookup-stub handling in the harness
- move the `HolonReference`-to-`HolonReference` definitional comparison logic into `holons_core` / the reference layer

Put differently: I’m not arguing that `ExecutionReference` itself belongs in the reference layer. I’m arguing that once the harness has normalized its inputs into ordinary `HolonReference`s, the comparison behavior from that point down is general MAP runtime semantics, not sweettest-only semantics.

On scope, I’d be comfortable with #546 staying focused on the sweettest fix if the branch now correctly uses saved `HolonId` anchoring and no longer uses keys for definitional equivalence. But I do think we should open a follow-up issue to extract the `HolonReference`-level definitional equivalence algorithm into a first-class reference-layer API, with the harness reduced to a thin adapter around fixture-only concerns.

evomimic (approved):
## Approval Review

I re-reviewed PR #554 against the narrower scope of Issue #546 and I think it is merge-ready.

The important clarification is that Issue #546 is about sweettest saved-content comparison, not about adding a general MAP-level definitional-equivalence API. Within `MatchSavedContent`, the expected side is a fixture-authored saved snapshot, and relationship members are resolved through `ExecutionHolons` to the committed `HolonId`s produced by the same fixture execution. In that context, comparing definitional relationship members by saved identity is appropriate: the assertion is checking that the saved graph points to the fixture-intended holons.

This branch addresses the original issue by:

- filtering saved-content graph comparison to descriptor-backed definitional relationships
- ignoring non-definitional persisted edges, including commit-generated inverse SmartLinks
- replacing key-based relationship-member matching with execution-ledger-backed saved `HolonId` comparison
- keeping saved-lookup stubs as a separate harness-specific key-based path for holons created outside the fixture ledger
- updating the schema flags needed for the descriptor-backed definitional relationship filter

I withdraw the earlier concern that `MatchSavedContent` must recurse structurally when saved target `HolonId`s differ. That behavior belongs in the broader reference-layer definitional-equivalence work now tracked separately in Issue #555. For this PR’s test-harness scope, a mismatched saved target id should fail because it means the persisted relationship points to a different saved holon than the fixture declared.

Verification reported:

- hApp build passes
- host build passes
- `npm test` passes
- `npm start` nominal
- all CI checks green

### Doc Updates

Additionally, 

- [x] I updated the PR description to reflect the revised scope
- [x] I updated the authoring guides and test specs in map-dev-docs per the recommendations in the PR description.

Approved for merge.

PR #568, opened 2026-06-29
560 - recovery dev mode gating

### LINKED ISSUE #560: Add a `no-recovery` startup mode for `npm start`
### 1. Summary (Required)

**What is the enhancement?**  
Add an explicit `no-recovery` startup mode for `npm start` that clears stale local recovery/session state before launching, while preserving recovery-enabled startup as the default behavior.

---

**Why is this needed?**  
`npm start` currently attempts to restore prior local recovery sessions on launch. That is desirable for crash recovery, but it becomes a dev-time blocker when the persisted recovery store contains snapshots serialized with an older staged-holon shape. In that case, startup can fail before the app finishes booting.

We need an explicit way to start from a clean recovery state without manually deleting the recovery database every time the staged serialization changes.

---

This depends on the existing local recovery store and startup path already used by `conductora`.

Relevant implementation points already in the repository:

- `host/conductora/src/runtime/init_runtime.rs`
- `host/conductora/src/setup/providers/local/setup.rs`
- `host/crates/recovery_receptor/src/storage/transaction_store.rs`
- `host/package.json`

This enhancement should not remove crash recovery entirely. It should add an explicit opt-in reset path for development and compatibility cleanup.

---

**How would you solve it?**  
Add a `no-recovery` mode to startup, exposed through the existing `npm start` workflow, that clears the local recovery store before runtime initialization.

Suggested behavior:

- default `npm start` continues to use recovery restore behavior
- `npm start` with a `no-recovery` flag clears the local recovery store first, then launches fresh
- `npm run clean` should also clear the local recovery store so it remains the canonical reset path

Implementation likely needs:

- a startup flag or environment-controlled switch in the host startup path
- a recovery-store cleanup method that deletes all local recovery state for the configured dev provider
- startup wiring that skips restore only when the explicit `no-recovery` mode is selected

---

**What does this impact?**  
This affects:

- host startup behavior
- local recovery store management
- developer workflows for `npm start` and `npm run clean`
- compatibility handling for stale recovery snapshots during local development

It should not change runtime recovery semantics for normal crash-recovery runs unless the explicit `no-recovery` mode is used.

---

- Verify default `npm start` still restores recoverable sessions when the store is valid.
- Verify `npm start` with `no-recovery` clears stale recovery state and launches successfully.
- Verify `npm run clean` removes the local recovery store.
- Add a targeted unit or integration test around the recovery-store cleanup behavior if one does not already exist.
- Confirm a stale `snapshots.db` can be cleared and startup succeeds afterward.

---

- [ ] Default startup behavior still supports crash recovery.
- [ ] A `no-recovery` startup mode exists for dev resets.
- [ ] The `no-recovery` mode clears the local recovery store before startup.
- [ ] `npm run clean` clears the local recovery store on demand.
- [ ] Startup no longer fails when an old recovery store is intentionally cleared.
- [ ] Tests or scripted verification cover both default recovery and no-recovery launch paths.

<details>

Manual deletion of `snapshots.db` works, but it is too easy to miss and too awkward for repeatable development. A flag or clean hook is more explicit and easier to remember.

---

### 9. Risks or Concerns

**What could go wrong?**  
If the cleanup path is wired too broadly, it could accidentally erase recovery data during normal startup. The reset behavior should be opt-in and clearly separated from default crash recovery.

---

### 10. Additional Context

**Any supporting material?**  
The local recovery store is created under:

    $APP_DATA_DIR/storage/local_recovery/snapshots.db

For this app, the Tauri identifier is `com.map-holons.tauri.dev`.

</details>

nphias:
## Work done

- `src/env.rs` — shared `hc_dev_mode_enabled()` replacing per-file copies
- `init_runtime.rs` — startup `restore_open_sessions()` skipped when HC_DEV_MODE on
- `local/setup.rs` — `create_snapshot_store` uses `/tmp/conductora_dev/session/` in dev mode, `app_data_dir/storage/session/` in production
- `package.json` — scripts clarified; `start:host` explicitly sets HC_DEV_MODE=1; `start:production` added; `clean:recovery` added
- `scripts/clean-recovery.sh` — cross-platform wipe of production snapshot data
- existing script `clean:hc:deep` removes all of `/tmp/conductora_dev` snapshot data

## notes

- Session recovery is a production resilience feature. Its value is entirely about surviving unexpected shutdowns — the user closes the app mid-transaction, crashes, loses power, and expects to resume on the next start. In dev mode:

  - You're restarting constantly and deliberately
  - You want a clean, predictable start every time — stale sessions from the last dev run are noise, not signal
  - The feature itself doesn't need to be exercised on every dev start
  - The only time you'd want to test it is explicitly — start in start:production, create an open session, kill the app,  restart in start:production, verify it restores. clean:recovery resets state between test attempts. That's it.

### DIFF
--- a/host/conductora/plugins/tauri-plugin-holochain/DEV_MODE.md
+++ b/host/conductora/plugins/tauri-plugin-holochain/DEV_MODE.md
@@ -17,16 +17,14 @@ Core goals:
 Dev mode is controlled by one launch-time environment variable:
 
 ```sh
-HC_DEV_MODE=1
+MAP_START_MODE=dev
 ```
 
-Truthy values are: `1`, `true`, `yes`, `on` (case-insensitive).
-
-When `HC_DEV_MODE` is unset (or falsey), runtime uses normal mode.
+When `MAP_START_MODE` is unset or set to `prod`, runtime uses normal mode.
 
 ## Runtime Behavior
 
-| Area | Normal Mode | Dev Mode (`HC_DEV_MODE=1`) |
+| Area | Normal Mode | Dev Mode (`MAP_START_MODE=dev`) |
 |---|---|---|
 | Keystore | `LairServerInProc` (in-process lair) | `DangerTestKeystore` (in-memory, ephemeral) |
 | Device seed | Created in lair under tag `"DEVICE_SEED"` | Not used; setup generates an explicit agent key via `generate_agent_pub_key()` |
--- a/host/conductora/src/env.rs
+++ b/host/conductora/src/env.rs
@@ -1,0 +1,6 @@
+pub fn hc_dev_mode_enabled() -> bool {
+    match std::env::var("HC_DEV_MODE") {
+        Ok(v) => matches!(v.trim().to_ascii_lowercase().as_str(), "1" | "true" | "yes" | "on"),
+        Err(_) => false,
+    }
+}
--- a/host/conductora/src/env.rs
+++ b/host/conductora/src/env.rs
@@ -1,6 +1,6 @@
-pub fn hc_dev_mode_enabled() -> bool {
-    match std::env::var("HC_DEV_MODE") {
-        Ok(v) => matches!(v.trim().to_ascii_lowercase().as_str(), "1" | "true" | "yes" | "on"),
+pub fn dev_mode_enabled() -> bool {
+    match std::env::var("MAP_START_MODE") {
+        Ok(v) => matches!(v.trim().to_ascii_lowercase().as_str(), "dev"),
         Err(_) => false,
     }
 }
--- a/host/conductora/src/lib.rs
+++ b/host/conductora/src/lib.rs
@@ -1,5 +1,6 @@
 // Module declarations
 mod config;
+pub mod env;
 mod map_commands;
 mod runtime;
 mod setup;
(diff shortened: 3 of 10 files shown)


evomimic:
## Review Findings

I'm OK with the change in requirement from having a separate "no recovery" flag (per my comment on Issue #560 ). 

However, I'm requesting two changes on this PR:

1. Change name of the env variable from HC_DEV_MODE to MAP_START_MODE per my comment
2. Include clean:recovery in the `npm run clean` script.

nphias:
done added MAP_START_MODE. .. how do you feel about SessionReceptor (responsible for undo/redo/checkpoints/recovery) instead of RecoveryReceptor?
[issue 570](https://github.com/evomimic/map-holons/issues/570)

evomimic:
## Approval

Approved.

This PR now satisfies the contract in issue #560:
- recovery remains available in the production-like startup path
- the start-mode naming is clearer and scoped to startup behavior
- `clean:recovery` is included in the canonical cleanup flow

Verification:
- CI passed on the latest head
- `build:happ` passed
- `build:host` passed
- `npm test` passed
- `npm start` passed
- `npm run clean:recovery` correctly removed the snapshot store

### NOTE: The app now has two startup modes:

- `npm start` runs the default development workflow.
- `npm run start:production` runs the production-like workflow.

Internally, `MAP_START_MODE` controls the underlying behavior:

- `MAP_START_MODE=dev` enables dev startup behavior.
- `MAP_START_MODE=prod` enables production-like startup behavior.

But exporting these variables to the shell that you use to start the app has no effect. The proper way to set the start mode is by choosing the proper start script.

Issue #562, opened 2026-06-26
Make `HasInverse` the declared relationship and `InverseOf` as its inverse.


evomimic:
# Make `HasInverse` the DeclaredRelationship for relationship-pair metadata

## Summary

Resolve the issue discovered during implementation of Issue #559 . 

`HasInverse` should be the declared relationship from `DeclaredRelationshipType` to `InverseRelationshipType`, and `InverseOf` should be its inverse.

### 2. Problem Statement (Required)

Only declared relationships can be directly mutated. When committing an instance of a declared relationship for a source holon, MAP must locate the relationship type's inverse and materialize the reciprocal SmartLink.

If the canonical pairing is represented only as:

    InverseRelationshipType --InverseOf--> DeclaredRelationshipType

then bootstrap can become problematic. During initial core schema load, the `InverseOf` relationship may not yet be committed, which means the loader may not be able to find the inverse relationship type at the moment it needs to create the reciprocal SmartLink.

By making `HasInverse` the declared relationship:

    DeclaredRelationshipType --HasInverse--> InverseRelationshipType

the inverse relationship type is available directly from the declared relationship type being committed.

---

### 3. Dependencies (Required)

This issue does not depend on any other issues. However, Issue 559 depends on delivery of this issue to avoid having to implement a throw-away workaround for the blockers discovered during implementation of that issue.

---

### 4. Proposed Solution (Required)

---

#### Proposed CoreSchema shape

1.  `(DeclaredRelationshipType)-[HasInverse]->(InverseRelationshipType)` should extend `DeclaredRelationshipType` instead of `InverseRelationshipType`. Note: no change is required to `(InverseRelationshipType)-[InverseOf]->(DeclaredRelationshipType)`
2. In the schema definition files, move the declaration of the inverse relationship from the inverse relationship to the declared relationship as illustrated in the following example. 

Current Schema Definition:

```
{
      "key": "(EnumValueType)-[Variants]->(EnumVariantValueType)",
      "type": "TypeDescriptor.HolonType",
      "properties": {
        "type_name": "Variants",
        "type_name_plural": "VariantsRelationships",
        "display_name": "Variants",
        "display_name_plural": "Variants Relationships",
        "description": "Declares the allowed symbolic variants of an enum value type.",
        "instance_type_kind": "TypeKind.Relationship",
        "is_abstract_type": false,
        "is_definitional": false,
        "is_ordered": false,
        "allows_duplicates": false,
        "min_cardinality": 0,
[...]

PR #573, opened 2026-07-02
536 descriptor effective surface discovery and validation helpers

### LINKED ISSUE #536: Descriptor Effective Surface Discovery And Validation Helpers
### 1. Summary (Required)

**What is the enhancement?**

Add a coherent set of **type-level methods** on `HolonDescriptor` and `ValueDescriptor` (`shared_crates/holons_core/src/descriptors`) that answer: **"What can I do with this type of holon?"**

These helpers operate on **type descriptors**, not instance `HolonReference` handles (with one deliberate exception: `available_relationships`, which filters the type-level navigable set by the state of a concrete source reference). They expose effective, inheritance-aware discovery, validation, and resolution for:

- `HolonDescriptor`: instance properties, instance relationships, afforded dances, afforded commands
- `ValueDescriptor`: afforded operators
- outbound relationship navigation validation and enumeration for query `Expand`
- effective key-rule resolution for instances of a type, via a new `KeyRuleDescriptor`

The discovery and validation helpers package existing descriptor behavior behind one discoverable API. Relationship navigation and key-rule resolution add the main new behavior.

---

**Why is this needed?**

Descriptor discovery behavior already exists, but it is spread across several method surfaces. Downstream consumers such as query `Expand`, command/dance affordance validation, and key generation would otherwise recompute descriptor semantics locally.

Existing foundations to reuse:

- `inheritance.rs`: `walk_extends_chain`, `ExtendsIter`, `ancestors`, `effective_descriptor_lineage`, `flatten_related_members`, `classify_relationship_direction`, `RelationshipDirection`
- `holon_descriptor.rs`: `instance_properties`, `instance_relationships`, `afforded_commands`, `afforded_dances`, `get_property_by_name`, `get_relationship_by_name`, `get_command_by_name`, `get_dance_by_name`, `get_inverse_relationship_by_name`
- `value_descriptor.rs`: `supported_operators`, `supports_operator`, `get_operator_by_name`, `supports_operator_by_name`
- `relationship_surface.rs`: `effective_relationship_declaration`
- `inverse_resolution.rs`: `resolve_inverse_relationship_name`
- relationship wrappers: `DeclaredRelationshipDescriptor` / `InverseRelationshipDescriptor` with `source_type()`, `target_type()`, `has_inverse()`, `inverse_of()`

The main gaps are:

1. **Outbound relationship navigation validation and enumeration.** Navigation is always outbound from the source holon. A relationship (declared **or** inverse descriptor) is navigable from type `T` iff `relationship.SourceType == T` (equals-or-extends). `InstanceRelationships` licenses only declared relationships, so there is no existing seam that answers "which relationships — including inverse relationships — are navigable from this type?" or "is this name a legal outbound relationship here?"
2. **Effective key-rule resolution.** There is no descriptor-owned Rust seam for resolving the effective `UsesKeyRule` through direct declaration plus `Extends` fallback.

---

Build on the existing `holons_core/src/descriptors` foundations listed above.

Canonical key-rule inputs come from:

- `host/import_files/map-schema/core-schema/MAP Schema Types-map-core-schema-keyrules-schema.json`

Known current rule names include `TypeNameRule.KeyRuleType`, `SchemaNameRule.KeyRuleType`, `TypeKindRule.KeyRuleType`, `EnumVariantRule.KeyRuleType`, `RelationshipRule.KeyRuleType`, `ExtendedTypeRule.KeyRuleType`, `NoneRule.KeyRuleType`, plus abstract `KeyRuleType` and `FormatRule.KeyRuleType`.

The older `shared_crates/json_schema_validation/.../metaschema-keyrules-schema.json` fixture is not authoritative.

New groundwork:

```rust
pub struct QualifiedRelationship {
    pub descriptor: RelationshipDescriptor,      // as matched: declared or inverse
    pub descriptor_direction: RelationshipDirection,
}
```

`RelationshipDirection` classifies the matched descriptor as declared or inverse. There is no separate traversal-direction axis: navigation is always outbound from the source, and inverse descriptors are first-class navigable relationships, not a syntactic alias for traversing the declared relationship backwards. Mapping syntactic inbound query forms onto outbound relationship names is a query-planner concern, out of scope here.

Out of scope:

- Query execution and access-path selection / normalization (future query optimization)
- Runtime validation inside `get_related_holons` / `related_holons` (future work; see §4.4)
- Key-string generation
- Converter/importer migration
- Command/dance execution redesign
- Hand-editing canonical core-schema import JSON
- Adding key-rule variants to `CoreHolonTypeName`
- Transaction-local target-side discovery for staged, uncommitted descriptor graphs unless PRS2 explicitly requires staged schema querying

Coordinates with:

- Query `Expand` / PRS2
- Issue 533 key generation

---

#### 4.1 Discovery surfaces

Expose effective, inheritance-aware plural accessors on descriptor wrappers (already present; consolidate rather than duplicating names):

| Method                                      | Returns                                           |
| ------------------------------------------- | ------------------------------------------------- |
| `HolonDescriptor::instance_properties()`    | `Result<Vec<PropertyDescriptor>, HolonError>`     |
| `HolonDescriptor::instance_relationships()` | `Result<Vec<RelationshipDescriptor>, HolonError>` |
| `HolonDescriptor::afforded_dances()`        | `Result<Vec<DanceDescriptor>, HolonError>`        |
| `HolonDescriptor::afforded_commands()`      | `Result<Vec<CommandDescriptor>, HolonError>`      |
| `ValueDescriptor::supported_operators()`    | `Result<Vec<OperatorDescriptor>, HolonError>`     |

#### 4.2 Validation entrypoints

Add name-based validation helpers that delegate to existing lookups:

```rust
HolonDescriptor::allows_property(name)  -> Result<PropertyDescriptor, HolonError>
HolonDescriptor::affords_dance(name)    -> Result<DanceDescriptor, HolonError>
HolonDescriptor::affords_command(name)  -> Result<CommandDescriptor, HolonError>
ValueDescriptor::affords_operator(name) -> Result<OperatorDescriptor, HolonError>
```

Prefer existing conversion traits for name arguments, such as `impl ToRelationshipName`, rather than forcing callers to construct concrete newtypes prematurely.

Relationship validation is outbound-only and takes no direction parameter:

```rust
HolonDescriptor::allows_relationship(
    name: impl ToRelationshipName,
) -> Result<QualifiedRelationship, HolonError>
```

The name may resolve to a declared relationship licensed on this type (via effective `InstanceRelationships`), or to an inverse relationship whose `SourceType` is this type (discovered through the materialized `TargetOf` index plus `HasInverse`). `QualifiedRelationship.descriptor` is the as-matched descriptor; `descriptor_direction` records whether it is declared or inverse.

#### 4.3 Navigable and available relationship enumeration

Type-level enumeration (state-agnostic):

```rust
HolonDescriptor::navigable_relationships() -> Result<Vec<QualifiedRelationship>, HolonError>
```

The union of declared relationships licensed on this type (or inherited) and inverse relationships whose `SourceType` is this type.

Reference-level enumeration (state-filtered), as a free helper in `descriptors/` (could later move onto `HolonReference`):

```rust
pub fn available_relationships(
    source_ref: &HolonReference,
) -> Result<Vec<QualifiedRelationship>, HolonError>
```

Availability filters the navigable set by source commit state, because inverse SmartLinks are only materialized at commit:

| Source reference state          | Declared | Inverse |
| ------------------------------- | -------- | ------- |
| `SmartReference` (saved)        | yes      | yes     |
| `StagedReference` committed     | yes      | yes     |
| `StagedReference` not committed | yes      | no      |
| `TransientReference`            | yes      | no      |

Discovery mechanics:

- **Declared candidates** come from forward effective `InstanceRelationships` (staged-safe, pre-commit available).
- **Inverse candidates** come from walking `TargetOf` on this type plus its `Extends` ancestors, validating endpoint compatibility (`TargetType` equals-or-extends) and licensing (the declared relationship must be present in its declared source type's effective `InstanceRelationships`), then following `HasInverse`.
- `InstanceRelationships` is the authoritative licensing surface; `TargetOf` is only a discovery/index path. Candidates that disagree in malformed schema state are rejected.
- `TargetOf` is materialized on commit, so inverse discovery requires a committed descriptor graph. An unsaved type endpoint with an empty `TargetOf` index is rejected with `UnsupportedStagedTraversal` rather than silently answering incompletely. `available_relationships` never hits this guard: uncommitted sources are declared-only by construction and the index is not consulted.

The module (`relationship_navigation.rs`) documents the separation of concerns: schema semantics (`SourceType`/`TargetType`/`HasInverse`), mutation semantics (only declared relationships are editable), navigation semantics (always outbound), runtime-state semantics (inverse requires committed source), and optimization semantics (access-path selection deferred to query planning).

#### 4.4 Future `get_related_holons` validation (documented TODO only)

`ReadableHolon::related_holons` carries a TODO describing the intended future validation:

- `InvalidRelationshipName` when the requested name is not a valid outbound relationship for the source holon's type (takes precedence).
- `RelationshipUnavailableInState` when the name resolves to a valid inverse relationship but the source is transient or staged-uncommitted.

Implementing this validation is out of scope for this issue.

#### 4.5 Effective key-rule resolution

Add `HolonDescriptor::effective_key_rule()` resolving the rule for instances of this type:

```rust
HolonDescriptor::effective_key_rule() -> Result<KeyRuleDescriptor, HolonError>
```

Resolution order:

1. Check direct `UsesKeyRule`.
2. If absent, walk `Extends` from most specific to most general.
3. Nearest inherited rule wins.
4. A found `NoneRule` means keyless, not error.
5. No `UsesKeyRule` anywhere fails with `NoEffectiveKeyRule`.

Introduce `KeyRuleDescriptor`, classified by reading the rule holon’s own type / `Extends`, not by adding key-rule names to `CoreHolonTypeName`.

`NoneRule` must remain a first-class resolved descriptor value. Do not encode it as `Ok(None)`, because `NoneRule` means a rule was found and says "keyless"; absence of any effective rule is an error.

Candidate helper:

```rust
KeyRuleDescriptor::is_keyless() -> Result<bool, HolonError>
```

#### 4.6 Errors

Reuse existing errors where they fit:

- `DescriptorDeclarationNotFound`
- `DuplicateInheritedDeclaration`
- `MissingRequiredRelationship`
- `MultipleRelatedHolons`

Specific new variants:

- `NoEffectiveKeyRule`
- `AmbiguousRelationshipTraversal { relationship, descriptor }`
- `UnsupportedStagedTraversal { relationship, descriptor }` for inverse discovery attempted against an unsaved type endpoint with no materialized `TargetOf` index

---

Affected areas:

- `shared_crates/holons_core/src/descriptors`
- `core_types` / `HolonError`
- descriptor unit tests

Expected downstream benefits:

- Query `Expand` can validate that a requested relationship is navigable outbound from a source type.
- Callers can enumerate the navigable relationships of a type, and the available relationships of a concrete holon in its current state.
- Declared navigation remains pre-commit safe via `InstanceRelationships`; inverse navigation is supported for committed descriptor graphs through `TargetOf`.
- Command, dance, property, relationship, operator, and key-rule discovery use one coherent descriptor API.

No change to:

- storage/DHT access
- Reference Layer boundaries
- wire contracts
- query execution
- key-string generation
- canonical core-schema import JSON

---

Add unit tests alongside existing descriptor tests.

Discovery tests:

- effective inherited properties, relationships, dances, commands, and operators
- ordering and duplicate semantics preserved

Validation tests:

- known names resolve
- unknown names fail clearly
- duplicate inherited declarations fail clearly

Relationship navigation tests:

- `allows_relationship` resolves a declared name outbound from the source type (`descriptor_direction = Declared`)
- `allows_relationship` resolves an inverse name outbound from the target type (`descriptor_direction = Inverse`)
- a declared name is **not** navigable from the target type (outbound-only semantics)
- unknown names fail with `DescriptorDeclarationNotFound` on a trustworthy (saved) index
- inherited declared relationships are discovered through `Extends`
- inherited/abstract target endpoint anchors are handled with equals-or-extends compatibility
- `TargetOf` candidates not licensed by the declared source’s effective `InstanceRelationships` are rejected
- missing `HasInverse` on a licensed `TargetOf` candidate fails clearly (schema defect, not generic not-found)
- ambiguous same-name matches fail with `AmbiguousRelationshipTraversal`
- unsaved endpoint with empty `TargetOf` index fails with `UnsupportedStagedTraversal`
- `navigable_relationships` enumerates declared plus inverse candidates
- `available_relationships` returns declared-only for transient and staged-uncommitted sources
- `StagedReference::is_committed()` matches any `StagedState::Committed(_)`

Key-rule tests:

- direct `UsesKeyRule`
- inherited `UsesKeyRule`
- nearest inherited rule wins
- `NoneRule` returns a resolved `KeyRuleDescriptor` whose classification is keyless
- missing effective rule fails with `NoEffectiveKeyRule`
- user-defined key-rule descriptor classification works without `CoreHolonTypeName`

Validation guidance:

- Run relevant shared-crate tests.
- If shared API behavior affects hApp-reachable code, run `npm run check -w map-happ`.

---

- `HolonDescriptor` and `ValueDescriptor` expose coherent discovery surfaces.
- Non-relationship validation helpers delegate to existing lookup methods.
- `QualifiedRelationship` exists with `descriptor` and `descriptor_direction`.
- `HolonDescriptor::allows_relationship(name)` validates outbound navigability and returns a `QualifiedRelationship`.
- `HolonDescriptor::navigable_relationships()` enumerates declared plus inverse navigable relationships.
- `available_relationships(source_ref)` filters the navigable set by source reference state per the availability table.
- Declared discovery uses forward `InstanceRelationships`; inverse discovery uses `TargetOf` only as a discovery/index path and validates candidates against the declared source’s effective `InstanceRelationships`.
- Endpoint compatibility is validated through `SourceType` / `TargetType` and equals-or-extends semantics.
- `relationship_navigation.rs` documents the schema/mutation/navigation/runtime-state/optimization separation of concerns; `related_holons` carries the future-validation TODO.
- `KeyRuleDescriptor` exists.
- `effective_key_rule()` returns a resolved `KeyRuleDescriptor`; `NoneRule` is represented as a keyless descriptor, not `Ok(None)`.
- Errors cover missing, ambiguous, malformed navigation/key-rule cases, and the unsaved-index guard.
- Unit tests cover discovery, validation, navigation, availability, and key-rule resolution.
- No query execution, key generation, converter migration, or canonical schema JSON edits are included.

---

<details>

- **Four-form traversal normalization (`normalize_relationship` + `TraversalDirection`): implemented, then dropped.** It normalized around the wrong axis — declared-vs-inverse surface-form equivalence — when the model is that navigation is always outbound and inverse descriptors are first-class relationships. Access-path normalization is deferred to query optimization.
- Source-anchored-only discovery: rejected because inverse relationships must be navigable from their own source type.
- Using `SourceOf` for all outbound discovery: rejected because declared outbound relationships are already licensed by forward `InstanceRelationships`, which is available pre-commit.
- A larger `NormalizedRelationshipTraversal` descriptor result: rejected because it mixes descriptor semantics with execution/planner read-channel metadata.
- Global schema scan: rejected because declared discovery uses `InstanceRelationships` and inverse discovery can use `TargetOf`.
- Loader/schema change to register inverse relationships on target type surfaces: rejected as unnecessary for committed descriptor graphs.
- `Result<Option<KeyRuleDescriptor>, HolonError>`: rejected because `NoneRule` is a found domain rule meaning keyless, while no effective rule is an error.
- `*Surface` struct types: rejected as extra ceremony over descriptor methods.
- Returning `HolonCollection`: rejected in favor of typed descriptor vectors.
- Adding key-rule names to `CoreHolonTypeName`: rejected to preserve user-defined key-rule support.
- Placing `available_relationships` on `HolonReference`: deferred; it lives as a free helper in `descriptors/` for now and may move later.

---

### 9. Risks or Concerns

- `TargetOf` relies on materialized inverse SmartLinks for committed data; inverse discovery over staged, uncommitted descriptor graphs is intentionally guarded (`UnsupportedStagedTraversal`).
- `InstanceRelationships` and `TargetOf` can disagree in malformed schema state; `InstanceRelationships` remains authoritative.
- Relationship base names are not globally unique; navigation helpers must detect ambiguity and filter by descriptor type/endpoint compatibility.
- New errors may cross IPC boundaries if later exposed through command/query wire layers.

---

### 10. Additional Context

Relevant files:

- `shared_crates/holons_core/src/descriptors/holon_descriptor.rs`
- `shared_crates/holons_core/src/descriptors/relationship_navigation.rs`
- `shared_crates/holons_core/src/descriptors/value_descriptor.rs`
- `shared_crates/holons_core/src/descriptors/inheritance.rs`
- `shared_crates/holons_core/src/descriptors/relationship_surface.rs`
- `shared_crates/holons_core/src/descriptors/inverse_resolution.rs`
- `shared_crates/holons_core/src/descriptors/declared_relationship_descriptor.rs`
- `shared_crates/holons_core/src/descriptors/inverse_relationship_descriptor.rs`
- `host/import_files/map-schema/core-schema/MAP Schema Types-map-core-schema-relationship-types.json`
- `host/import_files/map-schema/core-schema/MAP Schema Types-map-core-schema-keyrules-schema.json`
- `../map-dev-docs/docs/core/type-system/map-type-system.md`

</details>

### LINKED ISSUE #575: Reorganize outbound relationship helpers into effective and available surfaces
### 1. Summary (Required)

**What is the enhancement?**  
Refactor the outbound relationship helper surface introduced in PR 573 so the public terminology and ownership boundaries match the repository’s emerging model:

- descriptor-level, inheritance-resolved outbound relationship sets should use `effective` terminology and live on `HolonDescriptor`
- instance-level, runtime-state-filtered outbound relationship sets should use `available` terminology and live on `ReadableHolon` / `HolonReference`-reachable APIs
- the current mixed `relationship_navigation` / `relationship_surface` split should be reorganized into clearer descriptor-vs-instance responsibilities without changing the underlying navigation semantics validated in PR 573

The following diagram illustrates the terminology for classifying relationships.

Image

---

**Why is this needed?**  
PR 573 appears to have landed on the correct behavioral model for outbound relationship navigation, but it leaves the code organized around terms and seams that no longer fit that model cleanly.

Repository-grounded observations:

- `relationship_navigation.rs` currently mixes:
  - descriptor-level effective outbound relationship classification
  - instance-level availability filtering by source reference state
- `relationship_surface.rs` already exists and is still responsible for declared-only effective relationship lookup via `effective_relationship_declaration(...)`
- the codebase already uses `effective` elsewhere to mean “immediate + inherited,” including in `relationship_surface`
- the new `navigable_relationships()` term in PR 573 overlaps conceptually with “effective outbound relationships,” but introduces a second broad term for nearly the same inheritance-resolved descriptor-level set
- `available_relationships(source_ref)` is a narrower runtime-state overlay, but currently lives as a free helper in `descriptors/`, even though its behavior depends on concrete reference state and would fit more naturally on `ReadableHolon`

This creates two problems:

1. **Terminology drift**  
   The repo is converging on `effective` for inheritance-resolved sets, but PR 573 introduces `navigable` for the broad outbound descriptor-level union. That invites confusion between “descriptor-level effective set” and “runtime-available subset.”

2. **Responsibility overlap**  
   The declared portion of `relationship_navigation` overlaps with `relationship_surface`, while the availability logic belongs more naturally with reference-layer behavior than with descriptor utilities.

The result is logically correct code with avoidable conceptual duplication in a part of the system that is still settling its vocabulary and API shape.

---

Yes.

Prerequisites / context:

- Issue #536: Descriptor Effective Surface Discovery And Validation Helpers
- PR #573: `536 descriptor effective surface discovery and validation helpers`

This issue should be treated as a follow-on refactor/realignment issue after the navigation semantics established in PR 573 have been accepted as the intended behavior:

- outbound navigation is the governing axis
- inverse descriptors are first-class outbound relationships from their own `SourceType`
- effective/type-level relationship classification is distinct from runtime-state availability
- inverse availability depends on committed source state

This issue should not re-open the behavioral decision from PR 573 unless repository grounding during implementation reveals a contradiction.

---

**How would you solve it?**  
Reorganize the relationship helper surface around two public concepts only:

1. **Effective outbound relationships**  
   Descriptor-level, inheritance-resolved, state-agnostic.
2. **Available outbound relationships**  
   Instance-level, runtime-state-filtered subset of the effective outbound set.

#### 4.1 Rename and reorganize modules

Replace or evolve the current split:

- `relationship_surface.rs`
- `relationship_navigation.rs`

into a clearer separation such as:

- `effective_relationships.rs`
- `available_relationships.rs`

Exact filenames can follow repository convention, but the conceptual separation should be:

- descriptor-level effective outbound relationship computation in one place
- instance-level availability filtering in another place

#### 4.2 Move effective relationship APIs onto `HolonDescriptor`

Descriptor-level helpers should be methods on `HolonDescriptor`, because they assume a source holon type and do not depend on concrete reference state.

Target public methods:

```rust
HolonDescriptor::effective_declared_relationships()
HolonDescriptor::effective_inverse_relationships()
HolonDescriptor::effective_relationships()
```

Optional name-based lookup can remain or be renamed if useful, but should be derived from these core set methods.

Expected semantics:

- `effective_declared_relationships()`:
  outbound declared relationships licensed by effective `InstanceRelationships`
- `effective_inverse_relationships()`:
  outbound inverse relationships whose `SourceType` is this descriptor, discovered through `TargetOf` + `HasInverse`, with compatibility and licensing checks preserved
- `effective_relationships()`:
  union of declared + inverse effective outbound relationships

#### 4.3 Move available relationship APIs onto `ReadableHolon`

Instance-level availability should be exposed on the reference-facing surface, because it depends on runtime source state:

- `SmartReference` => declared + inverse available
- committed `StagedReference` => declared + inverse available
- uncommitted `StagedReference` => declared only
- `TransientReference` => declared only

Target public API shape:

```rust
ReadableHolon::available_relationships(&self)
```

or an equivalent reference-surface method.

This logic may continue to delegate internally to shared helpers, but the public responsibility should live with the instance/readable reference layer, not as a free helper in `descriptors/`.

#### 4.4 Consolidate declared-side duplication

The declared-side inheritance-resolved logic currently appears in both:

- `relationship_surface::effective_relationship_declaration(...)`
- the declared branch of `relationship_navigation`

This issue should consolidate that duplication around one authoritative descriptor-level computation for the effective declared outbound set, and have lookups/projected APIs build on top of it.

#### 4.5 Preserve the behavior from PR 573

This issue is about naming and structure, not changing the already-reviewed behavioral model. Preserve:

- outbound-only navigation semantics
- inverse descriptors as first-class outbound relationships
- `TargetOf` as discovery/index path for inverse relationships
- `InstanceRelationships` as authoritative licensing surface
- committed-state requirement for inverse availability
- staged/index guards and malformed-schema diagnostics, unless repository evidence shows they need correction

#### 4.6 Update docs and comments to the new vocabulary

Update file-level and method-level docs so they consistently use:

- `effective` for descriptor-level inheritance-resolved outbound sets
- `available` for runtime-state-filtered instance-level subsets

Avoid continuing to use `navigable` as the primary public term if the codebase-standard term is `effective`.

---

**What does this impact?**  
Affected areas likely include:

- `shared_crates/holons_core/src/descriptors/relationship_navigation.rs`
- `shared_crates/holons_core/src/descriptors/relationship_surface.rs`
- `shared_crates/holons_core/src/descriptors/holon_descriptor.rs`
- `shared_crates/holons_core/src/reference_layer/readable_holon.rs`
- `shared_crates/holons_core/src/reference_layer/*` where state-based filtering is delegated
- `shared_crates/holons_core/src/descriptors/mod.rs`
- re-exports in `shared_crates/holons_core/src/lib.rs`
- relationship helper tests

Impacts:

- descriptor API naming and discoverability
- reference-layer public API surface
- internal module organization
- documentation and terminology consistency
- downstream callers of PR 573 helper names, if already introduced in adjacent code

Non-goals:

- changing the outbound-navigation semantics adopted in PR 573
- changing mutation semantics for declared vs inverse relationships
- implementing future `InvalidRelationshipName` / `RelationshipUnavailableInState` validation in `related_holons`
- query planner access-path optimization

---

- Existing relationship-navigation tests from PR 573 should continue to pass after the rename/restructure.
- Descriptor-level tests should verify:
  - effective declared relationships
  - effective inverse relationships
  - effective outbound relationship union
  - duplicate inherited declaration behavior
  - licensed-vs-unlicensed inverse discovery
  - malformed `HasInverse` behavior
  - staged `TargetOf` guard behavior where still applicable
- Reference-level tests should verify:
  - available relationships on `SmartReference`
  - available relationships on committed `StagedReference`
  - declared-only availability on uncommitted `StagedReference`
  - declared-only availability on `TransientReference`
- Any renamed public API should be covered by compile-time call sites and unit tests to prevent drift between terminology and implementation.

Suggested validation:

- `npm run fmt:check`
- `npm run test:unit`
- `npm run check`
- targeted descriptor/reference-layer tests if faster while iterating

---

- Descriptor-level outbound relationship classification uses `effective` terminology consistently.
- Instance-level runtime-state-filtered outbound relationship access uses `available` terminology consistently.
- The public descriptor-level helper surface lives on `HolonDescriptor`.
- The public instance-level availability helper surface lives on `ReadableHolon` or an equivalent reference-layer receiver.
- The broad `navigable` terminology introduced by PR 573 is removed or reduced to internal implementation comments rather than remaining the primary public concept.
- The declared-side effective relationship logic is no longer duplicated across `relationship_surface` and navigation helpers.
- Module/file organization reflects descriptor-level vs instance-level responsibilities cleanly.
- Existing PR 573 navigation semantics remain intact unless explicit repository-grounded evidence requires adjustment.
- Docs/comments align with the repo’s established use of `effective`.
- Tests pass and cover both descriptor-level effective sets and instance-level available sets.

---

<details>

- Merge PR 573 as-is and defer cleanup indefinitely.  
  Rejected because this area is still solidifying its conceptual model, so terminology and ownership are part of maintainability, not just polish.

- Keep `relationship_navigation.rs` as the single home for both descriptor-level and instance-level helpers.  
  Rejected because it mixes type semantics with runtime-state filtering and overlaps conceptually with `relationship_surface`.

- Extract many private “collect_*” helpers up front.  
  Rejected as premature unless real reuse demands them. The preferred shape is to let the public semantic methods define the model first, then only factor private helpers where complexity forces it.

### 9. Risks or Concerns

**What could go wrong?**

- Renaming public helpers can create churn for nearby callers if they begin adopting PR 573 APIs before this refactor lands.
- Over-eager restructuring could accidentally change behavior while renaming; this work should preserve the reviewed semantics.
- Splitting modules without a crisp terminology block may just move confusion around rather than resolving it.

### 10. Additional Context

**Any supporting material?**

Repository-grounded context inspected:

- Issue #536 current body
- PR #573 description and discussion comments, especially:
  - updated relationship navigation semantics comment from the PR discussion
  - follow-up note that naming/structure are still unsettled
- Current code:
  - [shared_crates/holons_core/src/descriptors/relationship_navigation.rs](/Users/stevemelville/dev/map-proto/map-dev/map-holons/shared_crates/holons_core/src/descriptors/relationship_navigation.rs)
  - [shared_crates/holons_core/src/descriptors/relationship_surface.rs](/Users/stevemelville/dev/map-proto/map-dev/map-holons/shared_crates/holons_core/src/descriptors/relationship_surface.rs)
  - [shared_crates/holons_core/src/descriptors/holon_descriptor.rs](/Users/stevemelville/dev/map-proto/map-dev/map-holons/shared_crates/holons_core/src/descriptors/holon_descriptor.rs)
  - [shared_crates/holons_core/src/reference_layer/readable_holon.rs](/Users/stevemelville/dev/map-proto/map-dev/map-holons/shared_crates/holons_core/src/reference_layer/readable_holon.rs)
  - [shared_crates/holons_core/src/reference_layer/staged_reference.rs](/Users/stevemelville/dev/map-proto/map-dev/map-holons/shared_crates/holons_core/src/reference_layer/staged_reference.rs)

Assumptions:
- This follow-on issue is intended to supersede the public naming/structure introduced by PR 573 without re-opening its adopted outbound-navigation behavior.
- `docs/roadmap/desc-driven-impl-plan.md` and copied DevDocs track documents are not present in this checkout, so PR #573 discussion and Issue #536 were used as the current architecture-intent source for grounding.

</details>

owleyeview:
### Summary

Adds coherent, intention-revealing **type-level methods** on `HolonDescriptor` and `ValueDescriptor` that answer "what can I do with this type of holon?" — packaging existing discovery behavior behind validation entrypoints and adding two genuinely new behaviors: **relationship traversal normalization** (reducing all declared/inverse × inbound/outbound surface forms to one canonical declared relationship) and **effective key-rule resolution**. All work stays inside the descriptor layer in `shared_crates/holons_core/src/descriptors`.

Scope: **10 files, ~1.3k insertions.** No query execution, key-string generation, converter/importer migration, wire-contract, or canonical schema JSON changes.

### What changed

**Groundwork**
- Five new `HolonError` variants (+ `HolonErrorKind` discriminants + `ResponseStatusCode` mappings): `NoEffectiveKeyRule`, `UnsupportedKeyRule`, `IllegalRelationshipTraversal`, `AmbiguousRelationshipTraversal`, `UnsupportedStagedTraversal`.
- New `relationship_traversal` module: `TraversalDirection { Inbound, Outbound }` (navigation axis) and `QualifiedRelationship`, kept deliberately distinct from the existing `RelationshipDirection { Declared, Inverse }` (declared-vs-inverse classification).
- `equals_or_extends` helper in `inheritance.rs` for reference-identity endpoint compatibility.
- Re-added the `TargetOf` variant to `CoreRelationshipTypeName` (schema-authored as the inverse of the meta-level `TargetType`; previously removed as unused).

**Validation entrypoints** (thin delegators over existing lookups, inheriting not-found/duplicate semantics)
- `HolonDescriptor::allows_property`, `affords_command`, `affords_dance`
- `ValueDescriptor::affords_operator`

**Relationship validation + normalization** (`relationship_normalization` module)
- `HolonDescriptor::allows_relationship(name, direction) -> QualifiedRelationship` — resolves the as-requested descriptor and its classification.
- `HolonDescriptor::normalize_relationship(name, direction) -> DeclaredRelationshipDescriptor` — reduces any of the four equivalent forms `(A)-[R]->(B)`, `(A)<-[Rinv]-(B)`, `(B)-[Rinv]->(A)`, `(B)<-[R]-(A)` to the canonical declared relationship. Source-owned discovery uses forward `InstanceRelationships`; target-owned discovery uses the `TargetOf` index with a source-`InstanceRelationships` licensing cross-check and equals-or-extends endpoint compatibility.

**Effective key-rule resolution**
- New `KeyRuleDescriptor` wrapper with `is_keyless()` (classifies `NoneRule` via the rule's own `TypeName`/`Extends`, with no coupling to `CoreHolonTypeName`).
- `HolonDescriptor::effective_key_rule()` — walks the `Extends` chain self-first (nearest `UsesKeyRule` wins); `NoneRule` returns a real `KeyRuleDescriptor` meaning keyless, genuine absence returns `NoEffectiveKeyRule`.

### Design notes

- **Aligned with Issue #562's polarity:** `HasInverse` is the declared member of the pair (on `DeclaredRelationshipType`), `InverseOf` its inverse. Normalization always discovers the declared relationship and reaches the inverse via `has_inverse()`/`required_inverse()`, reusing the `inverse_resolution`/`effective_relationship_declaration` machinery rather than duplicating it.
- **Staged-traversal guard:** a staged/transient endpoint with an empty `TargetOf` index can't answer target-owned traversal, so it returns `UnsupportedStagedTraversal` rather than a misleading not-found. A saved endpoint's empty result is a true not-found. Source-owned (forward) traversal remains fully answerable pre-commit.
- **Error precedence** in normalization: Ambiguous → Illegal (wrong form) → MissingRequiredRelationship (definitional inverse absent) → UnsupportedStaged → NotFound.

### Testing

- Unit tests cover the full traversal matrix (all four forms, source- and target-owned, inherited endpoints), licensing rejection, ambiguity, illegal-form, missing-inverse, and both staged paths; plus key-rule direct/inherited/nearest-wins/keyless/missing/user-defined cases and the validation-entrypoint delegators.
- `cargo test -p holons_core descriptors::` — **218 passing.**
- `cargo fmt --check` — clean.
- `npm run check -w map-happ` — passes (confirms the shared crate stays WASM-safe).
- `npm run test` -- passes.

### Out of scope

Query execution, key-string generation, converter/importer migration, command/dance execution redesign, hand-editing canonical core-schema JSON, adding key-rule variants to `CoreHolonTypeName`, and transaction-local target-side discovery for staged descriptor graphs.


evomimic (changes requested):
## Review Findings

Exciting to see this close to delivery!

A few open issues related to error reporting.

1. `P1` Staged source-owned traversal can be misclassified as `UnsupportedStagedTraversal` instead of a normal not-found or successful source-owned result. In `shared_crates/holons_core/src/descriptors/relationship_normalization.rs`, the staged fallback is triggered whenever `TargetOf` is empty on an unsaved endpoint, even though source-owned traversal should remain answerable from `InstanceRelationships` alone. That conflicts with the issue contract for pre-commit-safe source-owned normalization and can reject valid staged workflows.

2. `P1` Missing-inverse handling is applied too broadly and can turn unrelated lookup misses into schema-defect errors. In `shared_crates/holons_core/src/descriptors/relationship_normalization.rs`, the code records a missing inverse whenever it encounters a candidate without `HasInverse`, without first proving the requested name matched that candidate. As a result, a typo or unrelated traversal request can surface `MissingRequiredRelationship` instead of `DescriptorDeclarationNotFound`, which breaks the intended error precedence.

3. `P2` `effective_key_rule()` does not validate that the resolved rule is actually a key-rule descriptor, so malformed schemas can pass through silently. In `shared_crates/holons_core/src/descriptors/holon_descriptor.rs` and `shared_crates/holons_core/src/descriptors/key_rule_descriptor.rs`, the returned wrapper only checks whether the rule is `NoneRule`-like. That leaves the new `UnsupportedKeyRule` path effectively unused and means invalid rule types can be treated as valid effective key rules.

PR 573 is not merge-ready yet.

owleyeview:
@evomimic 

Thanks for the feedback. I rechecked the three findings against the current implementation. The review surfaced one useful hardening gap, but a couple of details were missed.

For staged source-owned traversal, successful source-owned matches are not rejected by the staged `TargetOf` fallback. `resolve_unique` returns a unique declared match before the staged guard is considered, so source-owned traversal remains answerable from `InstanceRelationships` alone and stays pre-commit safe. The remaining edge is only a not-found/error-label ambiguity on unsaved endpoints when no source-owned match exists and no materialized `TargetOf` index is available.

For missing inverse handling, the current behavior is intentional for malformed schema state. If `HasInverse` is absent, there is no inverse name available to compare against the requested name, so inverse-direction traversal on that declaration surfaces the schema defect as `MissingRequiredRelationship` rather than treating it as a generic miss. Well-formed schemas still return normal not-found behavior for unknown names.

The key-rule validation point was valid. I made the minimal update there:
- `effective_key_rule()` now validates that the resolved `UsesKeyRule` target extends `KeyRuleType`.
- Invalid targets now return the existing `WrongDescriptorKind { expected: "KeyRuleType", .. }`.
- I removed the new `UnsupportedKeyRule` variant entirely, since this is descriptor-kind validation rather than an unsupported-rule policy.
- Key-rule tests now model concrete rules as extending `KeyRuleType`, and there is a new invalid-target test.
- I also added comments clarifying the staged `TargetOf` fallback and missing-inverse behavior.

Validation is passing, including `npm run test`.

evomimic:
# Updated Relationship navigation semantics

Upon reflection... I've realized `normalize_relationship` was normalizing around the wrong axis. I re-evaluated the situation and landed on the following actions.

### Drop `normalize_relationship` altogether.

It was normalizing around declared-vs-inverse status, but that is not the right axis for query navigation. If anything, normalization belongs later in query optimization, where the planner may choose an access path. That is out of scope for now.

### Update `available_relationships` Semantics

`available_relationships(source_ref)` answers:

Given a known source holon reference, what outbound relationships are actually available to navigate from that source right now?

The key rule is:

    Available = outbound from this holon type
    not
    Available = declared by this holon type

So the first filter is:

    source_type = type_of(source_ref)

    candidates =
      all RelationshipTypes
      where relationship.SourceType == source_type

Those candidates may include both:

- `DeclaredRelationshipType`
- `InverseRelationshipType`

Declared-vs-inverse status does not determine whether the relationship is conceptually navigable. The relationship is conceptually navigable when the source holon’s type is the relationship’s `SourceType`.

However, whether that relationship is truly navigable depends on upon the type and state of the source holon reference.  So the following additional logic should be applied.

After finding outbound candidates, `available_relationships(source_ref)` must filter by the state of the source reference:

    DeclaredRelationshipType:
      available

    InverseRelationshipType:
      available if source_ref is a SmartReference (i.e., referencing an already committed holon)
      available if source_ref is a StagedReference ONLY if it's `StagedState=Comitted`
      not available for TransientReference 

Rationale:

- Inverse relationships are populated on commit.
- Transient holons are never committed, so inverse relationships will not be populated.
- Unsaved staged holons have not yet been committed, so inverse relationships will not be populated.
- Saved holons or staged holons in a committed state, will have had  inverse relationships populated via their commit process

### `get_related_holons` validation (future)

When `get_related_holons(source_ref, relationship_name)` performs validation, it should distinguish two different errors (out of scope for this PR).

#### Invalid relationship name

Return `InvalidRelationshipName` when:

    relationship_name is not a valid outbound relationship for source_type

That means there is no relationship with:

    relationship.SourceType == type_of(source_ref)
    and relationship.name == relationship_name

This should take precedence.

#### Relationship unavailable in state

Return `RelationshipUnavailableInState` when:

    relationship_name is a valid outbound relationship for source_type
    but the relationship is an InverseRelationshipType
    and source_ref is Transient or unsaved Staged

This is not a schema error. The relationship exists and is valid for the source type, but it is not available from this particular source reference in its current state.

### Separation of concerns

This preserves a clean separation:

    Schema semantics:
      SourceType / TargetType / InverseOf

    Mutation semantics:
      only declared relationships are directly editable

    Navigation semantics:
      outbound relationships available from current source

    Runtime state semantics:
      inverse navigation requires a committed source state

    Optimization semantics:
      possible future access-path selection

### Provide Doc Comments

These semantics are subtle enough that we should use doc comments at both file and method level liberally to make clear the "why" and the "what" of the logic being implemented

owleyeview:
Reworked the relationship helpers per the updated navigation semantics:

- **Dropped `normalize_relationship` and `TraversalDirection`** — the four-form declared/inverse normalization was organized around the wrong axis. Navigation is always outbound from the source; inverse descriptors are first-class navigable relationships. Access-path selection (e.g. rewriting inverse navigation as a reverse scan of the declared relationship) is deferred to future query optimization.
- **`allows_relationship(name)`** (no direction parameter) validates that a name is navigable outbound from a type: declared via effective `InstanceRelationships`, or inverse via the `TargetOf` index + `HasInverse`, with endpoint compatibility and licensing checks preserved. `QualifiedRelationship` now carries just the as-matched descriptor and its declared/inverse classification.
- **New `HolonDescriptor::navigable_relationships()`** — type-level, state-agnostic enumeration of the outbound-navigable set (declared ∪ inverse).
- **New `available_relationships(source_ref)`** — filters the navigable set by source reference state: declared always available; inverse only for `SmartReference` or committed `StagedReference` (inverse SmartLinks materialize at commit). Uncommitted sources are declared-only by construction, so the `TargetOf` staged guard can never fire from this path. It lives as a free helper in `descriptors/relationship_navigation.rs` for now; it could reasonably move onto `HolonReference` later.
- **New `StagedReference::is_committed()`** predicate (matches any `StagedState::Committed(_)`).
- **Errors:** removed `IllegalRelationshipTraversal` (nothing constructs it without the direction axis) and dropped the `direction` field from `AmbiguousRelationshipTraversal`; `UnsupportedStagedTraversal` semantics unchanged.
- **Docs:** `relationship_navigation.rs` opens with the schema / mutation / navigation / runtime-state / optimization separation of concerns; `ReadableHolon::related_holons` carries a TODO for the future `InvalidRelationshipName` (takes precedence) / `RelationshipUnavailableInState` validation, out of scope for this PR.

Unit tests reworked accordingly (declared/inverse outbound resolution, outbound-only rejection of declared names from the target side, ambiguity, licensing, inherited discovery, missing-`HasInverse` defect, staged guard, enumeration, availability filtering). `npm run fmt:check`, `npm run test:unit`, `npm run check`, and `npm run test` all pass.

evomimic (changes requested):
## Review Findings

PR 573 appears to have landed on the correct behavioral model for outbound relationship navigation. The logic looks correct and I've confirmed all tests pass and npm start loads core and test schemas without any problem.

However, the code is organized around terms and seams that no longer fit that model cleanly.

Repository-grounded observations:

- `relationship_navigation.rs` currently mixes:
  - descriptor-level effective outbound relationship classification
  - instance-level availability filtering by source reference state
- `relationship_surface.rs` already exists and is still responsible for declared-only effective relationship lookup via `effective_relationship_declaration(...)`
- the codebase already uses `effective` elsewhere to mean “immediate + inherited,” including in `relationship_surface`
- the new `navigable_relationships()` term in PR 573 overlaps conceptually with “effective outbound relationships,” but introduces a second broad term for nearly the same inheritance-resolved descriptor-level set
- `available_relationships(source_ref)` is a narrower runtime-state overlay, but currently lives as a free helper in `descriptors/`, even though its behavior depends on concrete reference state and would fit more naturally on `ReadableHolon`

This creates two problems:

1. **Terminology drift**  
   The repo is converging on `effective` for inheritance-resolved sets, but PR 573 introduces `navigable` for the broad outbound descriptor-level union. That invites confusion between “descriptor-level effective set” and “runtime-available subset.”

2. **Responsibility overlap**  
   The declared portion of `relationship_navigation` overlaps with `relationship_surface`, while the availability logic belongs more naturally with reference-layer behavior than with descriptor utilities.

The result is logically correct code with avoidable conceptual duplication in a part of the system that is still settling its vocabulary and API shape.

### Proposal

I've defined Issue #575 to clean up the structure and terminology.  I suggest we implement this issue on the same branch (536-descriptor-effective-surface-discovery-and-validation-helpers) before merging PR #573 .

I've created the following diagram to clarify the terminology and serve as an aid to organizing the code.

Image

owleyeview:
Issue #575 has been implemented and is now part of this PR.

evomimic (approved):
Approve from me.

I re-reviewed the latest head and the follow-up `#575` restructuring has landed cleanly. The relationship helper surface now lines up with the intended model:

- descriptor-level outbound relationship classification uses `effective_*` terminology on `HolonDescriptor`
- instance-level runtime-state filtering uses `available_relationships()` on `ReadableHolon`
- the old mixed `relationship_surface` / navigation split has been cleaned up
- the naming and ownership concerns from the prior review appear resolved

Validation also looks solid on this head:

- CI checks are all green
- `build:happ` and `build:host` pass
- `npm test` passes
- `npm start` is nominal, including schema load

I don’t see any remaining blocking issues. This looks merge-ready to me.

Issue #444, opened 2026-04-09
Nursery-Level Duplicate Detection and Idempotent SmartLink Persistence


evomimic:
### 1. Summary

Introduce duplicate detection and idempotent behavior for relationship mutations:

- Enforce **duplicate prevention at mutation time** in the Nursery  
- Ensure **idempotent behavior** when the same relationship is added multiple times within a transaction  
- Apply **commit-time duplicate suppression** to handle narrow race conditions from concurrent updates  

This improves **fail-fast semantics**, reduces ambiguity, and ensures consistent graph structure under concurrent operations.

---

### 2. Problem Statement

The current system does not define clear behavior for duplicate relationships:

- Mutation APIs (e.g., `add_related_holon(...)`) may allow the same target to be added multiple times  
- There is no explicit idempotency guarantee when the same relationship is added repeatedly within a transaction  
- Commit logic does not account for duplicates introduced by concurrent unseen updates  

This leads to:

- Ambiguous mutation semantics  
- Potential duplicate `SmartLink`s in persisted graph state  
- Delayed or indirect detection of invalid operations  

From a developer and user experience perspective:

> Errors that can be detected locally should be surfaced immediately, not deferred to commit or left implicit.

---

### 3. Goals

- Provide **fail-fast feedback** during mutation  
- Ensure **deterministic and idempotent relationship operations**  
- Prevent duplicate relationships within a staged context  
- Handle unavoidable distributed race conditions safely at commit time  

---

#### 4.1 Nursery-Level Duplicate Detection

Update mutation APIs (e.g., `add_related_holon(...)`) to:

- Check for existing targets within the staged relationship collection  
- Reject duplicate additions  

This applies to:

- Newly created holons (`ForCreate`)  
- Staged clones of existing holons  

Because all relevant relationships are present in the staged context, this check is **authoritative at mutation time**.

---

#### 4.2 Idempotent Mutation Semantics

Within a single transaction:

- Adding the same relationship multiple times should behave **idempotently**  
- The resulting staged relationship set should contain at most one instance of a given target  

This allows callers to:

- compose operations safely  
- avoid defensive duplication checks in higher-level logic  

---

#### 4.3 Commit-Time Duplicate Suppression

For relationships ultimately anchored to a persisted holon:

- Before writing a `SmartLink`, check whether an identical link already exists  
- If it exists → treat as **idempotent no-op** (not an error)  

This handles the narrow race condition where:

- another transaction adds the same relationship  
[...]

PR #576, opened 2026-07-04
516 relationship persistence follow up 2 nursery level duplicate detection and idempotent smartlink persistence

### LINKED ISSUE #516: Relationship Persistence Follow-Up 2: Nursery-Level Duplicate Detection and Idempotent SmartLink Persistence
### 1. Summary (Required)

**What is the enhancement?**

Harden relationship mutation and relationship persistence by adding descriptor-aware duplicate handling, idempotent repeated-add behavior, commit-time SmartLink duplicate suppression, and relationship-pair uniqueness validation.

This is a narrowed update to Issue #516 after the landed relationship persistence work in #442, #536, and #562. Those issues established automatic inverse SmartLink materialization, `HasInverse` as the declared relationship-pair edge, descriptor-owned effective relationship discovery, and loader rejection of inverse-oriented instance input. This issue keeps the remaining hardening work: prevent invalid duplicate relationship occurrences, make repeated relationship operations predictable, avoid duplicate persisted SmartLinks, and close the inverse-side uniqueness validation gap.

---

**Why is this needed?**

Relationship mutation still permits duplicate relationship members in ordinary runtime APIs.

For example, a caller can stage the same relationship target twice:

```text
Book --AuthoredBy--> Alice
Book --AuthoredBy--> Alice
```

Current collection mutation appends members directly. Duplicate keys may produce a warning, but the duplicate member is still accepted. Descriptor metadata such as `allows_duplicates()` exists, but ordinary mutation does not use it.

Commit-time SmartLink persistence is also not idempotent. Commit Pass 2 materializes forward and inverse SmartLinks, but `save_smartlink()` writes unconditionally. If the same relationship is committed twice, retried after partial persistence, or concurrently introduced by another transaction, equivalent SmartLinks can be written again.

There is also a remaining relationship-pair validation gap: loader validation requires each declared relationship descriptor to author exactly one `HasInverse`, but it does not enforce the inverse-side uniqueness invariant. Two declared relationship descriptors can point at the same inverse descriptor, which can leave the inverse descriptor with ambiguous materialized `InverseOf` links.

---

This issue depends on landed relationship persistence and descriptor work:

- #442: automatic forward/inverse SmartLink materialization during relationship persistence.
- #562: `HasInverse` is now the declared relationship-pair edge; authored `InverseOf` is rejected by loader import validation.
- #536 / current effective-relationship branch: descriptor-owned effective relationship discovery, declared/inverse classification, and relationship metadata accessors, including `allows_duplicates()`.

No schema orientation change is expected. This issue should consume the current #562 polarity:

```text
DeclaredRelationshipType --HasInverse--> InverseRelationshipType
InverseRelationshipType --InverseOf--> DeclaredRelationshipType
```

Current relationship discovery vocabulary to use:

- `HolonDescriptor::allows_relationship(...)`
- `HolonDescriptor::effective_declared_relationships()`
- `HolonDescriptor::effective_relationships()`
- `ReadableHolon::available_relationships()`
- `QualifiedRelationship.descriptor_direction`
- `RelationshipDirection::{Declared, Inverse}`

Mutation semantics remain declared-side-only: inverse descriptors are valid outbound navigation relationships after commit, but they are not ordinary editable relationship inputs.

---

**How would you solve it?**

Implement duplicate and idempotency semantics in three places: import validation, ordinary relationship mutation, and SmartLink persistence.

#### 4.1 Enforce inverse-side uniqueness for relationship-pair metadata

Extend loader/import validation so every inverse relationship descriptor is targeted by at most one declared relationship descriptor via `HasInverse`.

Current validation already rejects authored `InverseOf` and requires each declared relationship descriptor to author exactly one `HasInverse`. Add the converse check:

- collect all `HasInverse` targets authored by declared relationship descriptors
- count targets across the broadest import/load-set validation boundary available
- reject any inverse descriptor referenced by more than one declared descriptor
- keep the existing target-kind check requiring the target to extend `InverseRelationshipType`

If parser validation is file-local, add load-set-level validation so duplicates split across bundles are not missed.

#### 4.2 Add descriptor-aware duplicate handling to ordinary relationship mutation

Before accepting `AddRelatedHolons` through ordinary caller APIs, resolve the requested
relationship's effective declared descriptor and read `allows_duplicates()`.

Use the instance-anchored declared lookup for mutation validation:

- `effective_relationship_declaration(source_holon, name)` is the established declared-side
  resolution helper (it is the commit Pass 2 contract). It resolves through
  `effective_descriptor_lineage`, which includes a descriptor holon's own `Extends` lineage in
  addition to its `DescribedBy` type. This matters for mutation: descriptor holons staging
  relationships such as `SourceType` during schema construction resolve via own-lineage, so using
  this helper directly narrows the bootstrap exceptions this issue would otherwise need.
- The returned `RelationshipDescriptor` carries `allows_duplicates()` directly.
- The `HolonDescriptor` methods (`effective_declared_relationships()`, `allows_relationship(...)`)
  are anchored on the `DescribedBy`-resolved type only; they are the right surface for
  type-level queries, but mutation validation should anchor on the source instance.

Consolidation note: `StagedReference::classify_relationship_mutation` currently resolves its
descriptor via `HolonDescriptor::get_relationship_by_name`, a second, `DescribedBy`-anchored
declared lookup that overlaps with `effective_relationship_declaration`. This issue touches
exactly that call site; consolidate onto one instance-anchored resolution path rather than
leaving two parallel declared lookups.

Mutation semantics remain declared-side-only. The current effective-relationship model
distinguishes navigation from mutation:

- declared relationships are editable mutation inputs
- inverse relationships are effective outbound navigation relationships only after commit
- inverse relationships are materialized as SmartLinks by commit and must not be staged directly
  by callers

For relationships where duplicates are disallowed:

- detect duplicate targets already present in the staged relationship collection (for `ForUpdate*` sources this is best-effort. Defer to §4.6/§4.4 for authority against the persisted graph.)
- treat exact repeated adds as idempotent no-ops
- fail clearly for ambiguous or non-identical duplicate attempts that resolve to the same
  relationship target identity
- do not append duplicate members to `HolonCollection`

For relationships where duplicates are allowed:

- preserve repeated-target behavior
- ensure tests cover an explicitly duplicate-allowed relationship

Descriptor resolution failure or unreadable duplicate metadata should fail loudly. Do not
silently fall back to permissive behavior. Note that declared-side resolution applies the eager
`DuplicateInheritedDeclaration` policy: a duplicate declared base name anywhere in the effective
lineage is a local schema defect and will now surface during mutation validation as well. This is
intended behavior, not a fallback case.

Bootstrap/import internals may keep narrowly scoped exceptions where descriptor graph
construction requires by-name staging before the full descriptor surface exists. Ordinary caller
mutation should not rely on those exceptions.

#### 4.3 Reject inverse relationship names in ordinary mutation

Loader/import paths already reject inverse-oriented instance input. Add the same
declared-side-only guard to ordinary relationship mutation APIs.

Resolve declared-first; classify inverse only on a miss:

- attempt declared resolution via `effective_relationship_declaration(source_holon, name)` — on
  success, the name is a valid mutation input and duplicate policy (4.2) applies
- only when declared resolution misses, optionally consult the inverse side (e.g.
  `HolonDescriptor::allows_relationship(...)` returning
  `RelationshipDirection::Inverse`) to distinguish "this name is an inverse relationship — stage
  the declared relationship from the declared source endpoint" from a plain invalid-name error

Do not classify via `allows_relationship(...)` as the first step. Its inverse arm consults the
`TargetOf` index, and on a source whose descriptor endpoint is staged-uncommitted it can surface
`UnsupportedStagedTraversal` instead of a clean declared/inverse verdict. Declared-first
resolution keeps the mutation path off the materialized-index guard entirely; the inverse
classification is a diagnostic refinement, not the gate.

`ReadableHolon::available_relationships()` is useful for state-filtered read/navigation
availability, but ordinary mutation should still reject inverse descriptors rather than treating
availability as editability.

If a caller attempts to stage an inverse relationship name as ordinary mutation input:

- reject it with a clear relationship error
- do not write an inverse-oriented staged collection
- direct callers to stage the declared relationship from the declared source endpoint

Do not silently normalize an inverse mutation request into the declared orientation. The current
policy is that inverse relationships are derived by commit, not caller-authored.

#### 4.4 Suppress equivalent SmartLinks at commit time

Before writing a SmartLink, check whether an equivalent persisted SmartLink is already visible from the same source.

An equivalent SmartLink should include:

- source `LocalId`
- target `HolonId` / local target id
- relationship name
- encoded smart property values used in the SmartLink tag

If an equivalent SmartLink already exists:

- do not create another link
- treat the operation as idempotent success
- apply the same behavior to forward and inverse SmartLinks

This should be implemented close to SmartLink persistence, such as by adding an idempotent save helper in `smartlink_adapter.rs`, so all commit callers share the same equivalence behavior.

#### 4.5 Keep forward and inverse persistence consistent

Commit Pass 2 already materializes forward and inverse SmartLinks from the same resolved relationship occurrence. Preserve that model.

Duplicate suppression must apply uniformly to:

- declared forward SmartLinks
- generated inverse SmartLinks
- relationship-pair metadata links such as materialized `HasInverse` / `InverseOf`

#### 4.6 Interaction with staged update states

Duplicate handling has different authority depending on the source holon's staged state, because
mutation-time detection can only see the staged relationship collection, not relationships already
persisted on an update target.

| Staged state | Duplicate authority | Rationale |
| --- | --- | --- |
| `ForCreate` | Mutation-time (authoritative) | All of the new holon's relationships are present in the staged context, so a mutation-time check against the staged collection is complete. |
| `ForUpdateGraphOnly` | Commit-time (authoritative) | Relationships attach to an existing persisted holon whose prior relationships are not in the staged collection. A mutation-time check is best-effort only; equivalent-SmartLink suppression at commit is what actually prevents a duplicate against the persisted graph. |
| `ForUpdateNewVersion` | Commit-time (new-version semantics) | A new version commits its own link set; suppression still applies uniformly per §4.4/§4.5. |

Implementation notes:

- Do not claim mutation-time detection is authoritative for update states. For any `ForUpdate*`
  source, the staged-collection check is a best-effort fast path; §4.4 commit-time suppression is
  the authoritative guarantee.
- `StagedReference::classify_relationship_mutation` already returns early for `ForCreate` and only
  runs descriptor classification for the update states, so the update states are exactly where the
  descriptor duplicate-policy path (4.2) is live. Keep the state branch explicit rather than
  assuming a single uniform mutation-time rule across all states.
- `Abandoned` and already-`Committed` staged references are not ordinary mutation targets and need
  no additional duplicate handling here.

---

**What does this impact?**

Primary implementation areas:

| Component | File(s) | Impact |
| --- | --- | --- |
| Loader/import validation | `host/crates/holons_loader_client/src/parser.rs`, loader load-set validation if needed | Enforce inverse-side uniqueness for `HasInverse` targets |
| Ordinary relationship mutation | `shared_crates/holons_core/src/reference_layer/staged_reference.rs`, `transient_reference.rs`, `holon_reference.rs` | Resolve descriptor state-aware duplicate policy (see §4.6) and reject inverse-oriented caller input |
| Relationship collections | `shared_crates/holons_core/src/core_shared_objects/holon_collection.rs`, `staged_relationship.rs` | Prevent invalid duplicate staged members without moving descriptor logic into low-level storage |
| Descriptor accessors | `shared_crates/holons_core/src/descriptors/*` | Consume existing `allows_duplicates()`, `required_inverse()`, and normalization helpers |
| Descriptor/effective relationship APIs | `shared_crates/holons_core/src/descriptors/effective_relationships.rs` (`effective_relationship_declaration`), `holon_descriptor.rs`, `reference_layer/available_relationships.rs` | Use instance-anchored `effective_relationship_declaration` for mutation validation; consolidate the overlapping `get_relationship_by_name` lookup in `classify_relationship_mutation` |
| Commit Pass 2 | `happ/crates/holons_guest/src/guest_shared_objects/commit_functions.rs` | Call idempotent SmartLink persistence for forward and inverse links |
| SmartLink persistence | `happ/crates/holons_guest/src/guest_shared_objects/smartlink_adapter.rs` | Add equivalent-link lookup/suppression |
| Tests | `shared_crates/holons_core`, `host/crates/holons_loader_client`, `happ`, `tests/sweetests` | Add duplicate/idempotency coverage |

Out of scope:

- Changing the #562 relationship-pair polarity.
- Reintroducing loader normalization of inverse-oriented instance input.
- Inverse deletion semantics.
- Cleanup or migration of already-duplicated persisted SmartLinks.
- Full cardinality enforcement beyond duplicate policy.
- Multi-space inverse propagation and deferred repair for external targets.
- Strong global race prevention beyond best-effort persisted-link suppression.

---

Add tests for:

- adding the same target twice through ordinary `AddRelatedHolons` when duplicates are disallowed
- exact repeated add returning success without adding a second collection member
- duplicate target rejection for non-identical references that resolve to the same target identity, if applicable
- duplicate targets being allowed when `allows_duplicates() == true`
- descriptor lookup failure during duplicate-policy resolution
- unreadable or missing `AllowsDuplicates` metadata producing an explicit error
- ordinary caller attempt to add a relationship name that resolves to `RelationshipDirection::Inverse`
- ordinary caller mutation accepting a relationship name that resolves to `RelationshipDirection::Declared`
- loader/import rejection when two declared descriptors target the same inverse descriptor via `HasInverse`
- cross-bundle inverse-side uniqueness if validation currently spans multiple files
- commit suppressing an already-persisted equivalent forward SmartLink
- commit suppressing an already-persisted equivalent inverse SmartLink
- existing forward/inverse SmartLink materialization still passing
- eager `DuplicateInheritedDeclaration` surfacing through mutation validation when the source's effective lineage contains a duplicate declared base name
- inverse relationship name rejection when the source's descriptor endpoint is staged-uncommitted, pinning that `UnsupportedStagedTraversal` does not leak into the mutation path
- - adding a target to a `ForUpdateGraphOnly` source that duplicates a relationship already persisted on the update target — mutation-time check does not see it, and commit suppresses the duplicate SmartLink
- adding a duplicate within the staged collection of a `ForUpdate*` source is handled by the same mutation-time policy as `ForCreate`

Suggested validation:

- focused Rust unit tests for descriptor/mutation behavior
- loader client parser tests for inverse-side uniqueness
- hApp/WASM check for shared-crate changes: `npm run check -w map-happ`
- broader regression as appropriate: `npm run check`, `npm run test:unit`, and relevant sweettests

---

- [ ] Loader/import validation rejects an inverse relationship descriptor targeted by more than one declared descriptor via `HasInverse`.
- [ ] Ordinary relationship mutation resolves descriptor duplicate policy before accepting a relationship add.
- [ ] Relationships with `allows_duplicates() == false` do not accumulate duplicate staged members.
- [ ] Exact repeated adds for duplicate-disallowed relationships are explicit, tested idempotent no-ops.
- [ ] Invalid or ambiguous duplicate attempts fail with clear diagnostics.
- [ ] Relationships with `allows_duplicates() == true` still permit repeated targets.
- [ ] Descriptor metadata failures produce explicit errors rather than permissive fallback behavior.
- [ ] Ordinary caller APIs reject inverse relationship names as mutation input.
- [ ] Commit-time SmartLink persistence suppresses equivalent already-persisted forward SmartLinks.
- [ ] Commit-time SmartLink persistence suppresses equivalent already-persisted inverse SmartLinks.
- [ ] Forward and inverse SmartLink idempotency use the same equivalence model.
- [ ] Existing relationship persistence, descriptor, loader, and sweettest coverage passes without regression.
- [ ] New tests cover mutation-time duplicates, inverse-side uniqueness validation, and commit-time duplicate suppression.
- [ ] Duplicate handling authority is correct per staged state: mutation-time for `ForCreate`, commit-time suppression authoritative for `ForUpdateGraphOnly` against already-persisted relationships.

---

<details>

One alternative is to handle duplicates only at commit time. That is insufficient because staged transaction state remains ambiguous and developers get delayed feedback.

Another alternative is to reject every repeated add. That is stricter but makes command composition and retry-like flows harder. Exact repeated adds should be idempotent no-ops when descriptor policy disallows duplicates.

Another alternative is to rely on Holochain link validation to reject duplicates. Current SmartLink validation is intentionally minimal and does not enforce relationship equivalence, so runtime/persistence logic must do the suppression.

---

### 9. Risks or Concerns

**What could go wrong?**

- Target identity normalization must work across staged and saved references without serializing bound runtime handles.
- `ForCreate` mutation validation may need carefully scoped bootstrap exceptions so schema loading is not broken.
- SmartLink equivalence must include enough tag metadata to avoid suppressing distinct relationship occurrences.
- Preflight persisted-link lookup is not a complete global concurrency guarantee in an eventually consistent DHT.
- Existing tests or fixtures may rely on permissive duplicate append behavior and need updates.
- Relationship descriptors whose schema explicitly allows duplicates must not be accidentally constrained.

---

### 10. Additional Context

**Any supporting material?**

Repository-grounded findings:

- `HolonCollection::add_references` appends members and only warns for duplicate keys.
- `StagedReference::classify_relationship_mutation` currently reads `IsDefinitional`, not `AllowsDuplicates`.
- `ForCreate` relationship mutation currently skips descriptor classification.
- Effective relationship discovery now distinguishes declared and inverse descriptors through `QualifiedRelationship.descriptor_direction`.
- `ReadableHolon::available_relationships()` filters declared/inverse availability by source reference state, but ordinary mutation still dispatches raw `AddRelatedHolons`.
- `save_smartlink()` calls `create_link()` unconditionally.
- `validate_create_smartlink_helper()` currently accepts SmartLink creates without duplicate validation.
- Loader/import validation currently rejects authored `InverseOf` and requires one `HasInverse` per declared descriptor, but does not enforce inverse-side uniqueness.

Relevant design context:

- `map-dev-docs/docs/core/holons-shared-objects-layer-design-spec.md`
- `map-dev-docs/docs/core/descriptors/descriptors-design-spec.md`
- `map-dev-docs/docs/core/descriptors/relationship-constraints-design-spec.md`
- `map-dev-docs/docs/core/validation/validation-arch.md`

</details>

owleyeview:
Closes #516 (and #444).

Hardens relationship semantics at three enforcement points: import validation, ordinary
staged mutation, and commit-time SmartLink persistence. Follows the landed #442/#536/#562
model: mutation stays declared-side-only, inverse links are derived at commit, and
`effective_relationship_declaration` remains the single declared-resolution contract.

## 1. Loader: inverse-side `HasInverse` uniqueness

`host/crates/holons_loader_client/src/parser.rs`

- Each inverse relationship descriptor may be targeted by **at most one** declared
  descriptor via `HasInverse`, enforced both within a file and **across the whole load
  set** (a `HasInverseTargetUniquenessTracker` held across the per-file loop).
- A violation surfaces as a `SchemaValidationFailure` import issue attributed to the file
  containing the second reference, naming both declared descriptors and the shared target.
- The existing checks (exactly one `HasInverse` per declared descriptor, target must extend
  `InverseRelationshipType`, authored `InverseOf` rejected) are unchanged.

## 2. Ordinary mutation: descriptor-aware duplicate policy (staged references)

`shared_crates/holons_core/src/reference_layer/staged_reference.rs`

- `classify_relationship_mutation` is replaced by a consolidated
  `relationship_mutation_policy` resolver anchored on the **instance** via
  `effective_relationship_declaration` (the commit Pass 2 contract), eliminating the
  parallel `DescribedBy`-anchored `get_relationship_by_name` lookup.
- On successful resolution the policy is **Enforced**: `AllowsDuplicates` is read
  (missing/non-bool metadata fails loudly — `EmptyField` / `UnexpectedValueType`), and
  update states additionally read `IsDefinitional` for state escalation.
- **Duplicate handling** when duplicates are disallowed: exact repeats (same reference
  identity) are dropped as idempotent no-ops — a pure no-op add does not escalate staged
  state; a same-key entry with a different identity fails with a clear `DuplicateError`;
  `allows_duplicates == true` preserves repeated targets exactly as before.
- **Inverse-name rejection**: resolution is declared-first; only on a miss is the name
  classified via `allows_relationship`, and an inverse name gets a clear error directing
  the caller to the declared source endpoint. Refinement errors (notably
  `UnsupportedStagedTraversal` from the `TargetOf` index guard on staged-uncommitted
  endpoints) never leak — the original miss is returned instead.
- `DuplicateInheritedDeclaration` propagates through mutation validation (eager local
  schema-defect surfacing, per the issue).
- **Removal** uses its own classifier: `ForCreate` sources skip descriptor classification
  entirely; update states resolve the declaration for `IsDefinitional` only (see decision
  notes below).

## 3. Commit-time SmartLink duplicate suppression

`happ/crates/holons_guest/src/guest_shared_objects/smartlink_adapter.rs`

- `save_smartlink` is now idempotent: it queries existing links with the **full encoded tag
  as `tag_prefix`** (narrowest possible filter), then compares **exact tag bytes** (guarding
  against a longer tag extending ours as a prefix) **and** target action hash. An equivalent
  link → idempotent success with a debug log; otherwise `create_link` as before.
- Equivalence = source `LocalId` + target hash + byte-identical tag, which covers
  relationship name, reference-type marker (incl. external space id), and smart property
  values. Tags are deterministic because `PropertyMap` is a `BTreeMap` (pinned by test).
- Both commit call sites flow through this single choke point, so forward links, generated
  inverse links, and `HasInverse`/`InverseOf` metadata links share one equivalence model.
- `commit_functions.rs` is unchanged; DHT link validation remains a no-op by design
  (sibling links are not cheaply visible to validation).

## Decisions and scope notes

- **Staged-only enforcement.** Duplicate policy and inverse-name rejection apply to
  `StagedReference` mutation only. `TransientReference` mutation is deliberately unchanged
  (transients never commit, so there is no SmartLink risk). This deviates from the issue's
  scope table, which also listed `transient_reference.rs` / `holon_reference.rs`.
- **Schema-gated enforcement.** An undescribed `ForCreate` source keeps legacy append
  behavior (this is what keeps undescribed fixtures and bootstrap staging working); a
  *described* source gets full validation with no permissive fallback.
- **Loader bootstrap exception.** Guest loader Pass 2 uses a documented
  `add_related_holons_ungoverned` escape hatch on `StagedReference`. Core-schema self-load
  is inherently circular: Pass 2a describes every holon before any `InstanceRelationships`
  edges exist, so ordinary policy resolution would fail on every described `ForCreate`
  source mid-load (the seed edge `(MetaHolonType)-[InstanceRelationships]->` is
  self-licensing). Write-access checks still apply on this path; ordinary callers do not
  use it.
- **`ForCreate` removal policy.** Removal on a `ForCreate` source does no descriptor work.
  Routing removal through full policy resolution created an un-removable staged member
  (add while undescribed → describe → remove failed with `DescriptorDeclarationNotFound`).
  Duplicate policy protects additions; removing an existing staged member needs no
  descriptor license. Update-state removal still resolves the declaration and errors on
  unknown names, preserving pre-#516 behavior.
- **Authority matrix (§4.6).** Mutation-time detection is authoritative for `ForCreate`
  (all relationships are staged) and best-effort for `ForUpdate*`; commit-time suppression
  is the authoritative guard against the persisted graph. Documented on
  `filter_duplicate_disallowed_entries` and pinned by tests on both sides.
- **Out of scope, per issue:** no cleanup of already-duplicated persisted links, no
  cardinality enforcement, no cross-space inverse propagation, no strong concurrency
  guarantees beyond best-effort suppression in an eventually consistent DHT.

## Testing

- **Unit — `holons_core` (19 staged_reference tests):** idempotent repeat (update-state and
  described-`ForCreate` variants), same-key/different-identity rejection, duplicates-allowed
  preservation, missing/invalid `AllowsDuplicates`/`IsDefinitional`, described-`ForCreate`
  unknown-name error, undescribed-`ForCreate` legacy append (schema gate pinned), inverse-name
  rejection on staged-uncommitted endpoints (pins no `UnsupportedStagedTraversal` leak),
  `DuplicateInheritedDeclaration` through mutation, `ForCreate` remove after describing,
  graph-only/new-version escalation, ungoverned loader path.
- **Unit — loader client (7 parser tests):** shared-inverse-target rejection (same file and
  cross-file), canonical core-schema imports validated jointly across all files.
- **Unit — smartlink adapter:** tag encode/decode round-trip, byte-determinism across
  insertion order, prefix-extension non-match (why exact-byte comparison is required).
- **Sweettest:** `stage_new_version_fixture` gains a replay leg — a second graph-only commit
  re-persists the cloned `Properties` collection against the same source, and the anchoring
  verifier now asserts **exactly one** forward and one inverse link for the replayed edge
  (plus the newly appended member's links, proving the commit wrote the touched collection
  rather than skipping it). Existing regression gates pass, including the core-schema load
  metrics (380 holons / 1778 links) and Book/Person forward+inverse materialization.
- **Commands:** `npm run check`, `npm run test:unit`, `npm run fmt:check`, and
  `npm run sweetest` (Nix) all green; `npm run check -w map-happ` proves WASM safety of the
  shared-crate changes.

### DIFF
--- a/happ/crates/holons_guest/src/guest_shared_objects/smartlink_adapter.rs
+++ b/happ/crates/holons_guest/src/guest_shared_objects/smartlink_adapter.rs
@@ -172,6 +172,12 @@ fn get_smartlink_from_link(
     Ok(smartlink)
 }
 
+/// Persists a SmartLink, suppressing the write when an equivalent link already exists.
+///
+/// Equivalence is source `LocalId` + target action hash + byte-identical encoded tag,
+/// which covers relationship name, reference-type marker (including external space id),
+/// and smart property values. `PropertyMap` is a `BTreeMap`, so equivalent inputs encode
+/// to byte-identical tags, making repeated commit passes idempotent at this choke point.
 pub fn save_smartlink(input: SmartLink) -> Result<(), HolonError> {
     // TODO: populate from property_map Null-separated property values (serialized into a String) for each of the properties listed in the access path
 
@@ -180,15 +186,44 @@ pub fn save_smartlink(input: SmartLink) -> Result<(), HolonError> {
         input.to_address.clone(),
         input.smart_property_values,
     )?;
-    create_link(
-        try_action_hash_from_local_id(&input.from_address)?,
-        try_action_hash_from_local_id(&input.to_address.local_id())?,
-        LinkTypes::SmartLink,
-        link_tag,
-    )
-    .map_err(|e| holon_error_from_wasm_error(e))?;
+    let source_hash = try_action_hash_from_local_id(&input.from_address)?;
+    let target_hash = try_action_hash_from_local_id(&input.to_address.local_id())?;
+
+    if equivalent_smartlink_exists(&source_hash, &target_hash, &link_tag)? {
+        debug!(
+            "SmartLink already persisted for relationship {:?} from {:?}; skipping duplicate write",
+            input.relationship_name.0 .0, input.from_address
+        );
+        return Ok(());
+    }
+
+    create_link(source_hash, target_hash, LinkTypes::SmartLink, link_tag)
+        .map_err(|e| holon_error_from_wasm_error(e))?;
 
     Ok(())
+}
+
+/// Returns true when a live SmartLink from `source_hash` to `target_hash` carrying exactly
+/// `link_tag` already exists.
+///
+/// The full encoded tag serves as the narrowest possible `tag_prefix` filter; exact tag
+/// bytes are then compared because a different tag may extend this one as a prefix (for
+/// example, the same relationship and target with additional smart property values).
+fn equivalent_smartlink_exists(
+    source_hash: &ActionHash,
+    target_hash: &ActionHash,
+    link_tag: &LinkTag,
+) -> Result<bool, HolonError> {
+    let links_query = LinkQuery::try_new(source_hash.clone(), LinkTypes::SmartLink)
+        .map_err(|e| holon_error_from_wasm_error(e))?
+        .tag_prefix(link_tag.clone());
+
+    let links = get_links(links_query, GetStrategy::default())
+        .map_err(|e| holon_error_from_wasm_error(e))?;
+
+    Ok(links.into_iter().any(|link| {
+        link.tag == *link_tag && link.target.into_action_hash().as_ref() == Some(target_hash)
+    }))
 }
 
 // HELPER FUNCTIONS //
@@ -482,5 +517,73 @@ mod tests {
         assert_eq!(Some(space_id), decoded_link_tag_object.proxy_id);
         assert!(decoded_link_tag_object.smart_property_values.is_some());
         assert_eq!(Some(property_values), decoded_link_tag_object.smart_property_values);
+    }
+
+    /// Pins the equivalence-model assumption behind duplicate suppression in
+    /// `save_smartlink`: equivalent inputs must encode to byte-identical tags
+    /// regardless of property insertion order (`PropertyMap` is a `BTreeMap`).
+    #[test]
+    fn test_encode_link_tag_is_deterministic_for_equivalent_inputs() {
+        let local_id =
+            LocalId("uhCkkLQ8hxxrt27W8TtkpcX1XAqbUyfD5_Rv5Us0X_-YeCT6RtMxU".as_bytes().to_vec());
+        let holon_id = HolonId::Local(local_id);
+        let relationship_name = RelationshipName(MapString("ex_relationship_name".to_string()));
+
+        let mut property_values_ascending: PropertyMap = BTreeMap::new();
+        property_values_ascending.insert(
+            PropertyName(MapString("ex_name_1".to_string())),
+            BaseValue::StringValue(MapString("ex_value_1".to_string())),
+        );
+        property_values_ascending.insert(
+            PropertyName(MapString("ex_name_2".to_string())),
+            BaseValue::StringValue(MapString("ex_value_2".to_string())),
+        );
+
+        let mut property_values_descending: PropertyMap = BTreeMap::new();
+        property_values_descending.insert(
+            PropertyName(MapString("ex_name_2".to_string())),
+            BaseValue::StringValue(MapString("ex_value_2".to_string())),
+        );
+        property_values_descending.insert(
+            PropertyName(MapString("ex_name_1".to_string())),
+            BaseValue::StringValue(MapString("ex_value_1".to_string())),
+        );
+
+        let tag_from_ascending =
+            encode_link_tag(&relationship_name, holon_id.clone(), Some(property_values_ascending))
+                .unwrap();
+        let tag_from_descending =
+            encode_link_tag(&relationship_name, holon_id, Some(property_values_descending))
+                .unwrap();
+
+        assert_eq!(tag_from_ascending.into_inner(), tag_from_descending.into_inner());
+    }
+
+    /// Pins why `equivalent_smartlink_exists` compares exact tag bytes after the
+    /// full-tag `tag_prefix` query: a tag carrying additional smart property values
+    /// extends the property-free tag as a prefix, so it would be returned by the
+    /// prefix query but must not count as an equivalent link.
+    #[test]
+    fn test_extended_link_tag_matches_prefix_but_not_exact_bytes() {
+        let local_id =
+            LocalId("uhCkkLQ8hxxrt27W8TtkpcX1XAqbUyfD5_Rv5Us0X_-YeCT6RtMxU".as_bytes().to_vec());
+        let holon_id = HolonId::Local(local_id);
+        let relationship_name = RelationshipName(MapString("ex_relationship_name".to_string()));
+
+        let base_tag = encode_link_tag(&relationship_name, holon_id.clone(), None).unwrap();
+
+        let mut property_values: PropertyMap = BTreeMap::new();
+        property_values.insert(
+            PropertyName(MapString("Key".to_string())),
+            BaseValue::StringValue(MapString("ex_key".to_string())),
+        );
+        let extended_tag =
+            encode_link_tag(&relationship_name, holon_id, Some(property_values)).unwrap();
+
+        let base_bytes = base_tag.into_inner();
+        let extended_bytes = extended_tag.into_inner();
+
+        assert!(extended_bytes.starts_with(&base_bytes));
+        assert_ne!(extended_bytes, base_bytes);
     }
 }
--- a/happ/crates/holons_loader/src/loader_ref_resolver.rs
+++ b/happ/crates/holons_loader/src/loader_ref_resolver.rs
@@ -936,10 +936,8 @@ impl LoaderRefResolver {
         }
 
         let number_of_targets = write_targets.len() as i64;
-        staged_source.add_related_holons_ungoverned(
-            declared_relationship_name.clone(),
-            write_targets,
-        )?;
+        staged_source
+            .add_related_holons_ungoverned(declared_relationship_name.clone(), write_targets)?;
 
         Ok(number_of_targets)
     }
--- a/happ/crates/holons_loader/src/loader_ref_resolver.rs
+++ b/happ/crates/holons_loader/src/loader_ref_resolver.rs
@@ -108,7 +108,7 @@ impl LoaderRefResolver {
     ///
     /// Multi-pass orchestration (deterministic):
     ///   1) Pass-2a: DescribedBy -> with_descriptor()
-    ///   2) Pass-2b: Extends -> add_related_holons()
+    ///   2) Pass-2b: Extends -> add_related_holons_ungoverned()
     ///   3) Pass-2c: process remaining declared relationship references
     pub fn resolve_relationships(
         context: &Arc<TransactionContext>,
@@ -903,7 +903,7 @@ impl LoaderRefResolver {
 
     /// Performs the actual write:
     /// - DescribedBy: exactly one target → `with_descriptor`
-    /// - Others: batch → `add_related_holons`
+    /// - Others: batch → `add_related_holons_ungoverned`
     fn write_relationship(
         mut staged_source: StagedReference,
         declared_relationship_name: &RelationshipName,
@@ -936,7 +936,10 @@ impl LoaderRefResolver {
         }
 
         let number_of_targets = write_targets.len() as i64;
-        staged_source.add_related_holons(declared_relationship_name.clone(), write_targets)?;
+        staged_source.add_related_holons_ungoverned(
+            declared_relationship_name.clone(),
+            write_targets,
+        )?;
 
         Ok(number_of_targets)
     }
--- a/host/crates/holons_loader_client/src/parser.rs
+++ b/host/crates/holons_loader_client/src/parser.rs
@@ -128,13 +128,15 @@ pub fn parse_files_into_load_set(
     // We collect all per-file issues rather than failing fast so the caller
     // can report a comprehensive summary to the user.
     let mut issues: Vec<ImportFileParsingIssue> = Vec::new();
+    let mut inverse_target_tracker = HasInverseTargetUniquenessTracker::default();
 
     for import_file in &content_set.files_to_load {
-        if let Err(issue) = parse_single_import_file_into_bundle(
+        if let Err(issue) = parse_single_import_file_into_bundle_with_tracker(
             context,
             &load_set_ref,
             import_file,
             &content_set.schema.raw_contents,
+            &mut inverse_target_tracker,
         ) {
             issues.push(issue);
         }
@@ -181,19 +183,24 @@ pub fn create_holon_load_set(
 ///
 /// On success, the `HolonLoadSet` now contains a new bundle for this file.
 /// On failure, a single `ImportFileParsingIssue` is returned.
-pub fn parse_single_import_file_into_bundle(
+fn parse_single_import_file_into_bundle_with_tracker(
     context: &Arc<TransactionContext>,
     load_set_ref: &HolonReference,
     import_file: &FileData,
     schema_json: &str,
+    inverse_target_tracker: &mut HasInverseTargetUniquenessTracker,
 ) -> Result<(), ImportFileParsingIssue> {
     let raw_json = import_file.raw_contents.as_str();
     let file_path = PathBuf::from(import_file.filename.clone());
 
     // 2) Validate against the loader JSON Schema and deserialize into
     //    `RawLoaderFileWithSlices<'_>`.
-    let raw_file_with_slices =
-        match validate_and_deserialize_loader_file(raw_json, schema_json, &import_file.filename) {
+    let (raw_file_with_slices, has_inverse_pairs) =
+        match validate_and_deserialize_loader_file_with_pairs(
+            raw_json,
+            schema_json,
+            &import_file.filename,
+        ) {
             Ok(wrapper) => wrapper,
             Err(err) => {
                 return Err(ImportFileParsingIssue {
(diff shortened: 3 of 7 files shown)


evomimic (approved):
Approve.

Reviewed against #516 and the linked #444 follow-up. This matches the issue contract:

- loader validation now rejects shared `HasInverse` targets, including cross-file/load-set cases
- ordinary staged mutation now uses instance-anchored declared resolution for duplicate policy, preserves duplicate-allowed relationships, treats exact repeats as idempotent no-ops, and rejects inverse names as mutation input
- commit-time SmartLink persistence now suppresses equivalent forward and inverse links with one shared equivalence model
- regression coverage was extended in unit tests and sweettests around duplicate handling, inverse-name rejection, load validation, and replayed graph-only commits

CI is green on PR head `18aaecbb448ac7153def8f71a3bca72f03ba351b`, and manual verification was reported good for `build:happ`, `build:host`, `npm test`, and `npm start`.

PR #589, opened 2026-07-13
Implement TDL R4 semantic validation and Canonical Holon IR diffnpm

### LINKED ISSUE #578: TDL R4: Layered validation and semantic diff over Canonical Holon IR
# TDL R4: Layered validation and semantic diff over Canonical Holon IR

## Summary
Implement TDL R4 by making validation layers explicit and adding semantic diff services over the shared Canonical Holon IR.

## Context
The current `map_schema_semantic` crate already acts as the shared semantic center for `tools/map-schema`, with JSON and TDL lowering into a common `SemanticModel`, shared diagnostics, and a derived `SymbolIndex`.

What is missing for TDL R4 is not another semantic core, but clearer separation of validation layers and a diff surface that works over the shared IR instead of JSON text noise.

Relevant repository facts:

- `shared_crates/map_schema_semantic` already provides:
  - `SemanticModel`
  - `Schema`
  - `TypeDescriptor`
  - `SymbolIndex`
  - `Diagnostic`
  - `ComparableSemanticModel`
- `tools/map-schema` already exposes:
  - `check`
  - `compile`
  - `decompile`
  - `symbols`
- there is no `diff` command yet
- diagnostics currently carry `severity`, `kind`, and `origin`, but not a named validation layer
- round-trip tests already compare semantic equivalence, but only as equality checks rather than a review-oriented diff

DevDocs sources inspected:

- `/Users/stevemelville/dev/map-dev-docs/docs/core/type-system/tdl/tdl-impl-plan-v2.md`
- `/Users/stevemelville/dev/map-dev-docs/docs/core/type-system/tdl/tdl-spec.md`
- `/Users/stevemelville/dev/map-dev-docs/docs/roadmap/desc-driven-impl-plan.md`

Repo files inspected:

- `/Users/stevemelville/dev/map-proto/map-dev/map-holons/.github/ISSUE_TEMPLATE/enhancement.md`
- `/Users/stevemelville/dev/map-proto/map-dev/map-holons/shared_crates/map_schema_semantic/src/lib.rs`
- `/Users/stevemelville/dev/map-proto/map-dev/map-holons/shared_crates/map_schema_semantic/src/diagnostics.rs`
- `/Users/stevemelville/dev/map-proto/map-dev/map-holons/shared_crates/map_schema_semantic/src/schema_ir.rs`
- `/Users/stevemelville/dev/map-proto/map-dev/map-holons/shared_crates/map_schema_semantic/src/schema_index.rs`
- `/Users/stevemelville/dev/map-proto/map-dev/map-holons/tools/map-schema/src/lib.rs`
- `/Users/stevemelville/dev/map-proto/map-dev/map-holons/tools/map-schema/src/main.rs`
- `/Users/stevemelville/dev/map-proto/map-dev/map-holons/tools/map-schema/src/tdl_compiler.rs`
- `/Users/stevemelville/dev/map-proto/map-dev/map-holons/tools/map-schema/src/schema_to_loader_ir.rs`

## Proposed change
Add a validation-and-diff layer on top of the Canonical Holon IR.

Primary work:

- introduce explicit validation-layer categorization for diagnostics
- preserve existing severity, kind, and origin behavior
- keep parser/source syntax errors separate from shared semantic validation
- add semantic diff support over `SemanticModel` and its derived schema view
- add generic holon/literal diff where needed for preserved literal residue
- add a `map-schema diff` CLI command that compares two corpora through the shared semantic layer

The implementation should keep the shared crate as the source of truth for semantic comparison and diagnostics, while `tools/map-schema` remains the native adapter and CLI surface.

## Scope
This issue should cover:

- `shared_crates/map_schema_semantic`
- `tools/map-schema`
- CLI/script wiring in the root workspace if needed for the new command
- unit tests for validation-layer attribution and diff output

This issue should not cover:

- runtime loader unification
- LSP/editor services
- changes to canonical core schema import JSON under `host/import_files/map-schema/core-schema/`
- runtime handles or transaction APIs in the semantic IR

## Dependencies
- TDL R1 toolchain stabilization is already treated as complete in `tdl-impl-plan-v2.md`
- TDL R3 appears to be substantially present in the repository
- existing JSON -> TDL -> JSON semantic round-trip behavior must remain intact
- `check`, `compile`, and `decompile` workflows must continue to work

## Acceptance criteria
- [ ] Diagnostics can report which validation layer produced them
- [ ] Validation layers are distinguishable at the shared semantic layer, not only in CLI text
- [ ] `map-schema check` renders layered diagnostics deterministically
- [ ] semantic diff ignores irrelevant JSON noise such as formatting and provenance-only metadata
- [ ] schema diff is derived from the shared semantic model, not from raw JSON text
- [ ] generic holon/literal diff remains separate from schema diff but uses the same semantic center
- [ ] a `map-schema diff` command exists and compares two corpora through Canonical Holon IR
- [ ] existing round-trip and semantic equivalence tests still pass
- [ ] new unit tests cover validation-layer attribution and diff behavior

## Testing
Add focused coverage in both the shared semantic crate and the map-schema tool crate.

Suggested tests:

- diagnostic layer attribution
- unresolved reference reporting remains intact
- schema diff ignores origin-only and formatting-only differences
- schema diff reports meaningful descriptor, property, and relationship changes
- generic holon diff handles preserved literal residue without conflating it with schema semantics
- CLI help and command behavior for the new `diff` subcommand

Suggested validation commands:

    cargo test --manifest-path shared_crates/map_schema_semantic/Cargo.toml
    cargo test --manifest-path tools/map-schema/Cargo.toml
    npm run map-schema:check:coreschema
    npm run map-schema:compile:coreschema

## Risks
- Diff output could become a second semantic source of truth if it is not clearly derived from `SemanticModel`
- layer terminology could drift from the DevDocs language unless it is kept explicit
- review-oriented diffs may need richer ordering than equality-oriented comparisons
- adding shared-crate coverage may surface workspace assumptions that need to stay WASM-safe

## Additional context
TDL R4 in `tdl-impl-plan-v2.md` calls for:

- explicit validation layers
- semantic review over the shared semantic representation
- semantic diff that ignores irrelevant JSON noise
- schema diff and generic holon diff that are related but not conflated

The TDL spec reinforces the design constraints:

- JSON and TDL both normalize through the Canonical Holon IR
- TDL is descriptor-only
- literal residue should be preserved when concise TDL cannot express content truthfully
- validation, not grammar, owns semantic rules such as extensibility and relationship constraints

evomimic:
# Summary

Implements TDL R4 semantic validation and adds Canonical Holon IR semantic diff support in `map-schema`.

This PR introduces layered semantic diagnostics over the shared schema IR, moves schema-authoring validation out of the symbol-index layer into a dedicated validator, and adds a new semantic diff workflow that compares fully lowered Canonical Holon IR models instead of source text or emitted JSON shape.

## Changes

- Add layered diagnostics in `map_schema_semantic` so findings are classified by validation responsibility, including:
  - syntax
  - IR structural
  - descriptor kind
  - reference symbol
  - schema-aware
  - semantic fidelity
  - runtime loader boundary
- Extend diagnostic kinds to cover the R4 validation surface, including:
  - duplicate local members
  - projected `TypeKind` mismatches
  - extends incompatibility and cycles
  - invalid cardinality bounds
  - inverse relationship completeness and mismatch cases
  - effective key-rule failures and authored-key mismatches
- Add a shared validator module that performs source-neutral Canonical Holon IR validation after lowering and reference resolution
- Narrow `SymbolIndex` responsibilities so it handles symbol construction, reference resolution, duplicate symbol detection, unresolved references, and role/kind compatibility, while leaving schema-aware validation to the validator layer
- Add validation-only normalization for relationship defaults and inverse-pair synthesis so validation can reason over effective semantics without mutating emitted or decompiled artifacts
- Wire validation into both lowering paths:
  - JSON import lowering
  - TDL lowering
- Add `map-schema diff` to compare two schema corpora by Canonical Holon IR semantics
- Require both sides of semantic diff to lower without blocking diagnostics before producing a diff
- Add coverage for:
  - core-schema validation remaining diagnostic-free
  - semantic fidelity across JSON and decompiled TDL
  - changed semantic fields being surfaced by diff
  - invalid diff inputs reporting diagnostics instead of partial comparison
- Fix bootstrap anchor lowering so `TypeDescriptor` compares consistently across JSON and TDL representations

## Why

Issue 578 needed the schema-authoring environment to validate TDL at least as strongly as the current Airtable-backed flow while keeping Canonical Holon IR semantics source-neutral.

This PR establishes that split cleanly:
- source adapters own parsing and source-format conveniences
- Canonical Holon IR owns semantic validation
- semantic comparison works over valid lowered models rather than textual or formatting differences

## Testing

- [x] `cargo test --manifest-path shared_crates/map_schema_semantic/Cargo.toml`
- [x] `cargo test --manifest-path tools/map-schema/Cargo.toml`
- [x] User-verified broader workspace health:
  - `npm clean`
  - hApp build
  - host build
  - `npm` tests
  - `npm start`

## Notes

- Validation-only normalization is intentionally kept out of emitted JSON and decompiled TDL so artifact shape remains faithful to authored/imported content.
- Semantic diff currently requires diagnostic-free lowering on both sides. Partial diff over invalid inputs remains deferred.
- General inheritance flattening remains deferred for Issue 578, except for the narrow effective-key behavior already agreed as in scope.

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -132,6 +132,82 @@ _Avoid_: Internal execution state
 The future materialization boundary that converts bound navigation state into descriptor-defined projection output.
 _Avoid_: Implicit row materialization by order, distinct, skip, or limit
 
+**Source Adapter**:
+A format-specific translator that lowers authored or imported content into the Canonical Holon IR or projects it out to a concrete output format.
+_Avoid_: Semantic core, IR owner
+
+**Validation Layer**:
+A named responsibility boundary for diagnostics, such as syntax, IR structural, schema-aware, or runtime-loader boundary validation.
+_Avoid_: Unclassified diagnostic bucket
+
+**Diagnostic Origin**:
+The source location, symbol, or authored/imported element that a diagnostic should be attributed to.
+_Avoid_: Validation responsibility category
+
+**Canonical Holon IR**:
+The source-neutral semantic middle shared by MAP schema tooling, validation, diffing, code generation, and future editor services.
+_Avoid_: JSON model, TDL AST, loader DTO
+
+**Semantic Diff**:
+A review-oriented comparison derived from valid Canonical Holon IR models rather than concrete source text.
+_Avoid_: JSON text diff, parser recovery diff
+
+**Semantic Fidelity Check**:
+A source-neutral comparison of normalized Canonical Holon IR models used to verify compile/decompile or adapter round-trips without treating formatting, ordering, or shorthand syntax as semantic differences.
+_Avoid_: Textual round-trip equality
+
+**Runtime Loader Projectability**:
+The Issue 578 boundary check that Canonical Holon IR can be projected into the existing loader/import shape without changing loader behavior or creating a new runtime import path.
+_Avoid_: Loader unification, runtime validation rewrite
+
+**TypeKind-Compatible Inheritance**:
+The schema-authoring rule that a type descriptor may extend another type descriptor only when both descriptors have the same TypeKind.
+_Avoid_: Non-extensible descriptor rule as the primary inheritance test
+
+**Projected TypeKind**:
+The TypeKind derived from a descriptor's source-neutral semantic kind for validation and comparison.
+_Avoid_: Treating authored `instance_type_kind` as the sole semantic authority
+
+**Relationship Pair Completeness**:
+The schema-authoring invariant that every declared relationship descriptor has exactly one inverse relationship descriptor paired with it.
+_Avoid_: Optional inverse metadata for declared relationships
+
+**Optional Property Suffix**:
+The `?` suffix on a descriptor property name that marks the property as optional in schema authoring.
+_Avoid_: Treating missing required properties as implicitly optional
+
+**Required Slot Table**:
+The fixed Issue 578 validation table that names required semantic slots per descriptor kind without deriving optionality from the full meta-descriptor graph.
+_Avoid_: Full meta-schema requiredness engine
+
+**Post-Lowering Requiredness**:
+The rule that required descriptor slots are validated on Canonical Holon IR after a source adapter has applied only the explicit defaults and conveniences of its source format.
+_Avoid_: Inventing semantic defaults during IR validation
+
+**Closed-World Authoring Uniqueness**:
+The Issue 578 uniqueness boundary that checks duplicates only inside the model being validated, not across the persisted MAP runtime or broader social graph.
+_Avoid_: Global MAP uniqueness validation
+
+**Blocking Semantic Diagnostic**:
+An Issue 578 validation finding that makes the Canonical Holon IR unsuitable for successful check, compile, or semantic diff.
+_Avoid_: Treating invalid schema semantics as warnings
+
+**Effective Descriptor View**:
+The flattened descriptor structure produced by following a descriptor's Extends chain and combining inherited structural declarations and affordances.
+_Avoid_: Local-only descriptor surface
+
+**Descriptor Flattener**:
+An existing descriptor-layer operation that computes effective inherited descriptor members, such as instance properties, instance relationships, commands, or dances.
+_Avoid_: Reimplemented inheritance flattening in schema tooling
+
+**Effective Key Rule**:
+The key rule selected for a descriptor or instance after resolving authored `UsesKeyRule`, applicable `Extends` lineage, and `DescribedBy`/type fallback according to MAP schema key-generation semantics.
+_Avoid_: General inherited descriptor flattening
+
+**Descriptor Semantics Service**:
+A small source-neutral service for descriptor-shaped graph rules that must be shared by TDL IR validation and runtime descriptor behavior.
+_Avoid_: Broad descriptor engine, ad hoc schema-tooling clone
+
 ## Relationships
 
 - Query PRO3 removes the transaction-level query envelope family rather than replacing it with a new command-owned query envelope.
@@ -160,6 +236,24 @@ _Avoid_: Implicit row materialization by order, distinct, skip, or limit
 - Without a future **ProjectStep**, navigation results should remain holon-native and selected by the **Output Binding**.
 - **Old-World Relationship Traversal Types** may remain for compatibility, but **New-World Query Contract** design must not depend on them.
 - A **Spec Revision Session** closes when the team produces a stable artifact for one coherent design slice, such as a revised issue body.
+- A **Source Adapter** owns format-specific parsing and syntax validation; the **Canonical Holon IR** owns source-neutral semantics.
+- A **Validation Layer** may classify diagnostics from adapters or derived semantic services, but adapter-specific validation responsibilities must not contaminate **Canonical Holon IR** semantics.
+- A diagnostic's **Validation Layer** identifies which responsibility boundary failed; its **Diagnostic Origin** identifies where the author or tool should look.
+- A **Semantic Diff** compares only valid **Canonical Holon IR** models; invalid inputs produce layered diagnostics instead of partial semantic differences.
+- A **Semantic Fidelity Check** compares canonical semantics, not source formatting, JSON field order, or equivalent adapter shorthand.
+- **Runtime Loader Projectability** checks whether schema-authoring semantics can flow into the existing loader/import shape without changing loader behavior.
+- **TypeKind-Compatible Inheritance** allows multi-step `Extends` chains, but rejects inheritance across TypeKind boundaries.
+- **Projected TypeKind** is used for **TypeKind-Compatible Inheritance**; authored `instance_type_kind` values are preserved and validated against the projection when present.
+- **Relationship Pair Completeness** requires every declared relationship to have exactly one inverse and every inverse relationship to point back to a declared relationship.
+- **Optional Property Suffix** is the authoring signal for property optionality; properties without the suffix are required by the corresponding meta-descriptor.
+- An **Effective Descriptor View** includes inherited instance properties, instance relationships, commands, dances, key rules, operators, and related structural declarations according to each descriptor family.
+- A **Descriptor Flattener** is the authoritative implementation for effective inherited descriptor members when inherited member validation is in scope; TDL R4 defers general member flattening.
+- **Effective Key Rule** resolution is the narrow TDL R4 inheritance exception because Airtable already performs key generation by walking `Extends` and, when needed, `DescribedBy`/type fallback.
+- A **Descriptor Semantics Service** is deferred for TDL R4 unless a required validation cannot be expressed without it.
+- The **Required Slot Table** is the R4 source for descriptor-kind requiredness checks.
+- **Post-Lowering Requiredness** keeps source-format defaults in the adapter and treats missing required Canonical Holon IR slots as validation diagnostics.
+- **Closed-World Authoring Uniqueness** covers duplicate canonical symbols or keys, duplicate local property names, duplicate local relationship names, and duplicate inverse ownership in the current input model.
+- **Blocking Semantic Diagnostics** include missing required slots, unresolved references, duplicate symbols, wrong descriptor or meta-kind, relationship inverse-pair violations, inheritance graph violations, TypeKind mismatches, effective-key failures, and generated-key mismatches.
 
 ## Example dialogue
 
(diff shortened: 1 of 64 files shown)


evomimic:
## Review Findings

1. `map-schema check` still does not produce structured syntax-layer diagnostics, so one of Issue 578’s explicit in-scope validation layers is still missing.

   In [tools/map-schema/src/tdl_compiler.rs](/Users/stevemelville/dev/map-proto/map-dev/map-holons/tools/map-schema/src/tdl_compiler.rs:147), `check_input_string` calls `parse_tdl_file(...)` and propagates parse failures as `anyhow::Result` errors. In [tools/map-schema/src/tdl_compiler.rs](/Users/stevemelville/dev/map-proto/map-dev/map-holons/tools/map-schema/src/tdl_compiler.rs:264), the parser still returns plain `anyhow!` errors for malformed TDL such as multiple schema declarations, unexpected braces, and unrecognized top-level declarations. I could not find any producer for `DiagnosticLayer::Syntax`, so syntax failures are not represented in the layered diagnostic model the issue and issue comment call for.

   Risk: malformed TDL still bypasses the new shared diagnostic structure instead of being surfaced as syntax-layer findings.

   Smallest fix direction: add a syntax-diagnostic path that lowers parser failures into `Diagnostic { layer: DiagnosticLayer::Syntax, ... }` with origin metadata, and cover it with `check`-path tests.

2. The new semantic diff only compares `ComparableSemanticModel`, which drops preserved literal residue and header content, so the PR does not satisfy the issue’s “generic holon/literal diff remains separate from schema diff” requirement.

   In [tools/map-schema/src/lib.rs](/Users/stevemelville/dev/map-proto/map-dev/map-holons/tools/map-schema/src/lib.rs:190), `diff_inputs` lowers both sides and immediately compares `left_model.comparable_signature()` vs `right_model.comparable_signature()`. In [shared_crates/map_schema_semantic/src/schema_ir.rs](/Users/stevemelville/dev/map-proto/map-dev/map-holons/shared_crates/map_schema_semantic/src/schema_ir.rs:394), `comparable_signature()` includes only a narrowed subset of schema and descriptor semantics. It does not include `header`, `literal_properties`, or `literal_relationships`, even though those fields are explicitly preserved on `Schema` and `TypeDescriptor` in the same file.

   Risk: diffs can incorrectly report `no semantic diff` when preserved literal residue or header-only content changes, which is exactly the class of review signal the issue calls out as needing a separate generic holon/literal diff path.

   Smallest fix direction: either implement the separate literal/holon diff surface promised by the issue, or narrow the PR and issue scope explicitly and remove that acceptance criterion before merge.

evomimic:
## Review Findings

1. **[P1] Property-value validation is hard-coded instead of being derived from Core Schema.**

   `TypeKind.HolonFoo` is a representative failure of general schema conformance, not a TypeKind-specific string mismatch. Core Schema already declares the complete validation chain: the holon's applicable property is `TypeKind.PropertyType`; that property has `ValueType -> TypeKind`; and the `TypeKind` enum declares its valid members through `Variants`. A schema-driven validator should follow that graph and reject an undeclared enum variant.

   Instead, `shared_crates/map_schema_semantic/src/validator.rs:206-235` special-cases `instance_type_kind`, while `type_kind_family_for_text` at lines 772-785 recognizes hard-coded prefixes. Consequently, `TypeKind.HolonFoo` is accepted as Holon. Tightening this code to an exact string table would still duplicate schema knowledge in the compiler and would not establish the reusable validation pattern needed for other enum-valued or constrained properties.

   I reproduced the symptom by changing the zoo's value to `TypeKind.HolonFoo`: `map-schema check` emitted no property-value/schema-conformance diagnostic for that descriptor.

   **Risk:** invalid holon property values pass unless each property receives bespoke compiler logic; Core Schema can evolve without the validator inheriting its rules, and TDL and JSON validation can drift.

   **Fix direction:** introduce a source-neutral holon schema-conformance path derived from the Canonical Holon IR/Core Schema graph. It should resolve the describing type, applicable property descriptor, `ValueType`, and value constraints; enum membership must be validated through `Variants`. TDL should invoke this after lowering, and loader-side validation should invoke the same semantic service after references are resolvable, returning failures as `HolonLoadError` relationships rather than host-parser exceptions. Remove the current TypeKind-specific prefix/family validation. The `TypeKind.HolonFoo` regression should assert the generic chain `property -> ValueType -> enum Variants -> undeclared variant`, with no branch keyed to `instance_type_kind`, `TypeKind`, or `HolonFoo`. Once values are schema-valid, the separate `Extends` rule can compare the resolved TypeKind enum-variant identities of child and parent.

2. **[P1] Malformed JSON still bypasses the structured syntax diagnostic layer.**

   Issue 578's clarification explicitly requires both malformed TDL and malformed JSON to produce syntax/source diagnostics. The follow-up now converts TDL parser failures correctly, but `tools/map-schema/src/lib.rs:644-663` still parses JSON with `serde_json::from_str(...)?` into an `anyhow::Error`. JSON inputs used by semantic diff therefore fail outside `Diagnostic { layer: Syntax, kind: InvalidSyntax, origin: ... }`, and parsing remains fail-fast across a corpus.

   **Risk:** the same authoring failure has different API and presentation behavior depending on source adapter, contrary to the layered-diagnostic contract.

   **Smallest fix:** have the JSON adapter convert serde parse failures into syntax diagnostics with JSON origin metadata, and cover the JSON semantic-loading/diff path with a malformed-input test.

3. **[P2] Relationship-literal syntax diagnostics report the wrong source line.**

   `parse_reference_block` in `tools/map-schema/src/tdl_compiler.rs:852-864` returns only strings. Its callers parse those strings after the complete block has been consumed and pass `self.current_origin()` (for example lines 469-471 and 644-646). A malformed relationship literal on line 6 is consequently reported at line 8, after the closing braces.

   **Risk:** structured diagnostics point authors at the wrong line, reducing the usefulness of the new syntax layer.

   **Smallest fix:** retain an `Origin` with each reference-block line and pass that saved origin into `parse_literal_relationship_line`; add a line-attribution regression.

The previous review findings are resolved: TDL parser failures now enter the syntax layer, and semantic diff reports preserved header/literal residue separately from canonical schema changes. Manual verification was reported successful for hApp build, host build, `npm test`, and `npm start`. CI `fmt` passes on `cfaad025`; CI `test` is still running. The PR is not merge-ready until the findings above are addressed and CI passes on the resulting head.

evomimic:
## Review Findings

1. **[P1] Production validation still bypasses full descriptor-driven conformance and retains the superseded required-slot table.**

   The authoritative TDL spec requires every non-bootstrap Canonical Holon IR holon to be validated through its resolved descriptor and explicitly says requiredness must come from effective property descriptors, not a descriptor-kind switch. The R4 plan likewise requires required/additional properties, relationships, cardinality, and descriptor/meta-descriptor conformance through the shared kernel.

   `shared_crates/map_schema_semantic/src/validator.rs:37-40` invokes `validate_canonical_model_values`, while `shared_crates/map_schema_semantic/src/descriptor_conformance.rs:32-44` filters the complete conformance result down to `InvalidConformanceValue` only. Required-property, additional-property, allowed-relationship, and relationship-cardinality violations are therefore discarded in the production path. At the same time, `validator.rs:99-147` still hard-codes required fields by `DescriptorKind`, including name-based exceptions, instead of deriving those obligations from descriptor data.

   This leaves the branch with two competing validation regimes and allows descriptor-governed invalidity outside primitive/enum values to escape `map-schema check`. It also conflicts with the updated DevDocs and with the branch's own shared-kernel architecture. `CONTEXT.md` still describes the obsolete `Required Slot Table` and projected-TypeKind approach, so it now disagrees with the authoritative specifications.

   **Fix direction:** make the production validator invoke complete descriptor-driven conformance, bring the canonical Core Schema corpus into conformance rather than filtering violations, implement the remaining descriptor-driven relationship checks required by R4, and remove the hard-coded required-slot/name rules once their descriptor-driven equivalents are active. Update `CONTEXT.md` to defer to the current DevDocs terminology and rules.

2. **[P1] Cross-TypeKind `Extends` edges are not validated.**

   `DiagnosticKind::TypeKindMismatch` exists at `shared_crates/map_schema_semantic/src/diagnostics.rs:61`, but there is no producer for it. `validate_inheritance_graph` at `shared_crates/map_schema_semantic/src/validator.rs:195-248` checks only that both endpoints are types and that the chain is structurally valid; it never compares the endpoints' descriptor-governed TypeKind identities.

   This is observable in the checked-in corpus. Running:

       cargo run --manifest-path tools/map-schema/Cargo.toml -- check schema-src tools/map-schema/examples/diagnostic-zoo.tdl

   does not report the `PropertyExtendsHolon` declaration at `tools/map-schema/examples/diagnostic-zoo.tdl:51`, even though a property type extends `ZooBook.HolonType`. This violates the TDL rule that `T Extends B` is valid only when `T` and `B` have the same TypeKind.

   **Fix direction:** after both endpoints pass ordinary descriptor conformance, resolve their TypeKind enum-variant identities through descriptor data and compare those identities in the graph-wide inheritance pass. Do not restore projected declaration-kind families, name inference, or string-prefix matching. Add the R4-plan tests for same-TypeKind chains and cross-TypeKind rejection.

3. **[P2] One duplicate symbol suppresses every graph-backed diagnostic in the corpus.**

   `CanonicalDescriptorGraph::new` rejects duplicate canonical identities, but `validate_model` uses `if let Ok(graph)` at `shared_crates/map_schema_semantic/src/validator.rs:38` and silently skips conformance, inheritance, and effective-key validation when graph construction fails. The diagnostic zoo demonstrates the consequence: with its duplicate `DuplicateSymbolThing`, neither `CycleA` nor `CycleB` receives an inheritance-cycle diagnostic. Removing only the duplicate declaration causes both cycle diagnostics to appear.

   A duplicate remains blocking, but it should not hide independent authoring errors elsewhere in the corpus. That defeats R4's accumulated-diagnostics goal and makes the diagnostic zoo materially less useful.

   **Fix direction:** preserve duplicate-symbol diagnostics while constructing a graph over unambiguous identities, or otherwise retain enough partial graph state to validate unaffected nodes. At minimum, graph construction failure must become a structured diagnostic rather than being silently discarded; the corpus test should prove that a duplicate and an unrelated cycle are reported together.

The shared descriptor-semantics extraction, JSON/TDL syntax diagnostics, semantic residue comparison, and loader error transport are materially improved and aligned with the revised architecture. However, the two missing R4 semantic behaviors above mean PR 589 is not merge-ready. GitHub `fmt` passes on `cced3833`; the full `test` job is still running at the time of review.

PR #610, opened 2026-07-29
609 pvl pr 5   holon update and delete validation

### LINKED ISSUE #609: PVL PR 5 - Holon Update and Delete Validation
### 1. Summary (Required)

**What is the enhancement?**

Add descriptor-independent lifecycle validation for `HolonNode` update and delete operations, across the two-layer PVL architecture:

- pure-core rules over a substrate-free description of a resolved lifecycle target
- substrate-adapter resolution of that target from the Holochain operation, without deserializing its entry
- Integrity wiring for all five lifecycle arms
- retirement of the superseded no-op delete path

This implements PVL Design Spec v0.5 §§3.3, 9.2, 10.2, 11, and 15 (decision 8), and PVL Implementation Plan v2.5 PR 5.

The rules are:

- **Update target** — an update's `original_action_address` must reference a `Create` action carrying the `HolonNode` entry type. A target `Update` is rejected even when it carries a `HolonNode`: update-to-update chains are invalid in MAP's root-addressed lineage topology.
- **Delete target** — a delete must name a `Create` **or** an `Update` carrying the `HolonNode` entry type. Both are valid because each names one exact persisted version.

**This is no longer gated on Storage SL2.** Design Spec v0.5 corrects the earlier claim that early activation "would reject every current update." It would not: `ForUpdateNewVersion` calls `create_entry` (`commit_functions.rs:409-417`), and `update_entry` appears nowhere in the workspace, so PVL sees no Holochain `Update` to reject. PR 5 lands now as proactive hardening against peer-authored `Update` operations, and SL2 inherits an already-enforced contract.

---

**Why is this needed?**

The Integrity zome currently applies no meaningful lifecycle validation. Across the five arms where `HolonNode` lifecycle semantics reach validation:

| Arm | `lib.rs` | Current behavior |
|---|---|---|
| `StoreEntry::UpdateEntry` | 75-77 | accepts unconditionally |
| `RegisterUpdate::Entry` | 80-91 | accepts unconditionally |
| `StoreRecord::UpdateEntry` | 254-276 | accepts a `Create` **or** an `Update` as the original |
| `RegisterDelete` | 92-102 | routes to a no-op |
| `StoreRecord::DeleteEntry` | 277-337 | fetches and fully deserializes the original entry, then routes to the same no-op |

Three specific defects follow:

1. **Update-to-update chains are accepted.** `StoreRecord::UpdateEntry` explicitly permits an `Update` original:

   ```rust
   if !matches!(original_record.action(), Action::Create(_) | Action::Update(_)) {
   ```

   The Knowledge Evolution Architecture requires every update to be rooted at the lineage-root `Create`. Nothing enforces that today, and the two other update arms enforce nothing at all.

2. **Delete validation is a no-op that reads as a real check.** `validate_delete_holon_node` (`holons_integrity/src/holon_node.rs:5-11`) discards its argument and calls `validate_delete_holon()` (`shared_validation/src/validation_helpers.rs:9-11`), which returns `Ok(())`. Worse, the `PersistenceDelete` it is handed carries `prev_action`, not `deletes_address` — the field a delete-target rule needs is commented out of the struct (`hc_action.rs:66-67`), so the current abstraction cannot express the rule even in principle.

3. **The delete arm pays an unbounded decode.** `StoreRecord::DeleteEntry` resolves `must_get_valid_record`, then calls `EntryTypes::deserialize_from_type` to learn the original entry's type — up to `MAX_HOLON_NODE_BYTES` (262,144) of decode work per validated delete, to recover a fact the action already carries in its `entry_type`. This is the cost multiplication Design Spec §12.2 forbids.

Consequently a peer-authored native Holochain update can today target another update, or an action carrying the wrong entry type, without receiving a deterministic PVL rejection.

---

**Required, all merged:**

- **PVL PR 1 — #591, merged in #595.** Provides `PvlViolation::InvalidUpdateTarget` (`MAP-PVL-1301`), `PvlViolation::InvalidDeleteTarget` (`MAP-PVL-1303`), and the `HolonError::PvlViolation` fan-out.
- **PVL PR 2 — #600, merged in #602.** Provides the raw-op envelope guard (`lib.rs:61`), the adapter seam in `holons_guest_integrity`, the `ExternResult<Result<T, PvlViolation>>` shape, and the synthetic-`Op` test fixtures this PR reuses.
- **PVL PR 4 — #604, merged in #605.** Provides the fixed diagnostic-token convention (`ACTION_HASH_LOCAL_ID_KIND`) that the lifecycle-kind tokens follow.

**Sequencing constraints:**

- The PR 2 raw-op envelope guard must remain the first thing the Integrity `validate` callback does, ahead of op flattening and any lifecycle dependency resolution.
- **Storage SL2 is not a prerequisite.** PR 5 owns and activates the root-addressed update rule; Storage Implementation Plan task 12 is revised so SL2 *verifies* it rather than wiring a second check. SL2 will be the first work to exercise it.
- **Storage SL1 part 2 — #594 / PR #606 — no architectural conflict.** That PR touches `persistence_layer/smartlink*`, `smartlink_adapter.rs`, `commit_functions.rs`, and `core_types/smartlink/types.rs`, none of which are HolonNode lifecycle arms. It does, however, touch `happ/Cargo.lock`, `host/Cargo.lock`, and `flake.lock`. If #606 merges first, rebase and keep its dependency resolution rather than regenerating lockfiles.

**Documentation prerequisite:** Design Spec v0.5 and Implementation Plan v2.5 are on the `pvl-evolution` branch of `map-dev-docs` and should land before or alongside this issue.

---

**How would you solve it?**

#### 4.1 Pure core — `shared_crates/shared_validation/src/holon_node_lifecycle.rs`

Lifecycle rules are the first PVL rules whose input the pure core cannot see for itself. Per Design Spec §3.3, the seam is a narrow substrate-free description of the *resolved* target — not a mirror of a Holochain action:

```rust
pub enum TargetActionKind { Create, Update, Other }
pub enum TargetEntryKind { HolonNode, OtherAppEntry, NonAppEntry, Absent }

pub struct LifecycleTarget {
    pub action_kind: TargetActionKind,
    pub entry_kind: TargetEntryKind,
}

pub fn validate_update_target(target: &LifecycleTarget) -> Result<(), PvlViolation>;
pub fn validate_delete_target(target: &LifecycleTarget) -> Result<(), PvlViolation>;
```

Use closed enums, not strings and not a general mirror of Holochain's `Action`. Each variant carries a fixed `&'static str` diagnostic token, following PR 4's `ACTION_HASH_LOCAL_ID_KIND` convention. Each rule checks action kind before entry kind, so the violation's `expected_target_kind` / `actual_target_kind` name the axis that actually failed.

The pure core never receives a Holochain `Action` or `Record`, and never requires entry content to reach a verdict. That narrowness is what makes the rules exhaustively testable without a conductor — which matters, because no current write path can exercise the update rule.

#### 4.2 Substrate adapter — `happ/crates/holons_guest_integrity/src/holon_node_lifecycle.rs`

Resolves the target and maps it to `LifecycleTarget`, per Design Spec §9.2:

- **Update target → `must_get_valid_record`.** The lineage root is the structural parent of every version in its lineage, so inductive validity is wanted; root-addressing bounds the induction at exactly one hop. A lineage must not be rooted at an entry the network reports invalid.
- **Delete target → `must_get_action`.** A delete names one exact version, asserts no lineage, and carries no structure forward.

Both read the target's action kind and entry type from the **action** — `record.action()` for the update path, the returned `SignedActionHashed` for the delete path. Neither requests or deserializes the target entry.

Per Design Spec §11, the adapter must never intercept a `must_get_*` short-circuit and convert it into a violation. Keep the `ExternResult<Result<T, PvlViolation>>` shape from PR 2: the outer result carries host and callback-execution failures, the inner result carries completed validation.

**Entry-type classification must use scoped app-entry identity.** Do not classify a target as `HolonNode` from `EntryDefIndex(0)` alone; the defining integrity-zome scope is part of the identity. This is defensive rather than urgent — `happ/workdir/dna.yaml` currently declares exactly one integrity zome, so `zome_index` is always `0` — but the check is cheap and a second integrity zome would otherwise silently misclassify.

Two ways to satisfy it, with different costs:

- **Update path (no host call):** compare the target's `AppEntryDef` against the *new* entry's `AppEntryDef` from the op being validated. The envelope guard has already established the new entry is `HolonNode`, so equality gives scoped identity for free.
- **Delete path:** there is no new entry to compare against, so obtain this zome's index via `zome_info()`. That is a local host call, not a DHT dependency, so it does not count against `MAX_VALIDATION_DEPENDENCIES_PER_OP` — but it does need a mock expectation in adapter tests.

Note this is deliberately stricter than the existing comment at `holon_node_envelope.rs:80-81` ("Since this code only runs for this zome, the entry index alone identifies whether this is a HolonNode"). Either tighten that path too or record why the target check is stricter than the op check.

#### 4.3 Integrity routing — `happ/zomes/integrity/holons_integrity/src/lib.rs`

Route all five arms in the §2 table through the adapter — one adapter path per operation type, so the five arms cannot drift apart. Map `Err(PvlViolation)` to:

```text
ValidateCallbackResult::Invalid(violation.to_string())
```

which produces the existing deterministic messages:

- `MAP-PVL-1301: update target is invalid`
- `MAP-PVL-1303: delete target is invalid`

Lifecycle checks must run **after** the raw-op envelope guard at `lib.rs:61`, preserving PR 2's exit criterion that an oversized entry pays no decode or dependency cost before rejection.

#### 4.4 Violation contract correction

Rename `InvalidUpdateTarget`'s fields from `expected_entry_kind` / `actual_entry_kind` to `expected_target_kind` / `actual_target_kind`, matching `InvalidDeleteTarget`. The variant's primary failure is an *action*-kind mismatch, so the current names misdescribe it. `MAP-PVL-1301` is unchanged, and the rename precedes any release.

Apply to:

- `shared_crates/type_system/integrity_core_types/src/pvl_error.rs` — variant, constructors, exhaustive matches, serde round-trip test
- `host/map-sdk/src/internal/wire-types/references.ts` — the `PvlViolationWire` variant shape (`:395`) and the `isPvlViolationWire` field checks (`:1006`). `HolonErrorWire` (`:416`) only wraps `{ PvlViolation: PvlViolationWire }` and needs no change.
- `host/map-sdk/tests/wire-types.test.ts`

`ImmutableNativeFieldChanged` (`MAP-PVL-1302`) is deliberately **not** implemented and must never be returned. The current persisted shape is `{original_id, property_map}` and SL2 removes `original_id`, leaving no native field carrying a cross-version invariant. Add a comment at its definition recording that it is reserved and unused.

#### 4.5 Retirement of the superseded delete path

Per AGENTS.md, obsolete paths are removed deliberately rather than left beside their replacements. This PR removes what it replaces:

- `validate_delete_holon_node` (`holons_integrity/src/holon_node.rs:5-11`)
- `validate_delete_holon` (`shared_validation/src/validation_helpers.rs:9-11`)
- both `PersistenceDelete::new` construction sites (`lib.rs:94`, `lib.rs:328`)
- the now-orphaned `PersistenceDelete` struct (`hc_action.rs:59-70`) — self-contained, since `PersistenceAction::Delete` is a unit variant

**Deliberately left in place:** `PersistenceAction`, `PersistenceCreate`, and `PersistenceUpdate` are already dead (never constructed anywhere) but are not orphaned *by this PR*; `PersistenceCreateLink` / `PersistenceDeleteLink` belong to the SmartLink arms and should be retired by PVL PR 6. Removing them here would be incidental cleanup. Their intent to be removed is recorded in Design Spec §3.3.

---

**What does this impact?**

**New files:**
- `shared_crates/shared_validation/src/holon_node_lifecycle.rs`
- `happ/crates/holons_guest_integrity/src/holon_node_lifecycle.rs`

**Modified:**
- `happ/zomes/integrity/holons_integrity/src/lib.rs` — five lifecycle arms; remove target-entry deserialization and the obsolete `PersistenceDelete` conversions
- `happ/zomes/integrity/holons_integrity/src/holon_node.rs` — remove `validate_delete_holon_node`
- `shared_crates/shared_validation/src/{lib.rs,validation_helpers.rs}` — export new module; remove no-op
- `happ/crates/holons_guest_integrity/src/lib.rs` — export new module
- `happ/crates/holons_guest_integrity/Cargo.toml` — add `hdi` `mock` feature as a **dev-dependency** for adapter tests
- `shared_crates/type_system/integrity_core_types/src/pvl_error.rs` — field rename, reserved-variant comment
- `shared_crates/type_system/integrity_core_types/src/hc_action.rs` — remove `PersistenceDelete`
- `host/map-sdk/src/internal/wire-types/references.ts` and `host/map-sdk/tests/wire-types.test.ts` — PVL wire-contract alignment
- `tests/sweetests/tests/fixture_cases/delete_holon_fixture.rs` — existing delete regression, if strengthening is needed

**Cross-boundary fan-out to call out in review:** the `InvalidUpdateTarget` rename is a PR 1 contract change riding in PR 5. A Rust build will not flag a stale TypeScript mirror, so the SDK change must land in the same PR.

**Behavioral impact:**

- Peer-authored native Holochain updates must be rooted directly at a valid `Create` carrying `HolonNode`; update-to-update chains are rejected.
- Deletes of exact versions authored by either a `Create` or an `Update` remain valid.
- Deletes naming unsupported action or entry kinds are rejected.
- Temporarily unavailable dependencies continue to produce `UnresolvedDependencies`.
- **No operation authored by current MAP coordinator code is rejected.** Creates, version-producing writes (which emit Creates), and deletes of valid HolonNodes all continue to pass. The new rejections are reachable only by peer-authored operations or by a future coordinator bug.

**Out of scope (Storage SL2 owns):** removing `original_id`; `VersionMetadata` / `StoredHolonNode`; switching `ForUpdateNewVersion` to `update_entry`; predecessor lineage resolution; saved-holon record decoding; `Predecessor` / `Successor` publication. The PR 2 breadcrumb comments marking the anticipated SL2 switch remain in place.

**Also out of scope:** enforcing immutable fields across versions; lineage-head selection, tombstoning, history traversal, or delete policy; descriptor-dependent lifecycle validation; SmartLink envelope or provenance validation (PVL PR 6); dependency-budget accounting (PVL PR 8); the un-coded `Invalid` strings on `HolonNodeUpdates` / `AllHolonNodes` link deletion (Storage SL5 removes those link types); broad cleanup of the remaining `Persistence*` types.

---

This PR has an unusual coverage profile worth stating up front. **The update rule cannot be exercised through a conductor** — no MAP path authors a Holochain `Update`. Negative delete targets are equally unconstructible in-DNA: `HolonNode` is this DNA's only app entry type, and `delete_entry` against a non-entry-creation action is expected to fail in the coordinator before an op reaches validation (worth confirming during implementation). Negative coverage is therefore pure-unit and mocked-adapter until Storage SL2.

#### Pure lifecycle tests

Exhaust the closed action-kind × entry-kind model for both rules. At minimum:

- update against `Create` + `HolonNode` is valid
- update against `Update` + `HolonNode` returns `InvalidUpdateTarget`
- update against another action kind fails before entry-kind evaluation
- update against `Create` + another entry kind returns `InvalidUpdateTarget`
- delete against `Create` + `HolonNode` is valid
- delete against `Update` + `HolonNode` is valid
- delete against another action kind returns `InvalidDeleteTarget`
- delete against `Create` or `Update` + another entry kind returns `InvalidDeleteTarget`
- violations expose the stable codes and deterministic messages

This is the payoff for giving the pure core a narrow enumerated input instead of a substrate mirror: no Holochain dependency, no mocks, full coverage.

#### Adapter and Integrity tests

Reuse and extend PR 2's synthetic `Op` fixtures (`holon_node_envelope.rs:191-292`). Unlike PR 2's pre-dependency envelope tests, these resolve dependencies, so they need HDI mocking: `hdi`'s `mock` feature provides `MockHdiT`, and `set_hdi` installs it. `HDI` is a `thread_local!`, so parallel cargo test threads are naturally isolated and each test sets its own mock. Mock `must_get_valid_record`, `must_get_action`, and `zome_info` where scoped classification needs it.

Prove that:

- each of the three update arms reaches the update rule, and each of the two delete arms reaches the delete rule
- update validation calls `must_get_valid_record` exactly once; delete validation calls `must_get_action` exactly once
- no lifecycle path requests or deserializes the target entry
- each violation maps to the exact `MAP-PVL` callback message
- a host or dependency error from either `must_get_*` propagates unchanged through the outer `ExternResult` and is never converted into a `PvlViolation`
- an oversized update entry is rejected by the envelope guard before any lifecycle dependency mock is called
- entry-type classification distinguishes the full scoped `HolonNode` identity from an equal entry index in another scope

The `must_get_*` propagation test is the one that matters most: it is the only mechanical check on the `Invalid` vs `UnresolvedDependencies` distinction, which no conductor test can reach.

#### Wire-contract tests

- `InvalidUpdateTarget` serializes with `expected_target_kind` / `actual_target_kind` (Rust round-trip)
- the TypeScript guard **accepts** the renamed payload
- the TypeScript guard **rejects** the obsolete `expected_entry_kind` / `actual_entry_kind` payload — this is what catches a half-finished rename

#### Conductor regression

Reuse the existing `tests/sweetests/tests/fixture_cases/delete_holon_fixture.rs` scenario: its create/commit/delete sequence must continue to commit with the delete-target rule active. Strengthen it only if it does not currently drive a delete to completion.

#### Validation commands

Narrowest applicable checks while iterating, then:

- `npm run fmt:check`
- `npm run check -w map-happ` (must confirm the `mock` dev-dependency does not leak into the WASM build)
- `npm run check`
- `npm run test:unit`
- `npm run sweetest` (requires Nix)

Report any check that could not be run.

---

- [ ] A substrate-independent lifecycle facts model exists in `shared_validation`, using closed enums rather than strings or an `Action` mirror.
- [ ] Update validation accepts only a `Create` carrying the scoped `HolonNode` entry type; update-to-update chains are rejected as `MAP-PVL-1301`.
- [ ] Delete validation accepts a `Create` or `Update` carrying the scoped `HolonNode` entry type; other action or entry kinds are rejected as `MAP-PVL-1303`.
- [ ] Action kind is evaluated before entry kind in both rules.
- [ ] `HolonNode` classification uses complete scoped app-entry identity, not entry-definition index alone.
- [ ] Update targets resolve through `must_get_valid_record`; delete targets through `must_get_action`.
- [ ] No lifecycle path requests or deserializes a target entry; each op performs at most one DHT dependency lookup.
- [ ] Dependency-resolution failures cannot be converted into a `PvlViolation`, proven by a mocked-adapter test.
- [ ] All three update arms and both delete arms route through the lifecycle adapter; `holons_integrity` contains callbacks and violation mapping only.
- [ ] The raw envelope guard remains ahead of flattening and dependency resolution; an oversized entry is rejected before any decode or dependency request.
- [ ] `InvalidUpdateTarget` uses `expected_target_kind` / `actual_target_kind` in Rust, in `PvlViolationWire`, and in `isPvlViolationWire`.
- [ ] Rust serde tests and TypeScript guards cover the renamed wire shape, including rejection of the obsolete field names.
- [ ] `ImmutableNativeFieldChanged` is documented as reserved and unused, and is never returned.
- [ ] `validate_delete_holon_node`, `validate_delete_holon`, both `PersistenceDelete::new` calls, and `PersistenceDelete` are removed.
- [ ] Pure lifecycle tests exhaust the closed action-kind × entry-kind model.
- [ ] Synthetic-op tests cover all five lifecycle arms and their exact PVL messages.
- [ ] The existing legitimate-delete sweetest still passes.
- [ ] Existing core-schema import, commit, and delete behavior is unchanged.
- [ ] Formatting, hApp/WASM compilation, unit, SDK, and sweetest checks pass, or any environmental limitation is reported.

---

<details>

**Defer PR 5 until Storage SL2 lands.** Rejected: the gate rested on a factual error (see §1), SL2 is not expected within the PVL plan's horizon, and deferring leaves three update arms accepting unconditionally and delete validation a no-op for that entire period.

**Split into PR 5a (delete, now) and PR 5b (update, with SL2).** Rejected: the two rules share one facts type, one adapter module, and one set of op arms. Splitting them would wire the same seam twice for no risk reduction, since the update rule rejects nothing MAP authors today.

**Permit update-to-update chains.** Rejected: MAP uses `Update.original_action_address` as a lineage-root pointer. Every native update must reference the root `Create`, not its immediate predecessor.

**Extend `PersistenceAction` / `PersistenceDelete` with the missing fields.** Rejected. That model mirrors Holochain's action shape rather than any rule's actual input, which is why all eight of its construction sites currently discard their output. Uncommenting `entry_type` would require inventing a `PersistenceEntryType` mirror as well, and the record-shaped abstraction pulls toward `must_get_valid_record` plus entry decode — the exact cost we are removing. A purpose-built facts type is narrower, exhaustively testable, and structurally forecloses the decode.

**Put the lifecycle rules directly in the adapter with no pure type.** Rejected: it places a normative PVL rule where coordinator preflight can never reach it, contradicting Design Spec §3.3 and PR 8's requirement that Integrity and preflight execute identical shared pure-core validation.

**Use `must_get_action` for the update target too.** Rejected: `must_get_action` explicitly does not guarantee the target was validated (hdi `entry.rs:40-58`), so a lineage could be rooted at an entry the network reports invalid. The lineage root is a structural parent, which is precisely the inductive-validation case `must_get_valid_record` exists for, and root-addressing bounds it at one hop.

**Use `must_get_valid_record` for the delete target too.** Rejected: a delete names one exact version and carries no lineage structure forward, so the cheaper action-only fetch establishes everything the rule needs.

**Continue deserializing target entries.** Rejected: the verdict needs only facts the target action already carries. Re-decoding adds bounded-but-large work per validated op and creates an avoidable validation cost multiplier.

---

### 9. Risks or Concerns

**A dormant rule is a rule nobody has watched run.** The update path gets no conductor coverage until SL2 switches the write path. Mitigation: exhaustive pure-core coverage of all kind combinations, mocked-adapter coverage of the resolution path, and an explicit note in Storage SL2's verification list that its "update rooted at an update is rejected" test is the first real exercise of this rule.

**Silently converting a `must_get_*` failure into a violation** would turn temporarily unavailable valid data into permanent rejection. Mitigation: the `ExternResult<Result<T, PvlViolation>>` return shape, a mocked propagation test, and an explicit review item.

**The `mock` dev-dependency must not reach the WASM build.** Enabling `hdi`'s `mock` feature pulls in `mockall`. Resolver v2 should keep dev-dependency features out of the normal build, but `npm run check -w map-happ` must confirm it, per AGENTS.md's WASM-safety and dependency-discipline rules.

**Entry-index-only classification could misclassify another zome's entry** if a second integrity zome is ever added. Currently theoretical — the DNA declares one integrity zome — which is exactly why it is easy to leave unfixed and expensive to discover later.

**Cost of `must_get_valid_record` on three update arms.** One authored update is validated by three separate ops, each fetching the full lineage-root record. This is per-op by design and within the budget of 8, but PR 8's dependency-budget work should confirm no arm is reachable more than once per op.

**Five arms could drift.** Each operation type must have exactly one adapter path; duplicated per-arm logic is how the current `StoreRecord` / `RegisterUpdate` inconsistency arose in the first place.

**TypeScript mirror drift.** The rename will not fail a Rust build if the mirror is missed. This is the standing risk PR 1 flagged; the durable fix remains a CI guard, tracked separately.

**Lockfile churn.** Storage SL1 part 2 (#606) may update Holochain patch versions and lockfiles while this work proceeds — integration churn, but no architectural dependency.

---

### 10. Additional Context

Authoritative sources on the `map-dev-docs` `pvl-evolution` branch:

- [PVL Design Specification v0.5](https://github.com/memetic-activation-platform/map-dev-docs/blob/pvl-evolution/docs/core/validation/pvl-design-spec.md) — §§3.3, 9.2, 10.2, 11, 15(8)
- [PVL Implementation Plan v2.5](https://github.com/memetic-activation-platform/map-dev-docs/blob/pvl-evolution/docs/core/validation/pvl-impl-plan.md) — PR 5
- [Storage Layer Implementation Plan](https://github.com/memetic-activation-platform/map-dev-docs/blob/pvl-evolution/docs/core/guest/storage-layer-services/storage-layer-impl-plan.md) — SL2 task 12, revised to verify rather than implement
- [Descriptor-Driven MAP Implementation Milestone Plan](https://github.com/memetic-activation-platform/map-dev-docs/blob/pvl-evolution/docs/roadmap/desc-driven-impl-plan.md)

Repository grounding:

- `happ/zomes/integrity/holons_integrity/src/lib.rs`
- `happ/zomes/integrity/holons_integrity/src/holon_node.rs`
- `happ/crates/holons_guest_integrity/src/holon_node_envelope.rs`
- `shared_crates/shared_validation/src/validation_helpers.rs`
- `shared_crates/type_system/integrity_core_types/src/pvl_error.rs`
- `shared_crates/type_system/integrity_core_types/src/hc_action.rs`
- `host/map-sdk/src/internal/wire-types/references.ts`
- `host/map-sdk/tests/wire-types.test.ts`
- `tests/sweetests/tests/fixture_cases/delete_holon_fixture.rs`
- `happ/workdir/dna.yaml` — one integrity zome

Current baseline:

- PVL PR 1 merged in #595; PR 2 in #602; PR 3 in #603; PR 4 in #605.
- `ForUpdateNewVersion` still publishes a new `Create`.
- No MAP write path invokes `update_entry`.

</details>

owleyeview:
## PVL PR 5 — Holon Update and Delete Validation

Closes #609. Implements PVL Design Spec v0.5 §§3.3, 9.2, 10.2, 11, 15(8) and PVL Implementation Plan v2.5 PR 5: descriptor-independent lifecycle validation for `HolonNode` update and delete operations across the two-layer PVL architecture.

### What changed

**Pure core** (`shared_crates/shared_validation/src/holon_node_lifecycle.rs` — new)
- Closed, substrate-free facts model: `TargetActionKind` × `TargetEntryKind` → `LifecycleTarget`, with fixed `&'static str` diagnostic tokens per variant.
- `validate_update_target`: only a `Create` carrying `HolonNode` is a valid update target — update-to-update chains are rejected (`MAP-PVL-1301`), because `original_action_address` is a lineage-root pointer in MAP's root-addressed topology.
- `validate_delete_target`: a `Create` **or** `Update` carrying `HolonNode` is valid (`MAP-PVL-1303` otherwise) — each names one exact persisted version.
- Action kind is evaluated before entry kind, so the violation reports the axis that actually failed.

**Substrate adapter** (`happ/crates/holons_guest_integrity/src/holon_node_lifecycle.rs` — new)
- Update targets resolve via `must_get_valid_record` (inductive validity for the structural lineage root, bounded at one hop); delete targets via `must_get_action` (exact version, no lineage carried forward).
- Both read action kind and entry type from the **action** — no target entry is ever requested or deserialized. This removes the up-to-262,144-byte decode the old delete arm paid per validated op.
- `HolonNode` classification uses complete scoped app-entry identity (zome index + entry index), not entry index alone: the update path compares against the validated op's own `AppEntryDef`; the delete path derives scope from a local `zome_info()` call.
- `ExternResult<Result<T, PvlViolation>>` separation preserved: `must_get_*` failures propagate as outer errors (`UnresolvedDependencies`) and are never converted into deterministic violations.

**Integrity routing** (`holons_integrity/src/lib.rs`)
- All five lifecycle arms route through one adapter path per operation type: `StoreEntry::UpdateEntry`, `RegisterUpdate::Entry`, `StoreRecord::UpdateEntry` → update rule; `RegisterDelete`, `StoreRecord::DeleteEntry` → delete rule.
- The PR 2 raw-op envelope guard remains the first thing `validate` does; an oversized entry is rejected before any decode or dependency lookup (now mechanically proven by a mock that fails if a dependency is requested).
- `RegisterDelete` now validates `deletes_address` — the old path only had `prev_action` available and validated nothing.

**Violation contract correction**
- `InvalidUpdateTarget` fields renamed `expected_entry_kind`/`actual_entry_kind` → `expected_target_kind`/`actual_target_kind` (matching `InvalidDeleteTarget`; the variant's primary failure is an action-kind mismatch). Applied in Rust, `PvlViolationWire`, and `isPvlViolationWire`; `MAP-PVL-1301` and messages unchanged.
- `ImmutableNativeFieldChanged` (`MAP-PVL-1302`) documented as reserved and never constructed.

**Retirement of the superseded path**
- Removed: `validate_delete_holon_node`, the no-op `validate_delete_holon`, both `PersistenceDelete::new` sites, and the `PersistenceDelete` struct. SmartLink `Persistence*` types are untouched (PVL PR 6's scope).

### Behavioral impact

- **No operation authored by current MAP coordinator code is rejected.** Version-producing writes still emit `Create`s (`update_entry` appears nowhere in the workspace), so the update rule is proactive hardening against peer-authored operations; Storage SL2 will be the first to exercise it natively.
- Deletes of exact `HolonNode` versions (`Create` or `Update` originals) remain valid; deletes naming any other action or entry kind are now rejected (the old arm early-accepted non-app originals).
- Temporarily unavailable dependencies still yield `UnresolvedDependencies`, never a PVL rejection.

### Testing

- **Pure core:** both rules exhaust the full 3×4 action-kind × entry-kind matrix, asserting exact violations, tokens, axis precedence, and stable codes/messages.
- **Mocked adapter:** hdi 0.7.1's documented `MockHdiT` is not actually generated by its `mock` feature; tests use a local `mockall` implementation of `HdiT` installed via `set_hdi` (the pattern hdi's docs recommend). Covers valid/invalid targets on both paths, scoped-identity discrimination (equal entry index in another zome scope), exactly-one-dependency-lookup per op, no entry fetch, and unchanged propagation of `must_get_*` failures — the mechanical check on the `Invalid` vs `UnresolvedDependencies` distinction, which no conductor test can reach.
- **Callback level:** synthetic-op tests prove all three update arms and both delete arms route through the lifecycle adapter (exact `MAP-PVL` messages), positive paths return `Valid`, and the envelope guard rejects oversized entries before any dependency mock is touched.
- **Conductor:** the negative cases are unconstructible in-DNA (no MAP path authors a Holochain `Update`; `HolonNode` is the only app entry type), so negative coverage is pure-unit and mocked by design. The existing `delete_holon_fixture` sweetest passes with the delete rule active.
- `mockall` is confined to `[target.'cfg(not(target_arch = "wasm32"))'.dev-dependencies]` and the hdi `mock` feature is not enabled, keeping the wasm32 test build (`test:unit:no-run`) and the WASM artifact clean.

All checks pass: `npm run fmt:check`, `npm run check`, `npm run test:unit`, `npm run sweetest`.

### Notes for reviewers

- The `InvalidUpdateTarget` rename is a PR 1 wire-contract change riding in this PR; the TypeScript mirror lands in the same commit, and a TS test rejects the obsolete field names to catch a half-finished rename.
- Lockfile churn is the `mockall` dev-dependency chain only.
- Out of scope (Storage SL2 / later PVL PRs): `original_id` removal, `update_entry` switch, predecessor lineage resolution, SmartLink validation (PR 6), dependency-budget accounting (PR 8), remaining `Persistence*` cleanup.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3485,6 +3485,7 @@ dependencies = [
  "holochain_serialized_bytes",
  "holons_guest_integrity",
  "integrity_core_types",
+ "mockall",
  "serde",
  "shared_validation",
 ]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2475,6 +2475,7 @@ dependencies = [
  "holochain_serialized_bytes",
  "holochain_serialized_bytes_derive",
  "holochain_wasmer_guest",
+ "mockall",
  "paste",
  "serde",
  "serde_bytes",
@@ -3471,6 +3472,7 @@ dependencies = [
  "hdi",
  "holochain_serialized_bytes",
  "integrity_core_types",
+ "mockall",
  "serde",
  "shared_validation",
 ]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2475,7 +2475,6 @@ dependencies = [
  "holochain_serialized_bytes",
  "holochain_serialized_bytes_derive",
  "holochain_wasmer_guest",
- "mockall",
  "paste",
  "serde",
  "serde_bytes",
--- a/happ/Cargo.lock
+++ b/happ/Cargo.lock
@@ -468,7 +468,6 @@ dependencies = [
  "holochain_serialized_bytes",
  "holochain_serialized_bytes_derive",
  "holochain_wasmer_guest",
- "mockall",
  "paste",
  "serde",
  "serde_bytes",
--- a/happ/Cargo.lock
+++ b/happ/Cargo.lock
@@ -770,6 +770,7 @@ dependencies = [
  "holochain_serialized_bytes",
  "holons_guest_integrity",
  "integrity_core_types",
+ "mockall",
  "serde",
  "shared_validation",
 ]
--- a/happ/Cargo.lock
+++ b/happ/Cargo.lock
@@ -31,6 +31,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "anstyle"
+version = "1.0.14"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "940b3a0ca603d1eade50a4846a2afffd5ef57a9feac2c0e2ec2e14f9ead76000"
+
+[[package]]
 name = "arrayref"
 version = "0.3.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -294,6 +300,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "downcast"
+version = "0.11.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1435fa1053d8b2fbbe9be7e97eca7f33d37b28409959813daefc1446a14247f1"
+
+[[package]]
 name = "dunce"
 version = "1.0.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -316,6 +328,15 @@ name = "fnv"
 version = "1.0.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3f9eec918d3f24069decb9af1554cad7c880e2da24a9afd88aca000531ab82c1"
+
+[[package]]
+name = "fragile"
+version = "2.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8878864ba14bb86e818a412bfd6f18f9eabd4ec0f008a28e8f7eb61db532fcf9"
+dependencies = [
+ "futures-core",
+]
 
 [[package]]
 name = "futures"
@@ -447,6 +468,7 @@ dependencies = [
  "holochain_serialized_bytes",
  "holochain_serialized_bytes_derive",
  "holochain_wasmer_guest",
+ "mockall",
  "paste",
  "serde",
  "serde_bytes",
@@ -735,6 +757,7 @@ dependencies = [
  "hdi",
  "holochain_serialized_bytes",
  "integrity_core_types",
+ "mockall",
  "serde",
  "shared_validation",
 ]
@@ -865,6 +888,32 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f8ca58f447f06ed17d5fc4043ce1b10dd205e060fb3ce5b979b8ed8e59ff3f79"
 
 [[package]]
+name = "mockall"
+version = "0.13.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "39a6bfcc6c8c7eed5ee98b9c3e33adc726054389233e201c95dab2d41a3839d2"
+dependencies = [
+ "cfg-if",
+ "downcast",
+ "fragile",
+ "mockall_derive",
+ "predicates",
+ "predicates-tree",
+]
+
+[[package]]
+name = "mockall_derive"
+version = "0.13.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "25ca3004c2efe9011bd4e461bd8256445052b9615405b4f7ea43fc8ca5c20898"
+dependencies = [
+ "cfg-if",
+ "proc-macro2",
+ "quote",
+ "syn 2.0.114",
+]
+
+[[package]]
 name = "must_future"
 version = "0.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -915,6 +964,32 @@ checksum = "12922b6296c06eb741b02d7b5161e3aaa22864af38dfa025a1a3ba3f68c84577"
 dependencies = [
  "der",
  "spki",
+]
+
+[[package]]
+name = "predicates"
+version = "3.1.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ada8f2932f28a27ee7b70dd6c1c39ea0675c55a36879ab92f3a715eaa1e63cfe"
+dependencies = [
+ "anstyle",
+ "predicates-core",
+]
+
+[[package]]
+name = "predicates-core"
+version = "1.0.10"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cad38746f3166b4031b1a0d39ad9f954dd291e7854fcc0eed52ee41a0b50d144"
+
+[[package]]
+name = "predicates-tree"
+version = "1.0.13"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d0de1b847b39c8131db0467e9df1ff60e6d0562ab8e9a16e568ad0fdb372e2f2"
+dependencies = [
+ "predicates-core",
+ "termtree",
 ]
 
 [[package]]
@@ -1202,6 +1277,12 @@ dependencies = [
  "quote",
  "unicode-ident",
 ]
+
+[[package]]
+name = "termtree"
+version = "0.5.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8f50febec83f5ee1df3015341d8bd429f2d1cc62bcba7ea2076759d315084683"
 
 [[package]]
 name = "thiserror"
--- a/happ/crates/holons_guest_integrity/Cargo.toml
+++ b/happ/crates/holons_guest_integrity/Cargo.toml
@@ -28,5 +28,9 @@ path = "../../../shared_crates/type_system/integrity_core_types"
 [dependencies.shared_validation]
 path = "../../../shared_crates/shared_validation"
 
+[dev-dependencies]
+hdi = { version = "0.7", features = ["mock"] }
+mockall = "0.13"
+
 [dev-dependencies.base_types]
 path = "../../../shared_crates/type_system/base_types"
--- a/happ/crates/holons_guest_integrity/Cargo.toml
+++ b/happ/crates/holons_guest_integrity/Cargo.toml
@@ -29,7 +29,6 @@ path = "../../../shared_crates/type_system/integrity_core_types"
 path = "../../../shared_crates/shared_validation"
 
 [dev-dependencies]
-hdi = { version = "0.7", features = ["mock"] }
 mockall = "0.13"
 
 [dev-dependencies.base_types]
--- a/happ/crates/holons_guest_integrity/src/holon_node_envelope.rs
+++ b/happ/crates/holons_guest_integrity/src/holon_node_envelope.rs
@@ -14,7 +14,7 @@ use crate::HolonNode;
 /// `HolonNode` is currently the sole app entry declared by this integrity
 /// zome, so its index is fixed at zero. Adding or reordering app-entry
 /// definitions requires updating this constant and the associated op tests.
-const HOLON_NODE_ENTRY_DEF_INDEX: EntryDefIndex = EntryDefIndex(0);
+pub(crate) const HOLON_NODE_ENTRY_DEF_INDEX: EntryDefIndex = EntryDefIndex(0);
 
 /// Fixed structured-diagnostic reason used when exact `ActionHash` parsing fails.
 ///
@@ -77,8 +77,10 @@ fn holon_node_entry_bytes(op: &Op) -> Option<&[u8]> {
         _ => return None,
     };
 
-    // Since this code only runs for this zome, the entry index alone
-    // identifies whether this is a HolonNode.
+    // The op envelope is interpreted through this zome's declared EntryTypes, so
+    // its entry index selects HolonNode without another host call. Lifecycle
+    // targets are resolved actions that may belong to another integrity-zome
+    // scope and therefore require the stricter full AppEntryDef comparison.
     match entry_type {
         EntryType::App(app_entry_def)
             if app_entry_def.entry_index == HOLON_NODE_ENTRY_DEF_INDEX => {}
--- a/happ/crates/holons_guest_integrity/src/holon_node_lifecycle.rs
+++ b/happ/crates/holons_guest_integrity/src/holon_node_lifecycle.rs
@@ -1,0 +1,380 @@
+//! Holochain adapter for descriptor-independent `HolonNode` lifecycle validation.
+//!
+//! This module resolves the action named by an update or delete and projects
+//! only the facts needed by `shared_validation`. Target entry content is never
+//! requested or decoded: the action already carries both its lifecycle kind and
+//! its entry type.
+//!
+//! Host and dependency-resolution failures remain in the outer
+//! [`ExternResult`]. The inner result represents a completed PVL verdict, which
+//! prevents temporarily unavailable dependencies from becoming deterministic
+//! validation failures.
+
+use hdi::prelude::*;
+use integrity_core_types::PvlViolation;
+use shared_validation::{
+    validate_delete_target, validate_update_target, LifecycleTarget, TargetActionKind,
+    TargetEntryKind,
+};
+
+use crate::holon_node_envelope::HOLON_NODE_ENTRY_DEF_INDEX;
+
+/// Resolves and validates the lineage root named by a `HolonNode` update.
+///
+/// [`must_get_valid_record`] provides inductive validity for the structural
+/// lineage root. MAP's root-addressed topology bounds this resolution at one
+/// hop: the target must itself be a `Create`, never another `Update`.
+///
+/// `new_entry_def` comes from the already envelope-validated operation. Comparing
+/// the complete [`AppEntryDef`] gives scoped entry identity without another host
+/// call and prevents an equal entry-definition index in another integrity zome
+/// from being mistaken for `HolonNode`.
+pub fn validate_holon_node_update_target(
+    original_action_hash: &ActionHash,
+    new_entry_def: &AppEntryDef,
+) -> ExternResult<Result<(), PvlViolation>> {
+    let target_record = must_get_valid_record(original_action_hash.clone())?;
+    let target = classify_target(target_record.action(), |target_entry_def| {
+        target_entry_def == new_entry_def
+    });
+
+    Ok(validate_update_target(&target))
+}
+
+/// Resolves and validates the exact persisted version named by a `HolonNode` delete.
+///
+/// A delete carries no lineage structure forward, so [`must_get_action`] supplies
+/// every fact the rule needs without fetching the target record or entry.
+/// [`zome_info`] is a local host call used only to establish the defining
+/// integrity-zome scope for `HolonNode`; it is not a DHT dependency lookup.
+pub fn validate_holon_node_delete_target(
+    deletes_address: &ActionHash,
+) -> ExternResult<Result<(), PvlViolation>> {
+    let target_action = must_get_action(deletes_address.clone())?;
+    let holon_node_entry_def =
+        AppEntryDef::new(HOLON_NODE_ENTRY_DEF_INDEX, zome_info()?.id, EntryVisibility::Public);
+    let target = classify_target(target_action.action(), |target_entry_def| {
+        target_entry_def == &holon_node_entry_def
+    });
+
+    Ok(validate_delete_target(&target))
+}
+
+/// Projects a resolved Holochain action into the closed pure lifecycle model.
+///
+/// The caller supplies scoped app-entry classification because the update path
+/// derives scope from the new entry while the delete path derives it from
+/// `zome_info`. Keeping resolution outside this helper makes it impossible for
+/// pure validation to trigger additional host calls.
+fn classify_target(
+    action: &Action,
+    is_holon_node: impl FnOnce(&AppEntryDef) -> bool,
+) -> LifecycleTarget {
+    let action_kind = match action {
+        Action::Create(_) => TargetActionKind::Create,
+        Action::Update(_) => TargetActionKind::Update,
+        _ => TargetActionKind::Other,
+    };
+    let entry_kind = match action.entry_type() {
+        Some(EntryType::App(entry_def)) if is_holon_node(entry_def) => TargetEntryKind::HolonNode,
+        Some(EntryType::App(_)) => TargetEntryKind::OtherAppEntry,
+        Some(_) => TargetEntryKind::NonAppEntry,
+        None => TargetEntryKind::Absent,
+    };
+
+    LifecycleTarget { action_kind, entry_kind }
+}
+
+#[cfg(test)]
+mod tests {
+    use integrity_core_types::PvlViolation;
+    use mockall::mock;
+    use shared_validation::{
+        CREATE_ACTION_KIND, CREATE_OR_UPDATE_ACTION_KIND, HOLON_NODE_ENTRY_KIND, OTHER_ACTION_KIND,
+        OTHER_APP_ENTRY_KIND, UPDATE_ACTION_KIND,
+    };
+
+    use super::*;
+
+    const HOLON_ZOME_INDEX: ZomeIndex = ZomeIndex(0);
+
+    // HDI 0.7.1 exposes mockall behind its `mock` feature but does not generate
+    // the documented `MockHdiT` type. Define the test double locally and install
+    // it through the same thread-local `set_hdi` seam used by the host wrappers.
+    mock! {
+        Hdi {}
+
+        impl HdiT for Hdi {
+            fn verify_signature(&self, input: VerifySignature) -> ExternResult<bool>;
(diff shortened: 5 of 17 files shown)


evomimic (approved):
Approve. PR 610 appears to satisfy issue #609 as written.

The lifecycle validation is now descriptor-independent and split cleanly across the pure-core rules, the Holochain adapter, and the integrity routing. Update targets are restricted to root `Create` actions carrying `HolonNode`, delete targets accept exact `Create` or `Update` `HolonNode` versions, and the obsolete no-op delete path has been removed rather than left in parallel.

I also checked the contract-carrying details called out in the issue:
- all three update arms and both delete arms now route through the shared lifecycle path
- target entry content is no longer fetched or deserialized for lifecycle decisions
- `InvalidUpdateTarget` was renamed to `expected_target_kind` / `actual_target_kind` across Rust and the TypeScript wire mirror
- reserved `MAP-PVL-1302` is documented and still not constructed

CI is passing on the current PR head `474417bb2fa9dafea783ad20bc8b206d3ec48e73`, and manual verification was reported nominal for `build:happ`, `build:host`, `npm test`, and `npm start`.

No blocking logic or scope issues. Non-blocking note only: the most important negative cases are necessarily proven through pure/unit and mocked adapter tests rather than sweetest, but that matches the issue’s stated coverage plan until Storage SL2 starts emitting native `Update` operations.

PR #611, opened 2026-07-30
607 storage sl2

### LINKED ISSUE #607: Storage SL2 - Version-Aware Holon Persistence
### 1. Summary (Required)

**What is the enhancement?**  
Implement Storage SL2, `Version-Aware Holon Persistence`, as the storage-layer `HolonNode` persistence and exact-read contract for `map-holons`: add record-derived `VersionMetadata`, exact-version read APIs, and a semantic `HolonWriteRequest` persistence path that translates MAP create/version intent into Holochain `Create` or root-addressed `Update` operations behind the storage boundary.

---

**Why is this needed?**  
The current guest persistence path does not preserve MAP’s intended version topology at the storage boundary.

Current repository state shows three core mismatches with the storage-layer design:

- `ForUpdateNewVersion` still persists the next holon version through `create_holon_node`, which always calls `create_entry`, so every saved version is currently recorded as an unrelated root `Create` rather than a root-addressed `Update`. See [happ/crates/holons_guest/src/guest_shared_objects/commit_functions.rs](/Users/stevemelville/dev/map-proto/map-dev/map-holons/happ/crates/holons_guest/src/guest_shared_objects/commit_functions.rs:409) and [happ/crates/holons_guest/src/persistence_layer/holon_node.rs](/Users/stevemelville/dev/map-proto/map-dev/map-holons/happ/crates/holons_guest/src/persistence_layer/holon_node.rs:18).
- `try_from_record` inflates `SavedHolon` from the entry body plus the record action hash, but discards whether the record action was a `Create` or `Update` and discards `Update.original_action_address`. See [happ/crates/holons_guest/src/persistence_layer/saved_holon_node.rs](/Users/stevemelville/dev/map-proto/map-dev/map-holons/happ/crates/holons_guest/src/persistence_layer/saved_holon_node.rs:23).
- `HolonNode.original_id` still persists version-related lineage information inside semantic entry content, and higher layers still consume that shape directly. See [shared_crates/type_system/integrity_core_types/src/holon_node_model.rs](/Users/stevemelville/dev/map-proto/map-dev/map-holons/shared_crates/type_system/integrity_core_types/src/holon_node_model.rs:17), [shared_crates/holons_core/src/core_shared_objects/holon/saved.rs](/Users/stevemelville/dev/map-proto/map-dev/map-holons/shared_crates/holons_core/src/core_shared_objects/holon/saved.rs:20), and [happ/zomes/coordinator/holons/src/lib.rs](/Users/stevemelville/dev/map-proto/map-dev/map-holons/happ/zomes/coordinator/holons/src/lib.rs:105).

Without SL2:

- storage cannot distinguish exact version identity from lineage membership in a storage-owned way
- guest code outside storage still interprets Holochain record/action semantics indirectly through entry content
- version-producing MAP writes cannot branch correctly from a lineage root
- exact-version reads cannot return the record-derived metadata required by the Knowledge Evolution Architecture
- the repository remains coupled to the misleading “original holon” naming and `original_id` entry field that SL2 is intended to retire

---

Architecturally, Storage SL2 is independent of SmartLink SL1 and may proceed alongside it per the implementation plan. For this issue, assume the current main branch baseline already includes the merged SL1 work, including the follow-up fixes implied by PR 606 / issue 594 review findings.

Dependencies and sequencing notes:

- Design sources:
  - `/Users/stevemelville/dev/map-dev-docs/docs/core/guest/storage-layer-services/storage-layer-design-spec.md`
  - `/Users/stevemelville/dev/map-dev-docs/docs/core/guest/storage-layer-services/storage-layer-impl-plan.md`
  - `/Users/stevemelville/dev/map-dev-docs/docs/roadmap/desc-driven-impl-plan.md`
- Current repo template:
- Repository-local copies of the DevDocs artifacts are not currently present under `map-holons/docs/`. This issue should reference the source paths above, and the docs should be imported into `map-holons/docs/` as follow-through for repo-local traceability.
- Storage SL5 depends on SL2, so this issue should not remove `AllHolonNodes` or `HolonNodeUpdates` as general cleanup beyond what SL2 strictly needs for correctness.

---

**How would you solve it?**  
Add a semantic storage contract for version-aware `HolonNode` persistence and exact retrieval, keeping Holochain action selection and record interpretation entirely inside the guest storage layer.

Required storage-boundary information shape:

- Add `LineageId`
- Add `VersionMetadata { version_id, lineage_id }`
- Add `StoredHolonNode { holon_node, version_metadata }`
- Add `HolonWriteRequest` with semantic variants equivalent to:
  - `PublishRoot { holon_node }`
  - `PublishVersion { holon_node, predecessor_ids }`

Required storage behavior:

1. Add exact-version reads under `persistence_layer`:
   - `get_holon(&LocalId) -> Result<Option<StoredHolonNode>, HolonError>`
   - `get_holons(&[LocalId]) -> Result<Vec<Option<StoredHolonNode>>, HolonError>`
2. Decode record-derived metadata as follows:
   - `Create` record: `version_id` is the action hash and `lineage_id` is absent
   - `Update` record: `version_id` is the action hash and `lineage_id` is `Update.original_action_address`
   - unsupported actions or malformed persisted records fail explicitly
3. Add semantic `persist_holon(HolonWriteRequest)` under `persistence_layer`:
   - `PublishRoot` uses `create_entry`
   - `PublishVersion` requires at least one predecessor id, loads each exact predecessor record, resolves the lineage root from those records, rejects mismatched lineage roots, then uses `update_entry` against the shared root `Create`
4. Keep immediate predecessor semantics above storage:
   - existing commit processing should continue staging authoritative `Predecessor` and inverse `Successor` SmartLinks
   - storage owns only the `HolonNode` version record and its record-derived metadata
5. Remove `original_id` from persisted `HolonNode` entry content and move lineage/version facts to `VersionMetadata`
6. Retire misleading exact-read naming such as `get_original_holon_node` from the new path; exact reads should be version-addressed, not root-addressed by name
7. Ensure non-storage code consumes abstract storage results instead of interpreting Holochain `Action` or `Record` semantics directly

Repository-grounded implementation areas likely include:

- [happ/crates/holons_guest/src/persistence_layer/holon_node.rs](/Users/stevemelville/dev/map-proto/map-dev/map-holons/happ/crates/holons_guest/src/persistence_layer/holon_node.rs:18)
- [happ/crates/holons_guest/src/persistence_layer/saved_holon_node.rs](/Users/stevemelville/dev/map-proto/map-dev/map-holons/happ/crates/holons_guest/src/persistence_layer/saved_holon_node.rs:23)
- [happ/crates/holons_guest/src/guest_shared_objects/commit_functions.rs](/Users/stevemelville/dev/map-proto/map-dev/map-holons/happ/crates/holons_guest/src/guest_shared_objects/commit_functions.rs:409)
- [happ/crates/holons_guest/src/guest_shared_objects/guest_holon_service.rs](/Users/stevemelville/dev/map-proto/map-dev/map-holons/happ/crates/holons_guest/src/guest_shared_objects/guest_holon_service.rs:237)
- [shared_crates/type_system/integrity_core_types/src/holon_node_model.rs](/Users/stevemelville/dev/map-proto/map-dev/map-holons/shared_crates/type_system/integrity_core_types/src/holon_node_model.rs:17)
- [happ/crates/holons_guest_integrity/src/holon_node.rs](/Users/stevemelville/dev/map-proto/map-dev/map-holons/happ/crates/holons_guest_integrity/src/holon_node.rs:17)
- [shared_crates/holons_core/src/core_shared_objects/holon/saved.rs](/Users/stevemelville/dev/map-proto/map-dev/map-holons/shared_crates/holons_core/src/core_shared_objects/holon/saved.rs:20)
- [happ/zomes/integrity/holons_integrity/src/lib.rs](/Users/stevemelville/dev/map-proto/map-dev/map-holons/happ/zomes/integrity/holons_integrity/src/lib.rs:74)
- [happ/zomes/coordinator/holons/src/lib.rs](/Users/stevemelville/dev/map-proto/map-dev/map-holons/happ/zomes/coordinator/holons/src/lib.rs:105)

Implementation constraints:

- do not expose Holochain `Action`, `Record`, `Update`, or `original_action_address` through non-storage APIs
- do not add latest-version selection, head selection, revision-history traversal, get-all semantics, caching, hydration, or version-graph traversal in this issue
- do not absorb descriptor or relationship semantics into storage

---

**What does this impact?**  
In scope:

- guest storage-layer `HolonNode` persistence and exact retrieval
- record-derived version metadata
- integrity validation for native MAP `Update` handling
- guest commit mapping from staged semantic intent to storage semantic write requests
- `SavedHolon` or adjacent shared runtime shapes that currently depend on persisted `original_id`
- coordinator-side signal/event construction that currently infers lineage from entry content

Likely impacted modules:

- `happ/crates/holons_guest/src/persistence_layer/`
- `happ/crates/holons_guest/src/guest_shared_objects/`
- `happ/crates/holons_guest_integrity/src/`
- `happ/zomes/integrity/holons_integrity/src/`
- `happ/zomes/coordinator/holons/src/`
- `shared_crates/holons_core/src/core_shared_objects/holon/`
- `shared_crates/type_system/integrity_core_types/src/`

Out of scope:

- latest-version or revision-history query APIs
- removal of `AllHolonNodes` and `HolonNodeUpdates` as broad cleanup beyond what SL2 strictly requires
- SmartLink occurrence semantics from SL3
- retirement of the temporary SmartLink facade from SL4
- query/reference-layer hydration or version-graph traversal
- descriptor-owned semantics and policy

---

Add storage-focused and integrity-focused coverage for:

- publishing a root holon and reading it back with `version_id = action hash` and absent `lineage_id`
- publishing two updates against one root and reading them back with distinct `version_id` values but identical `lineage_id`
- publishing sibling updates against the same root to prove storage allows branching
- rejecting a version write when predecessor ids resolve to different lineage roots
- rejecting invalid update topology when `original_action_address` does not reference a `Create` containing a `HolonNode`
- proving `ForCreate` maps to a Holochain `Create`
- proving `ForUpdateNewVersion` maps to a Holochain `Update` rooted at the lineage-root `Create`
- proving a second-generation update remains rooted at the original `Create`, not the immediate predecessor
- exact retrieval of an existing version
- `None` for a missing exact action id
- positional batch retrieval preserving input order, duplicates, missing entries, and cardinality
- explicit failure for unsupported record actions or invalid persisted entry content
- persisted `HolonNode` bytes no longer include `original_id` or any lineage field
- existing `Predecessor` and inverse `Successor` relationship staging still flows through the ordinary SmartLink path unchanged

Suggested validation posture:

- narrow guest compile and unit coverage while iterating
- storage/integrity sweettests for create/update record behavior and retrieval semantics
- regression checks around coordinator signal emission and `SavedHolon` inflation behavior once the storage result shape changes

User-reported branch-start baseline for adjacent PR review work:
- `build:happ` nominal
- `build:host` nominal
- `npm test` nominal
- `npm start` nominal

Those reports are helpful context but are not substitutes for SL2-specific storage and integrity coverage.

---

- A storage-boundary result exists that returns decoded `HolonNode` content together with record-derived `VersionMetadata`
- `get_holon` returns exact-version results by `LocalId` without discarding whether the enclosing record was a `Create` or `Update`
- `get_holons` returns one positional result per requested id, preserving order, duplicates, and missing entries
- A semantic `persist_holon(HolonWriteRequest)` storage API exists and owns Holochain action selection internally
- `PublishRoot` persists via `create_entry`
- `PublishVersion` persists via root-addressed `update_entry` after resolving and validating a shared lineage root from exact predecessor ids
- Version writes fail when predecessor ids resolve to different lineage roots
- Integrity validation rejects MAP update topology that targets an update rather than a lineage-root `Create`
- Persisted `HolonNode` entries no longer carry `original_id`
- Non-storage guest code no longer depends on entry-body `original_id` to infer lineage or exact version semantics
- Immediate predecessor semantics continue through `Predecessor` / `Successor` SmartLink staging rather than through entry-body lineage fields
- No Holochain `Action`, `Record`, or host-function concept escapes the storage-layer boundary
- New tests cover the required create/update/read/error cases, or any unable-to-run checks are explicitly documented

---

<details>

One option is to keep `create_holon_node` as the single persistence path and continue modeling version lineage through `HolonNode.original_id`. That would preserve short-term compatibility, but it keeps record semantics outside storage and prevents native root-addressed `Update` behavior from becoming the storage truth.

Another option is to add latest-version or revision-history APIs in the same PR. That would broaden the issue beyond the storage contract and mix foundational exact-version persistence with higher-level traversal policy. The implementation plan explicitly defers that.

A third option is to leave `SavedHolon` unchanged and bolt metadata onto side channels. That would reduce churn, but it would leave the shared runtime wrapper out of sync with the storage contract and likely preserve the same lineage leakage this issue is trying to remove.

---

### 9. Risks or Concerns

**What could go wrong?**  
- `SavedHolon`, clone behavior, and wire bindings currently assume `original_id` remains part of the persisted `HolonNode` shape. Removing it will likely require coordinated changes across shared runtime types, boundary bindings, and fixtures.
- Coordinator signals currently derive `affected_holon` from entry-body `original_id`; once lineage becomes record-derived metadata, that logic will need a replacement abstraction rather than direct reuse of current fields.
- Integrity validation for updates is currently only scaffolded in comments. Activating native update behavior without tightening validation would leave MAP topology under-enforced.
- Existing helper names such as `get_original_holon_node` and `get_latest_holon_node` encode storage semantics that SL2 is explicitly trying to narrow. Renaming and deprecating those paths without accidental scope creep will take care.
- `HolonNodeUpdates` link infrastructure already exists in the integrity zome, but the SL2 plan says storage should be root-addressed native Holochain updates, not a separate custom revision system. The implementation should be careful not to harden both models in parallel.

---

### 10. Additional Context

**Any supporting material?**  
Source documents inspected:

- `/Users/stevemelville/dev/map-dev-docs/docs/roadmap/desc-driven-impl-plan.md`
- `/Users/stevemelville/dev/map-dev-docs/docs/core/guest/storage-layer-services/storage-layer-design-spec.md`
- `/Users/stevemelville/dev/map-dev-docs/docs/core/guest/storage-layer-services/storage-layer-impl-plan.md`

Repository grounding notes:

- The repo currently has no imported copies of the referenced DevDocs files under `map-holons/docs/`
- The SL2 implementation plan explicitly marks this work as `Planned`, `8` dev points, `Storage`, and “may proceed alongside SL1”
- Assume issue 594 / PR 606 baseline fixes are already merged before starting this issue, so SL2 can target the post-SL1 storage API surface without reopening SmartLink review findings

</details>

nphias:
### Storage SL2 — Version-Aware Holon Persistence (closes #607)

MAP's guest persistence layer did not preserve the version topology MAP intends. Every write —
including a version-producing update — was a `create_entry`, so each "new version" landed as an
unrelated root. `update_entry` appeared nowhere in the repo. Record conversion discarded whether an
action was a `Create` or an `Update` along with `Update.original_action_address`, and version
lineage was carried in entry content via `HolonNode.original_id` — a field that, in practice, was
*always `None`*, so the coordinator's signal arms never actually recovered a lineage.

This PR moves version identity out of entry content and into the record, and makes the storage
layer own which Holochain action a write becomes.

### What this delivers

**A storage-boundary vocabulary** — `core_types::holon_storage`: `LineageId`, `VersionMetadata`,
`StoredHolonNode`, `HolonWriteRequest`, plus the pure `resolve_shared_lineage_root` rule and its
unit tests. No Holochain types cross this boundary. The module header also defines the intended
terms — *version id*, *lineage id*, *predecessor*, *clone source* — because "original" had been
used for all four at different times.

**Storage owns action selection** — `persistence_layer::holon_storage`:

- `persist_holon(HolonWriteRequest)` — `PublishRoot` authors a `Create`; `PublishVersion` resolves
  the lineage root shared by every predecessor and authors an `Update` **against that root**. A
  version of a version therefore stays one hop from the original `Create`, never chaining off its
  immediate predecessor.
- `get_holon(&LocalId)` / `get_holons(&[LocalId])` — exact-version reads. The batch read is
  positional: one slot per requested id, preserving order, duplicates, and gaps. Absence is `None`;
  a defect fails the whole call.
- Mismatched lineage roots, empty or unpersisted predecessors, unsupported action kinds, and
  undecodable entries are each rejected with their own diagnosis.

**Lineage left the entry** — `original_id` is gone from `HolonNodeModel` and `HolonNode`. The
identifier validation stage it required went with it (the reusable width rule and `MAP-PVL-1201`
were kept deliberately). `SavedHolon.original_id` survives to keep the wire shape stable, but is
now *meaningfully populated* with the record-derived lineage root where it was previously always
`None`.

**Integrity activated, not rewritten** — `validate_holon_node_update_target` already implemented
the correct root-addressed one-hop rule but had never executed, because no `Update` was ever
authored. It now runs on every version write. No rule changed.

**Coordinator signals derive lineage from records** rather than entry content. The signal enum
shape is unchanged, so the host receptor needed no code change.

**Commit-response logging** makes the new topology visible from the console: `(Create)` vs
`(Update)` per holon with lineage and predecessor, action selection at the storage boundary, and a
response summary at *both* exits from `commit` — including the incomplete early return, which was
previously the thinnest case. Raise with `WASM_LOG=info`.

### Testing

`npm run sweetest` **43/43**; `npm run check`, `npm run test:unit`, `npm run fmt:check` all pass.
No `Cargo.lock` drift, no SDK fixture drift.

- **18 new sweettests** (`holon_storage_tests.rs`) covering root/version publication, third-
  generation rooting, branching, positional batch reads, every rejection path, and both directions
  of the Integrity update-target rule.
- Action topology is asserted against the **raw stored `Action`**, read through a pre-existing
  extern, so it does not merely prove this feature's decoder agrees with its own writer.
- Pure lineage rules are unit-tested in `core_types`, the tier that actually executes.

### Reviewer notes

1. **This is DHT-breaking.** Pre-SL2 entries were a 2-field msgpack struct and are now permanently
   invalid. Pre-production, so no migration is provided — but run `npm run clean:session` before
   any dev run against existing state.
2. **Two existing sweettest expectations changed**, both direct consequences of a version being a
   version rather than a new holon: the DB-count assertion in `stage_new_version_fixture.rs`, and
   `descriptor_verification_executor.rs`, which now reaches the new version by traversing
   `Successor` from the lineage root instead of finding both versions in get-all. Every original
   assertion is preserved.
3. **`holon_storage_test_probes.rs` ships in production wasm.** It authors a deliberately invalid
   topology so an Integrity rule that is otherwise unreachable from a live conductor can be
   proven. Isolated and documented; not feature-gated because sweettests run against the same DNA
   the host ships, which would make a gate decorative.
4. **Get-all now returns lineage roots**, so after a version-producing update it shows the *old*
   content. Real interim gap — see below.

### What remains in focus

**Next, and load-bearing** — head selection. Get-all and `get_latest_holon_node` return lineage
roots, so a client enumerating holons sees pre-update content until a head-resolution API exists.
This is the motivating problem for the next storage issue, and everything else here waits behind
it: `get_latest_holon_node` and `get_all_revisions_for_holon_node` remain inert, and
`SavedHolon.version` is still hardcoded to `1`. That work should also decide whether to build on
Holochain's native update graph (`get_details`) or on the `HolonNodeUpdates` link index — this PR
deliberately hardens neither.

**Naming debt, ready to collect** — `SavedHolon.original_id` → `lineage_id`, together with the 7
scaffolded `original_holon_node_hash` extern parameters. Deferred only because the first changes
the serialized `HolonWire::Saved` shape and pulls in the SDK, UI, and tryorama mirrors. The
vocabulary is already defined and every affected doc already says what the field means; what is
left is the identifier. A natural companion to the next issue that touches that wire contract.

**Signal contract** — `HolonUpdated.previous_holon` now always equals `affected_holon` under
root-addressed topology. Either drop the field or redefine it as the immediate predecessor sourced
from the `Predecessor` SmartLink; either way it is a wire change deserving its own issue.

**Unverified rather than broken** — multi-agent lineages. `PublishVersion` and Integrity both
assume the lineage root is resolvable at author time, which holds for every current flow because
all authoring is local.


evomimic (approved):
Looks good to me.

This matches Issue #607’s Storage SL2 contract closely: version-aware persistence is now owned by the storage boundary, exact reads return record-derived version metadata, version-producing writes become root-addressed Holochain `Update`s, and lineage no longer depends on `HolonNode.original_id` in entry content.

I also checked the merge-readiness gates on the latest PR head `e18d6f192b1ac7e1e1f672cfb4b89cd0155265b9`:
- PR 611 CI is passing (`ci_pass`, `fmt`, and `test`)
- Manual verification was reported nominal for `build:happ`, `build:host`, `npm test`, and `npm start`

The remaining head-selection/get-all gap is documented in the PR and reads as deliberate follow-up work rather than a miss for this issue, since Issue #607 explicitly scoped exact-version storage in and latest/head resolution out.

Approved.

PR #613, opened 2026-07-30
network free sweetests & dev_mode fix

### LINKED ISSUE #612: [Enhancement] Sweetests should run without being slowed down by holochain networking
### 1. Summary (Required)

**What is the enhancement?**

Sweettests spin up a single in-process conductor that never peers, yet they currently connect to
the public Holochain bootstrap/signal servers and the iroh canary relay, and perform UPnP/SSDP
NAT discovery on the local network.

Two changes make the suite network-free:

- Point every WAN endpoint (`bootstrap_url`, `signal_url`, `relay_url`) at an unroutable local
  address.
- Vendor `portmapper` with UPnP/PCP/NAT-PMP disabled by default, applied via `[patch.crates-io]`
  in the root workspace. This second part is unavoidable: no configuration reachable from our code
  can disable iroh's port mapper.

A third, closely related fix rides along: **the Tauri host's dev mode had the same endpoint gap.**
`launch/config.rs` overrode `bootstrap_url` and `relay_url` but never `signal_url`, so
`MAP_START_MODE=dev` silently kept the public default. That is a one-line config omission with the
same root cause as the sweettest problem — an endpoint left unassigned inherits a public server —
so it is fixed here rather than tracked separately.

**Note the boundary:** this covers the host's *signal endpoint* only. Suppressing the host's
*port-mapping* traffic is a different problem with a different mechanism, and is explicitly out of
scope.

---

**Why is this needed?**

Observed during sweettest runs: an outbound connection to a public IPv6 address (firewall-flagged),
plus SSDP multicast on UDP 1900. This is new behaviour — sweettests did not previously do this.

**The endpoints.** `setup_test_conductor()`
(`tests/sweetests/src/harness/helpers/mock_conductor.rs`) built its conductor with
`SweetConductor::from_standard_config()`, which resolves to `NetworkConfig::default()`
(`holochain_conductor_api-0.6.3/src/config/conductor.rs:359-382`). That hardcodes real public
servers:

| Field | Default |
|---|---|
| `bootstrap_url` | `https://dev-test-bootstrap2.holochain.org` |
| `signal_url` | `wss://dev-test-bootstrap2.holochain.org` |
| `relay_url` | `https://use1-1.relay.n0.iroh-canary.iroh.link/` |

`mem_bootstrap` defaults to `true` under `test-utils`, so bootstrap itself is in-memory — but the
iroh transport still initialises, dials the relay, and runs NAT traversal.

**The port mapper — and why the URLs alone are not enough.** Redirecting the three URLs removes
WAN dialing but leaves the SSDP traffic untouched, because iroh's NAT discovery does not consult
them. Verified in source:

1. `iroh-holochain-0.95.1/src/magicsock.rs:1803` builds the client as
   `portmapper::Client::with_metrics(Default::default(), ..)` — hardcoded to the crate default,
   with no plumbing to override it.
2. `portmapper-0.12.0/src/lib.rs:130-139` — that default enables UPnP, PCP and NAT-PMP.
3. `magicsock.rs:1292` calls `procure_mapping()` at the top of every direct-address update,
   **before** the `if self.relay_map.is_empty() { return }` early-out on line 1299.

So a single-conductor test that asserts local DHT behaviour performs LAN NAT discovery on every
startup regardless of its network config.

**The host dev-mode signal leak.** The same class of bug exists in the Tauri host.
`host/conductora/plugins/tauri-plugin-holochain/holochain_runtime/src/launch/config.rs` set three
fields under `dev_mode`:

```rust
network_config.bootstrap_url = Url2::parse("http://127.0.0.1:1");
network_config.relay_url = Url2::parse("https://127.0.0.1:1");
network_config.target_arc_factor = 0;
network_config.advanced = None;
```

`signal_url` is never assigned, and nothing earlier in `conductor_config` sets it, so dev mode
retained the default `wss://dev-test-bootstrap2.holochain.org`. This contradicts the documented
intent — *"This prevents accidental WAN signal/relay usage in dev startup"* (`DEV_MODE.md`) —
where *signal* is named explicitly.

**Why it started now.** `tests/sweetests` is a member of the root workspace (`Cargo.toml:33`), so
it resolves against the root `Cargo.lock`. The `55102122` merge (PR #606 / issue #594) rewrote
that lock, taking holochain `0.6.1` → `0.6.3`. `SweetConductorConfig::standard()` is byte-identical
between the two, so the change lives inside holochain's networking startup, not in our
configuration.

**Three costs.**

1. **Firewall noise.** Every rebuild changes the test binary's code hash, so an outbound-firewall
   rule keyed on that hash re-prompts on each run:
   `… wants to connect to 239.255.255.250 on UDP port 1900 (ssdp) — Code Signature: Not signed`.
2. **Startup latency.** `net_report` probes and relay dialing cost time per conductor spun up, and
   every sweettest creates one.
3. **Correctness of intent.** A single-conductor test has no reason to touch the network at all;
   the dependency is accidental, not designed.

---

No blockers. Baseline is `origin/main` at or after `55102122` (holochain 0.6.3).

Independent of Storage SL2 (#607) in both directions.

Related but deliberately **not** a prerequisite: an upstream switch to disable port mapping (§8)
would retire the vendored patch entirely. Worth filing regardless; not worth waiting for.

---

**How would you solve it?**

**Part 1 — endpoints.** Replace `SweetConductor::from_standard_config()` with a config that
overrides every WAN endpoint, mirroring the host's dev-mode profile rather than inventing a second
convention:

```rust
fn local_only_config() -> SweetConductorConfig {
    SweetConductorConfig::standard().tune_network_config(|network| {
        network.bootstrap_url = url2::url2!("http://127.0.0.1:1");
        network.signal_url = url2::url2!("wss://127.0.0.1:1");
        network.relay_url = url2::url2!("https://127.0.0.1:1");
    })
}
```

Two details that are easy to get wrong:

- **TLS schemes for signal and relay**, not `ws`/`http`. Nothing here sets `signalAllowPlainText`
  or `relayAllowPlainText`, and both transports reject a plaintext URL without the matching flag
  (`kitsune2_transport_iroh` / `_tx5` `validate_config`). An `http://` relay URL fails conductor
  startup outright.
- **Leave `advanced` alone.** Unlike the host's dev mode, it must not be cleared here:
  `SweetConductorConfig::standard()` stores its `k2Gossip` interval tuning in that field, and
  clearing it would silently revert those timings for every sweettest. `target_arc_factor` is
  likewise left alone — with no peers there is nothing to gossip with, so lowering it would change
  DHT arc behaviour without removing any traffic.

`url2` must be added to `tests/sweetests/Cargo.toml`; it is already in the root lock at the same
version as a transitive holochain dependency, so this adds no resolution.

**Part 2 — port mapper.** Vendor `portmapper` 0.12.0 under `patches/` with its three protocol
flags defaulting to `false`, wired via `[patch.crates-io]` in the root `Cargo.toml`. The diff
against upstream is three booleans in `impl Default for Config`.

With all three false no packet is sent, not merely fewer: `Probe::from_output`
(`lib.rs:296,307,319`) gates each probing task on `(enable_x && !x).then(..)` so all three become
`None`, and the mapping selection (`lib.rs:676,695,709`) falls through every branch because the
`upnp`/`pcp`/`pmp` booleans it also tests come from a probe that never ran.

**Part 3 — host dev-mode signal endpoint.** Assign the missing field alongside the existing three,
and extract the block so the invariant is unit-testable without standing up a `FileSystem` and a
keystore:

```rust
fn apply_dev_mode_network_overrides(network_config: &mut NetworkConfig) {
    network_config.bootstrap_url = Url2::parse("http://127.0.0.1:1");
    network_config.signal_url = Url2::parse("wss://127.0.0.1:1");   // was never assigned
    network_config.relay_url = Url2::parse("https://127.0.0.1:1");
    network_config.target_arc_factor = 0;
    network_config.advanced = None;
}
```

The TLS-scheme reasoning from Part 1 applies here too, and here it is load-bearing in a second way:
dev mode *does* clear `advanced`, so a `ws://` signal URL would be rejected outright if the tx5
transport were ever enabled. While in this code, re-confirm that `advanced = None` is still correct
for the iroh transport at 0.6.3 — the only advanced key dev mode would otherwise need is
`irohTransport.relayAllowPlainText`, and that is required only for an `http://` relay URL, which
the `https://` scheme avoids. Update the `DEV_MODE.md` "Network Config → Dev mode" list so it
enumerates every field dev mode actually overrides rather than three of four.

---

**What does this impact?**

In scope:

- `tests/sweetests/src/harness/helpers/mock_conductor.rs` and `tests/sweetests/Cargo.toml`
- `patches/portmapper/` (vendored) and `patches/README.md` (rationale, removal criteria)
- Root `Cargo.toml` — a new `[patch.crates-io]` section
- `host/conductora/plugins/tauri-plugin-holochain/holochain_runtime/src/launch/config.rs` — the
  missing `signal_url` override, extracted into `apply_dev_mode_network_overrides()`
- `host/conductora/plugins/tauri-plugin-holochain/DEV_MODE.md` — the "Network Config" section

Out of scope:

- **Suppressing the host's port-mapping traffic.** Distinct from the `signal_url` fix above, which
  *is* in scope. `[patch.crates-io]` is unconditional, so adding it to `host/Cargo.toml` would
  disable NAT traversal for *production* `conductora` builds, not just dev mode. A per-invocation
  mechanism exists (`cargo --config`) but the rebuild blast radius covers the whole holochain
  stack, which defeats what dev mode is for. Investigated and **rejected** 
- **Host production-mode networking.** The `signal_url` change lives entirely inside the
  `if dev_mode` branch, so normal mode is untouched.
- Multi-conductor / peering topologies. `SweetConductorConfig::rendezvous(bootstrap)` exists for
  tests that genuinely need peers; nothing in this repo needs one today.
- Production network configuration.
- Codesigning the test binaries, and firewall-rule management generally.

The patch reaching only sweettests is a property worth stating: the happ zomes compile to wasm and
never pull iroh, so `tests/sweetests` is the only thing in the root workspace that reaches
`portmapper`.

No behavioural change to any holon, dance, or persistence path.

---

Existing coverage suffices for correctness — this changes how the test conductor is configured, not
what any test asserts. `npm run sweetest` must stay fully green.

New tests worth adding, because the invariant is otherwise unguarded:

- Config-level assertions on `local_only_config()`: every endpoint resolves to `127.0.0.1`; signal
  and relay use TLS schemes; `advanced` still carries the `k2Gossip` tuning from `standard()`.
  These are deterministic, unlike the observational evidence below.
- The patch stays wired: cargo must **not** warn `Patch 'portmapper …' was not used in the crate
  graph`, and the `Cargo.lock` entry for `portmapper` must have **no `source = "registry+…"` line**.
- Host dev-mode overrides, against `apply_dev_mode_network_overrides()`: all three endpoints
  resolve to `127.0.0.1`; signal and relay use TLS schemes; `advanced` is `None`. Worth a third
  case asserting that an incoming `HolochainConfig` already carrying public endpoints is
  overridden rather than inherited — that is the failure mode, not a fresh `NetworkConfig`.

Observational validation (needs a human — firewall dialogs are not visible to tooling):

- **No firewall prompt** on a freshly built test binary.
- **Socket count.** Sample `lsof -nP -i -a -p <pid>` during a run. Expect roughly two UDP sockets
  per conductor — one IPv4, one IPv6 magicsock — and no discovery sockets.
- **Host dev mode:** launch `npm run start:host` with `RUST_LOG=holochain_runtime=debug` and confirm
  no connection attempt to `dev-test-bootstrap2.holochain.org` on the signal path. Then confirm a
  normal launch still reaches UI readiness — this touches a runtime the team uses daily.

Reference measurements from the investigation, same test, same ~21-25s runtime:

| | distinct UDP sockets |
|---|---|
| before | 445 |
| URL overrides only | 17 |
| + portmapper patch | **8** |

Eight is exactly two per conductor across four conductors, i.e. the irreducible QUIC binding.

---

- Sweettests build a conductor whose `bootstrap_url`, `signal_url` and `relay_url` all point at
  unroutable local addresses, using TLS schemes for signal and relay
- `advanced` retains the `k2Gossip` tuning that `SweetConductorConfig::standard()` sets
- `portmapper` is vendored under `patches/` with all three protocol flags defaulting to false, and
  the patch is confirmed wired (no "not used in the crate graph" warning; no `source` line in the
  lock entry)
- `patches/README.md` documents the rationale, the closed alternatives, and the removal criteria
- `host/Cargo.toml` gains **no** `[patch.crates-io]` section
- Host dev mode overrides `signal_url` alongside `bootstrap_url` and `relay_url`, with unit tests
  pinning that all three are local-only and that a config carrying public endpoints is overridden
- `DEV_MODE.md`'s Network Config section lists every field dev mode actually overrides
- The `advanced = None` choice for dev mode is re-confirmed (or corrected) against iroh at 0.6.3
- A normal `npm run start:host` still reaches UI readiness
- A full `npm run sweetest` run stays green with no changes to any test's assertions
- A full run triggers no outbound-firewall prompt on a freshly built binary
- Distinct UDP sockets per run drop to ~2 per conductor, recorded in the PR
- Suite wall-clock time is recorded before and after, and the delta stated in the PR — **including
  if there is no improvement**
- Lockfile drift is limited to the `url2` dependency line and the loss of `source`/`checksum` on
  `portmapper` (the signature of a wired patch)

---

<details>

| Option | Verdict |
|---|---|
| URL overrides alone | **Insufficient.** Removes WAN dialing and most socket churn (445 → 17) but leaves SSDP, because `procure_mapping()` runs before the relay-map check. |
| The `test-utils` disable flags (`disable_bootstrap` / `disable_publish` / `disable_gossip`) | Gate the kitsune modules rather than the transport, so iroh still binds and probes. Not a substitute. |
| Turn off a cargo feature for the port mapper | **Not possible.** `portmapper` is a non-optional dependency of `iroh` (no `optional = true`), and no iroh feature gates it. |
| A runtime config knob | **Not possible.** `IrohTransportConfig` exposes only `relay_url`, `relay_allow_plain_text`, `max_frame_bytes`, `connect_timeout_s` and auth material. Nothing reaches the port mapper. |
| Drop the `transport-iroh` feature from sweettests | **Does not compile.** `error[E0063]: missing field 'transport'` in `kitsune2_api::Builder` (`kitsune2-0.4.1/src/lib.rs:49`) — only one of the two transport cfgs initialises the field, there is no mem-transport feature, and `holochain_p2p/src/spawn/actor.rs:504` calls `kitsune2::default_builder()` with no injection hook. |
| `SweetConductorConfig::rendezvous(false)` | Intended for multi-conductor tests and expects a rendezvous server; heavier than needed for a single in-process conductor. |
| Leave it and add a firewall rule | Solves nothing for CI or other developers. The rule is keyed on the binary's code hash, which changes every build, so it re-prompts anyway. |
| Pin holochain back to `=0.6.1` | Reverts an unrelated upgrade, and the accidental network dependency would still be there. |
| **Upstream a `disable_port_mapping` switch** | **Best long-term fix.** One `Option<portmapper::Config>` on iroh's endpoint builder, or a kitsune2 passthrough, removes the need for the vendored patch. File against `n0-computer/iroh` or `holochain/kitsune2`. |

### 9. Risks or Concerns

**What could go wrong?**

- **Vendoring carries maintenance cost.** ~3050 lines across 15 files to change three booleans.
  On any `portmapper` bump the crate must be **re-vendored, not edited in place**, and the patch
  re-confirmed as wired. `patches/README.md` records this.
- **A silently unused patch.** If `portmapper`'s version moves out from under the patch, cargo warns
  but the build still succeeds — and the SSDP quietly returns. The lockfile assertion in §6 is the
  guard.
- **Someone copies the patch to the host workspace**, quietly weakening production NAT traversal.
  `patches/README.md` calls this out explicitly; §5 records why.
- **Divergence between test and production dependency graphs.** The delta is three booleans in one
  crate's `Default` impl and it affects only test binaries, but it is a real difference.
- **Some sweettest may depend on network startup succeeding** in a non-obvious way. Mitigated by
  the full-suite gate: any such dependency shows up as a failure immediately.
- **The host change touches a runtime the team uses daily.** It is confined to the `if dev_mode`
  branch, but verify a normal `npm run start:host` still reaches UI readiness before merging.
  Pointing `relay_url` at a placeholder while leaving `advanced = None` may also surface new
  startup warnings in place of the current ones — read the run output rather than assuming silence.
- **Wall-clock may not improve.** In the investigation the before/after was within noise, because
  that machine's baseline showed no `net_report` probe timeouts to reclaim. The firewall and
  intent arguments stand on their own; the DoD requires reporting the timing honestly either way.

### 10. Additional Context

**Any supporting material?**

Key source locations:

- `tests/sweetests/src/harness/helpers/mock_conductor.rs` — the `from_standard_config()` call site
- `Cargo.toml:33` — `tests/sweetests` is a root workspace member, so it uses the root lock
- `holochain-0.6.3/src/sweettest/sweet_conductor_config.rs:80` — `SweetConductorConfig::standard()`,
  including the `with_gossip_*` calls that populate `advanced`
- `holochain-0.6.3/src/sweettest/sweet_conductor_config.rs:125` — `tune_network_config()`
- `holochain_conductor_api-0.6.3/src/config/conductor.rs:359-382` — `NetworkConfig::default()`
- `holochain_conductor_api-0.6.3/src/config/conductor.rs:341-356` — the `test-utils` disable flags
- `iroh-holochain-0.95.1/src/magicsock.rs:1803` — hardcoded `Default::default()` port-mapper config
- `iroh-holochain-0.95.1/src/magicsock.rs:1292` / `:1299` — `procure_mapping()` before the
  empty-relay-map early-out
- `portmapper-0.12.0/src/lib.rs:130-139` — the default that enables all three protocols
- `kitsune2_transport_iroh-0.4.1/src/lib.rs:333` — the plaintext-relay rejection
- `kitsune2_transport_tx5-0.4.1/src/lib.rs:179` — the plaintext-signal rejection
- `host/conductora/plugins/tauri-plugin-holochain/holochain_runtime/src/launch/config.rs` — the
  dev-mode network overrides, and the `signal_url` omission
- `host/conductora/src/env.rs` — `dev_mode_enabled()`
- `host/conductora/plugins/tauri-plugin-holochain/DEV_MODE.md` — the Network Config section to
  correct

</details>

nphias:
*1. Sweettests no longer touch the network.** `SweetConductorConfig::standard()` inherits
`NetworkConfig::default()`, which hardcodes the public bootstrap/signal servers and the iroh
canary relay. `local_only_config()` points all three at `127.0.0.1:1`.

Root workspace only — the happ zomes compile to wasm and never pull iroh. Deliberately **not** in
`host/Cargo.toml`: a `[patch]` is unconditional and would disable NAT traversal for production
builds too. Rationale and removal criteria in `patches/README.md`.

**2. Dev mode no longer leaks the public signal server.** `launch/config.rs` overrode
`bootstrap_url` and `relay_url` but never `signal_url`, so dev mode silently kept
`wss://dev-test-bootstrap2.holochain.org`. Now overridden alongside the other two and extracted
into `apply_dev_mode_network_overrides()` so the invariant is unit-testable.

**3. TLS schemes, deliberately.** `signal_url`/`relay_url` use `wss`/`https` rather than
`ws`/`http` in both places, because neither sets `signalAllowPlainText`/`relayAllowPlainText` and
both transports reject a plaintext URL without it. `advanced = None` re-confirmed correct for dev
mode against iroh at 0.6.3. In the sweettest config `advanced` is instead left untouched —
`standard()` stores its `k2Gossip` tuning there. `DEV_MODE.md` updated to list every field dev mode
overrides.

### Testing

- Full sweettest suite green, no changes to any test's assertions
- New config-level tests pin the endpoints, the TLS schemes, and that `k2Gossip` tuning survives;
  `holochain_runtime` gains equivalents for dev mode
- `cargo fmt --check` clean across root, host and sweettests
- Lockfile drift is two lines: `+ url2`, and the loss of `source`/`checksum` on `portmapper`
  (the signature of a wired patch). `host/Cargo.lock` untouched.

### Note

this code rebases off issue 607-storage-sl2..  it should not be merged before 607 (PR 611)


evomimic (approved):
Approve.

This satisfies Issue #612 for me. Sweettests now use an explicit local-only conductor config for all WAN endpoints, while preserving `standard()` gossip tuning; host dev mode now overrides `signal_url` alongside bootstrap and relay; and the vendored `portmapper` patch is tightly scoped to the root workspace rather than host production builds.

I paid particular attention to the patching risk. The vendored `portmapper-0.12.0` source differs from upstream only in the documented `Config::default()` behavior: UPnP, PCP, and NAT-PMP default to disabled. The root-only `[patch.crates-io]` wiring, lockfile signature, and `patches/README.md` removal criteria address the maintenance/blast-radius concern.

CI is passing on PR head `0b83b9a7512e1cda097644baaacc179a22af8aef`. Manual verification reported: `build:happ` nominal, `build:host` nominal, `npm test` nominal, and `npm start` nominal.

Non-blocking polish: the PR body could paste the observed firewall/socket/timing measurements requested by the issue DoD, but I do not see a code blocker.


Thanks for contributing this, @nphias !

Issue #631, opened 2026-08-11
Storage SL5 — Retire Obsolete Persistence Indexes


evomimic:
### NOTE: Some Scope Deferred to Later PR

PR 637 is closing this issue. A new issue #642 has been opened to cover the remaining 631 functionality not delivered as part of PR 637.

### 1. Summary (Required)

Retire the obsolete `AllHolonNodes` and `HolonNodeUpdates` persistence indexes after their consumers have migrated.

Remove their storage writes, readers, link types, integrity-validation branches, test probes, and index-specific tests. Retain exact action retrieval, deletion/detail helpers, and bootstrap path lookup, which serve distinct persistence responsibilities.

---

### 2. Problem Statement (Required)

The current persistence layer retains two index mechanisms that are outside the intended storage algebra:

- `AllHolonNodes` provides whole-space lineage discovery.
- `HolonNodeUpdates` was intended as a revision index, but native root-addressed Holochain updates are now authoritative and no write path creates those links.

Repository grounding confirms both are transitional:

- `persistence_layer/holon_storage.rs` writes an `AllHolonNodes` link for `PublishRoot`.
- `persistence_layer/all_holon_nodes.rs` reads that index and calls `get_latest_holon_node`, but the latter cannot find a newer version because no `HolonNodeUpdates` links are authored.
- `persistence_layer/holon_node.rs` still exposes scaffolded `get_all_revisions_for_holon_node` and `get_latest_holon_node` externs that traverse `HolonNodeUpdates`.
- The integrity zome still declares both link types and routes their create/delete validation.
- `AllHolonNodes` links remain the implementation behind `GuestHolonService::get_all_holons_internal`.

The resulting whole-space read is not a valid latest-version API: it indexes roots, and its current latest-version helper returns the root after version-producing updates. Keeping these indexes also retains a second, unused revision-topology model beside Holochain’s native update graph.

---

### 3. Dependencies (Required)

- Depends on Storage SL2: Version-Aware Holon Persistence.
- Depends on there being no remaining global get-all consumer.

Current blocking prerequisite:

- `GuestHolonService::get_all_holons_internal` still reads `AllHolonNodes`.
- `TransactionAction::GetAllHolons` is exercised by Sweettest command, database-count, descriptor-verification, lookup, and print flows.

The coordinator-owned replacement for whole-space discovery is outside the Storage plan. It must land, and all callers must adopt it, before this retirement issue can remove the index.

[...]

PR #657, opened 2026-08-23
642 Storage SL5b allholonNodes fix

### LINKED ISSUE #642: Storage SL5b — Replace AllHolonNodes with HolonSpace ownership
# Storage SL5b — Replace AllHolonNodes with HolonSpace ownership

### 1. Summary (Required)

Replace the legacy `AllHolonNodes` global persistence index with ordinary `OwnedBy` / `Owns` membership in the current `HolonSpace`.

New holon lineages acquire `OwnedBy → current HolonSpace` from their bound `TransactionContext`; ordinary commit materializes the inverse `HolonSpace ─Owns→ Holon` SmartLink. The existing `GetAllHolons` command and SDK API retain their current contract while their guest implementation moves from the global index to the current space's `Owns` relationship.

Once the compatibility retrieval path is migrated, remove the `AllHolonNodes` storage write, reader, link type, validation, probes, and extern surface.

The `LocalHolonSpace` anchor is not a member of its own `Owns` collection. `GetAllHolons` returns ordinary holon lineages owned by the current space; callers needing the anchor use `TransactionContext::get_space_holon()`.

---

`AllHolonNodes` is a storage-owned global-discovery index. It duplicates a semantic topology already defined in the MAP Core Schema:

    Holon --OwnedBy--> HolonSpace
    HolonSpace --Owns--> Holon

The schema already defines `HolonSpace`, `OwnedBy`, and its `Owns` inverse; `OwnedBy` is an `InstanceRelationship` of `HolonType`. The missing behavior is runtime materialization and lifecycle policy, not a new query model.

Today:

- `MutationFacade::new_holon` creates a transient holon without `OwnedBy`.
- `GuestHolonService::get_all_holons_internal` reads `get_all_holon_ids()` from `AllHolonNodes`.
- Guest commit processing already persists a declared relationship and its descriptor-resolved inverse as ordinary SmartLinks.
- Version-producing commits replay their full relationship scope, so membership must be protected from becoming one link per historical version.
- Existing count fixtures include the local-space bootstrap holon because it is persisted; this issue intentionally changes `GetAllHolons` to return owned members rather than the space anchor.

The complete existing retrieval path is:

    SDK getAllHolons()
      → TransactionActionWire::GetAllHolons
      → TransactionAction::GetAllHolons
      → map_commands_runtime::transaction_handler
      → context.lookup().get_all_holons()
      → HolonServiceApi::get_all_holons_internal(context)
      → GuestHolonService::get_all_holons_internal(context)
      → get_all_holon_ids() through AllHolonNodes

SL5b preserves that public path and changes only the guest retrieval implementation:

    GuestHolonService::get_all_holons_internal(context)
      → context.get_space_holon()
      → space.related_holons(CoreRelationshipTypeName::Owns)
      → GuestHolonService::fetch_related_holons_internal(...)
      → persistence_layer::expand_from_source(...)

---

- Depends on PR #637 / Storage SL5a landing: `HolonNodeUpdates` must be retired first.
- Depends on the bootstrap core schema loading `HolonSpace`, `OwnedBy`, and `Owns`.
- Does not depend on full QueryExpression or QueryDance implementation. Future query work may express the same operation as `Expand(Owns)`.
- Existing command, wire, SDK, loader, and Sweettest consumers are migration targets, not additional dependencies.

---

1. Define the runtime invariant: `OwnedBy` represents a holon lineage's membership in the current `HolonSpace`, not membership of every persisted version.

2. Add context-derived ownership when a new lineage is created or independently derived:

   - `MutationFacade::new_holon`;
   - `stage_new_holon` where a transient bypasses ordinary creation;
   - `stage_new_from_clone`, rebinding inherited ownership to the current context's space.

   The `HolonSpace` must be inferred through `TransactionContext::get_space_holon()`; do not add a space parameter to ordinary creation APIs.

3. Preserve ownership across `stage_new_version` without re-emitting it on a version-producing commit. A new version must not add another `Owns` member for the same lineage.

4. Keep local-space bootstrap exempt from ordinary ownership creation. The local space remains the transaction context anchor and is not self-owned.

5. Use ordinary relationship commit behavior to persist `OwnedBy` and its `Owns` inverse; do not add a new storage index or direct DHT access path.

6. Migrate the complete legacy `GetAllHolons` retrieval path, not only its guest leaf:

   - retain `TransactionAction::GetAllHolons`, `LookupFacade::get_all_holons()`, host dispatch, wire binding, and SDK `getAllHolons()` as compatibility surfaces;
   - replace only `GuestHolonService::get_all_holons_internal`'s `AllHolonNodes` lookup with current-space `Owns` traversal;
   - use the established Reference Layer `related_holons(Owns)` path and existing SmartLink expansion;
   - do not introduce a new command, wire type, SDK endpoint, or query API.

7. Update count and retrieval fixtures for the explicit contract: `GetAllHolons` returns owned members and excludes the `LocalHolonSpace` anchor.

8. Remove the complete `AllHolonNodes` surface:

   - `index_under_all_holon_nodes` and the `PublishRoot` write;
   - `persistence_layer/all_holon_nodes.rs`, `get_all_holon_nodes`, and `get_all_holon_ids`;
   - `ALL_HOLON_NODES_PATH`, `LinkTypes::AllHolonNodes`, validators, tests, probes, and coordinator-surface exports;
   - test-harness liveness calls that invoke `get_all_holon_nodes`.

9. Ensure deletion does not leave discoverable stale membership. A deleted holon must not remain in `Owns` retrieval as an unresolved member.

---

Expected creation, staging, and ownership areas:

- `shared_crates/holons_core/src/core_shared_objects/transactions/mutation_facade.rs`
- `shared_crates/holons_core/src/core_shared_objects/transactions/transaction_context.rs`
- staging and clone lifecycle code under `shared_crates/holons_core/src/core_shared_objects/`
- `shared_crates/holons_core/src/reference_layer/holon_service_api.rs`
- `shared_crates/holons_core/src/core_shared_objects/transactions/lookup_facade.rs`

Expected guest storage and retrieval areas:

- `happ/crates/holons_guest/src/guest_shared_objects/commit_functions.rs`
- `happ/crates/holons_guest/src/guest_shared_objects/guest_holon_service.rs`
- `happ/crates/holons_guest/src/persistence_layer/holon_storage.rs`
- `happ/crates/holons_guest/src/persistence_layer/holon_node.rs`
- `happ/crates/holons_guest/src/persistence_layer/smartlink.rs`
- `happ/crates/holons_guest/src/persistence_layer/all_holon_nodes.rs` — remove
- integrity link declarations and infrastructure validation
- coordinator surface/probes and Sweettest harnesses

Expected compatibility-path areas:

- `host/crates/map_commands_runtime/src/transaction_handler.rs`
- `host/crates/map_commands_contract/src/transaction_command.rs`
- `host/crates/map_commands_wire/src/transaction_wire.rs`
- `host/map-sdk/src/internal/commands/transaction.ts`
- host/client HolonService implementations and command fixtures that exercise `GetAllHolons`
- loader, count, lookup, descriptor-verification, and print-flow tests.

Out of scope:

- QueryExpression, QueryDance, query planning, filtering, ordering, pagination, or distributed query execution.
- A new public whole-space query API.
- Key/type indexes or optimization of consumers that currently scan `GetAllHolons`.
- Head selection or version-history traversal.
- Changes to canonical core-schema JSON inputs.

---

Add or update tests for:

- a new holon receives context-inferred `OwnedBy` before commit;
- commit produces the inverse `Owns` member on the local space;
- an independent clone is owned by the current space, not an inherited source space;
- a version-producing update does not create an additional ownership member;
- `GuestHolonService::get_all_holons_internal` resolves the current space and traverses `Owns`;
- `TransactionAction::GetAllHolons`, command dispatch, wire binding, and SDK `getAllHolons()` retain their current request/result contracts while returning the ownership-backed collection;
- loader, count, key-lookup, descriptor-verification, and print consumers continue to work through the unchanged command surface;
- `GetAllHolons` returns the expected owned lineage members and excludes the local-space anchor;
- deleting a holon does not leave it discoverable through `Owns`;
- bootstrap behavior remains valid;
- exact stored-holon reads, SmartLink expansion, and `LocalHolonSpace` behavior remain intact;
- workspace search finds no live `AllHolonNodes`, `ALL_HOLON_NODES_PATH`, `get_all_holon_nodes`, or `get_all_holon_ids` production usage.

Run relevant unit tests, `npm run check`, `npm run test:unit`, and applicable Sweettest flows.

---

- [ ] New and independently cloned lineages acquire `OwnedBy` from their transaction context without a caller-supplied space parameter.
- [ ] Ordinary commit persists the matching `Owns` inverse.
- [ ] Version-producing updates do not create duplicate historical membership.
- [ ] The local-space bootstrap anchor is not self-owned.
- [ ] `GuestHolonService::get_all_holons_internal` retrieves owned members through `context.get_space_holon()` and `related_holons(Owns)`.
- [ ] `TransactionAction::GetAllHolons`, LookupFacade, host runtime dispatch, wire contract, and SDK `getAllHolons()` remain unchanged and pass end-to-end tests.
- [ ] All live consumers continue to pass through the existing command/SDK surface.
- [ ] Deleted holons are not returned as live owned members.
- [ ] `AllHolonNodes` writes, reads, module, externs, link type, validation, constants, probes, and tests are removed.
- [ ] No general query implementation or replacement storage index is introduced.
- [ ] Relevant checks and integration suites pass.

<details>

Keep `AllHolonNodes` until full query implementation. Rejected: the existing `Owns` topology and relationship traversal provide the required semantic replacement now.

Add another storage enumeration API or index. Rejected: it repeats the storage/coordinator boundary violation SL5 exists to remove.

Implement full QueryExpression / `Expand` first. Deferred: it should eventually consume `Owns`, but is larger and depends on incomplete query-plan realignment.

Re-anchor membership on each version. Rejected: every version-producing update would need to remove the predecessor's `OwnedBy` / `Owns` pair and write a new pair for the successor. That creates avoidable link churn, partial-commit recovery concerns, and branch/merge ambiguity. Membership is lineage-scoped; exact-version selection remains separate storage/query work.

Include the `LocalHolonSpace` anchor through a self-`OwnedBy` / `Owns` relationship. Rejected: it makes the container its own member, introduces a self-edge into future traversal semantics, and complicates bootstrap and deletion. The anchor remains directly available from `TransactionContext`.

### 9. Risks or Concerns

This issue deliberately changes legacy count semantics: the persisted `LocalHolonSpace` anchor is no longer included in `GetAllHolons`. Count fixtures and callers must distinguish ordinary owned members from the infrastructure anchor.

Relationship cleanup on deletion must be explicit enough that an `Owns` expansion never returns a stale, unresolved member.

The lookup-by-key consumers continue to scan the returned collection in this slice. That preserves their current behavior but is not a future query/indexing strategy.

### 10. Additional Context

Sources inspected:

- `docs/roadmap/desc-driven-impl-plan.md`
- `docs/core/guest/storage-layer-services/storage-layer-design-spec.md`
- `docs/core/guest/storage-layer-services/storage-layer-impl-plan.md`
- transaction creation/staging, Reference Layer lookup, guest commit/retrieval/storage, integrity, command/wire/SDK, and Sweettest surfaces.

</details>

nphias:
## Summary

Whole-space discovery becomes ordinary graph traversal. Publishing a lineage writes
`Holon ─OwnedBy→ HolonSpace` and its `HolonSpace ─Owns→ Holon` inverse; `get_all_holons_internal`
reads that `Owns` collection instead of the `AllHolonNodes` path index. This is the
coordinator-owned replacement #631 §3 and §9 named as SL5b's precondition.

`GetAllHolons`, the legacy dance, the wire binding, and the TS SDK keep their contracts.

**One semantic change:** the `LocalHolonSpace` anchor is no longer in `GetAllHolons` results — it is
not a member of its own `Owns` collection (#642 §1, §9), so every DB count drops by one. Callers
that want the anchor use `TransactionContext::get_space_holon()`.

`AllHolonNodes` is still written and all of its tests still pass. Nothing reads it, which makes this
a reversible checkpoint rather than a cutover.

## Changes

**`commit_functions.rs`** — new `persist_space_membership`, called from `commit_holon`'s `ForCreate`
arm immediately after `persist_holon` returns. It writes the `OwnedBy` and `Owns` SmartLinks
directly through the existing `persist_smartlink`, in the node pass.

Membership is deliberately *not* staged for the relationship pass. See the first note below — this
is the load-bearing decision in the PR.

**`staged.rs`** — `relationship_collections_for_commit` excludes `OwnedBy` from every scope. Storage
authors membership; the coordinator never replays it. A staged `OwnedBy` can still arrive (cloning a
saved holon carries its persisted membership into the staged map) and is dropped rather than
re-anchored to the clone or to a new version.

**`guest_holon_service.rs`** — `get_all_holons_internal` resolves the space and delegates to the
existing `fetch_related_holons_internal` with `Owns`; no new traversal code. `delete_holon_internal`
gains `retract_space_membership`, deleting the `Owns` link on the space and the `OwnedBy` link on the
holon before the node delete. `create_local_space_holon` gains a comment recording that its staging
bypass is now load-bearing.

Reading through `fetch_related_holons_internal` is a small win: the old code built its collection
with `add_references` and no keys, whereas the `Owns` link's canonical key *is* the owned holon's
key — so the collection is now keyed from tag bytes with no hydration, which is what every
`get_by_key` consumer behind this call already wanted.

**Contract docs** — `holon_service_api.rs`, `holon_dance_adapter.rs`, `get_all_holons_dance.rs`,
`transaction_command.rs`, `transaction_wire.rs`, `map-sdk/.../transaction.ts`, and
`schema-src/commands/schema.tdl` (with the generated JSON regenerated — a one-line diff). All said
"all persisted holons".

**Fixtures** — `count_saved()` drops the `+ 1` for the anchor, carrying eight fixtures at once;
`load_holons_internal_fixture`'s five hardcoded `MapInteger(1)` baselines become `MapInteger(0)`;
`ENSURE_DB_EMPTY` and the inline "+1 for the space holon" comments corrected. New
`partial_commit_membership_fixture` — see Testing.

## Notable decisions

- **Membership is authored in the node pass, beside `persist_holon`.** This is where the design
  landed after two earlier attempts, and the reason matters. Membership is what whole-space
  discovery reads, so a holon that publishes without it is persisted and permanently invisible —
  something `AllHolonNodes` could not produce, because it indexed inside `PublishRoot`.

  Staging the edge for the relationship pass reintroduces exactly that split. The relationship pass
  is skipped outright whenever *any* staged holon fails to publish, so a commit of two holons where
  one is rejected leaves the other published and unowned, with nothing to repair it later. Ordering
  `OwnedBy` first within the pass — which an earlier revision of this PR did — narrows the window
  but does not close it. Writing membership beside the node write restores the coupling
  unconditionally.

- **The coordinator never replays membership.** Since storage authors it, replaying a staged copy
  can only do harm: a version-producing commit replays the full scope against the *new* version's
  id, which would make the space own the lineage root and every version after it separately (#642
  §8 rejects re-anchoring per version). Excluding `OwnedBy` from every scope is one filter and
  covers the version case, the clone case, and any future path that stages it.

-`resolve_inverse_relationship_name` never sees `OwnedBy`, so the descriptor
  short-circuit added for it is gone and the descriptor system is back to having no exceptions.

- **Deletion retracts membership both ways.** `delete_holon_node` deletes only the path link and the
  entry; SmartLinks survive it, which would leave deleted holons discoverable (#642 §9). The
  retraction lives in `delete_holon_internal`, not the persistence layer, which is deliberately
  descriptor-unaware.

- **A latent harness bug surfaced.** `FixtureHolons::counts()` charged a deleted holon `saved -= 1`
  *in addition to* its head no longer counting as `Saved`. The old anchor allowance cancelled it
  exactly, which is why it went unnoticed and why `delete_holon_fixture`'s post-delete assertion was
  commented out with a TODO. `Deleted` now contributes nothing and that assertion is enabled — it is
  how the retraction above is tested.

- **The saved-content comparator ignores `OwnedBy`.** `assert_saved_content_eq` refused to compare an
  undescribed actual holon carrying any relationship content, and every saved holon now carries
  membership. Infrastructure-supplied edges never appear in fixture snapshots — the same reasoning
  that already tolerates commit-materialized inverse SmartLinks there.

- `all_holon_nodes.rs` and its `mod.rs` wiring; `index_under_all_holon_nodes` and its `PublishRoot`
call; `ALL_HOLON_NODES_PATH`; `LinkTypes::AllHolonNodes` and its four dispatch arms; both validators,
the `AllHolonNodesDelete` rejection, and their re-exports; the `all_holon_nodes_delete_for_test`
probe and the `RootIndexLinkType::AllHolonNodes` variant; both `coordinator-surface.toml`
`[[export]]` blocks; and the `AllHolonNodes` tests and table legs. `mock_conductor.rs`'s liveness
probe calls the `get_all_holon_nodes` extern purely as a callability check and needs repointing.

- `validate_root_index_create` and the shared rejection variants stay — `LocalHolonSpace` still uses
them (#631 §9). Removing the enum variant renumbers `LocalHolonSpace` 1→0 and `SmartLink` 2→1 and so
changes the DNA hash: the same accepted consequence as SL5a, inert by construction since the enum is
the single source of truth for both the write and the read of every link.
## Testing

**The write and retract paths are both directly exercised**, which matters because discovery now
depends entirely on them:

- *Create.* Every `ensure_database_count` step pins membership exactly, now that the anchor
  allowance is gone: pre-commit steps assert the anchor is excluded, post-commit steps assert new and
  cloned lineages appear, and `stage_new_version_fixture` asserts a version adds no member.
- *Partial commit.* `partial_commit_membership_fixture` (new) stages two holons, one carrying more
  properties than PVL permits, so its `PublishRoot` is rejected and the commit reports `Incomplete`
  before the relationship pass runs. It then asserts the other holon is still an owned member. This
  is the case the earlier staged-membership design got wrong; the fixture was confirmed to fail
  against that design and pass against this one.
- *Delete.* `delete_holon_fixture`'s post-delete count assertion, re-enabled here, asserts the
  deleted holon is gone from `GetAllHolons` — the retraction's only test, and the reason the harness
  counting bug above had to be fixed.

**Not measured: the `Owns` fan-in cost.** Every inverse `Owns` link is based on the space holon, and
`put_smartlink` scans all live links on the base for conflict detection — so a 191-holon core-schema
load is ~O(n²) tag decodes against one base. Numbers on this branch: `dance_tests` 144.9s,
`holon_storage_tests` 68.8s, `pvl_validation_tests` 47.9s, `smartlink_tests` 19.1s. A before/after
comparison still needs a baseline run on the parent commit. This is the open SL3 candidate-scan cost,
now on a hot path; conflict-detection semantics were deliberately left alone.

**Environment note.** The three workspaces share one `target/`; running `cargo` outside
`nix develop` mixes rustc versions and the next nix-side build fails with `E0514` plus misleading
`cannot find type X` errors in untouched files. Also, `npm run sweet:test` skips the WASM rebuild and
the audit.

## Scope

Unchanged: `QueryExpression`, `QueryDance`, query planning (#642 §5); any new public whole-space
query API; key or type indexes; the core-schema JSON inputs beyond one regenerated command
description; `persistence_layer::smartlink` and the Tag v1 codec; `LocalHolonSpace` bootstrap
behavior; exact-version reads and Storage SL2 record behavior.

`OwnedBy` and `Owns` still set neither `IsDefinitional` nor `AllowsDuplicates` in the TDL. Nothing
reads them — storage authors membership directly and the coordinator never resolves it through
descriptor policy — so it is not blocking, but it is a trap for anyone who later routes ownership
through that policy.

Still open, unaffected: the `put_smartlink` candidate-scan cost (SL3), now more load-bearing than
before, and the reference-layer/storage disagreement over what "the same link" means (SL4).

### DIFF
--- a/generated/json-imports/commands/schema.json
+++ b/generated/json-imports/commands/schema.json
@@ -896,7 +896,7 @@
         "TypeNamePlural": "GetAllHolonsCommands",
         "DisplayName": "Get All Holons Command",
         "DisplayNamePlural": "Get All Holons Commands",
-        "Description": "Returns all holons visible in the active transaction."
+        "Description": "Returns the holons owned by the active transaction's holon space."
       },
       "relationships": [
         {
--- a/happ/crates/holons_guest/src/guest_shared_objects/commit_functions.rs
+++ b/happ/crates/holons_guest/src/guest_shared_objects/commit_functions.rs
@@ -436,7 +436,6 @@ fn commit_holon(
             StagedState::ForCreate => {
                 trace!("StagedState::ForCreate — publishing a new HolonNode lineage");
                 staged_holon.prepare_full_relationship_commit_scope()?;
-                stage_space_membership(staged_holon, context)?;
                 let stored = persist_holon(HolonWriteRequest::PublishRoot {
                     holon_node: staged_holon.into_node_model(),
                 })?;
@@ -446,7 +445,10 @@ fn commit_holon(
                     short_hex(&stored.version_metadata.version_id, 8)
                 );
 
-                staged_holon.to_committed(stored.version_metadata.version_id)?;
+                let new_local_id = stored.version_metadata.version_id;
+                persist_space_membership(context, &new_local_id, staged_holon.key())?;
+
+                staged_holon.to_committed(new_local_id)?;
                 Ok(CommitOutcome::Saved)
             }
 
@@ -468,10 +470,7 @@ fn commit_holon(
             StagedState::ForUpdateNewVersion => {
                 trace!("StagedState::ForUpdateNewVersion — publishing the next HolonNode version");
                 let predecessor_id = staged_holon.get_versioned_source_id()?;
-                // Full replay, except space membership: the lineage already belongs to the
-                // space, and re-emitting `OwnedBy` here would make the space `Owns` both the
-                // lineage root and every version published after it.
-                staged_holon.prepare_new_version_relationship_commit_scope()?;
+                staged_holon.prepare_full_relationship_commit_scope()?;
                 // Storage decides how a version is written: it resolves the lineage this
                 // predecessor belongs to and roots the new version there. Immediate-predecessor
                 // ordering is not carried by the node — it is staged below as a SmartLink.
@@ -526,57 +525,57 @@ fn commit_holon(
     }
 }
 
-/// Stamps `OwnedBy → current HolonSpace` on a lineage-creating commit.
+/// Records `Holon ─OwnedBy→ HolonSpace` and its `Owns` inverse for a newly published lineage.
 ///
-/// Space membership is infrastructure-supplied, and this is the one place it can be supplied
-/// reliably. Staging can happen in a client transaction that has not yet been told which space
-/// it is bound to, whereas by the time a commit reaches the guest the local space anchor has
-/// always been resolved. Stamping here also keeps membership and publication together — the same
-/// coupling the global index it replaces got from being written during `PublishRoot`.
+/// Written here, in the node pass, rather than staged for the relationship pass. Membership is
+/// what whole-space discovery reads, so it has to hold whenever the node itself was published —
+/// and the relationship pass is skipped entirely when *any* staged holon fails to publish, which
+/// would otherwise leave the holons that did publish permanently undiscoverable. Authoring it
+/// beside `persist_holon` gives membership the same coupling to publication that
+/// `index_under_all_holon_nodes` had.
 ///
-/// Only lineage-creating commits stamp: a new version inherits the membership its lineage root
-/// already holds, and re-emitting it would make the space own every version separately.
+/// Only lineage-creating commits call this. A new version inherits the membership its lineage
+/// root already holds; re-emitting it would make the space own the root and every version after
+/// it separately.
 ///
-/// Idempotent, so a lineage that was already stamped upstream keeps its existing owner.
-fn stage_space_membership(
-    staged_holon: &mut StagedHolon,
+/// Both writes are idempotent (`put_smartlink` reports `AlreadyPresent`), so re-publishing a
+/// lineage that is already a member is a no-op.
+fn persist_space_membership(
     context: &Arc<TransactionContext>,
+    source_id: &LocalId,
+    source_key: Option<MapString>,
 ) -> Result<(), HolonError> {
-    let owned_by = CoreRelationshipTypeName::OwnedBy.as_relationship_name();
-
-    let already_owned = {
-        let collection_rc = staged_holon.get_staged_relationship(&owned_by)?;
-        let collection = collection_rc.read().map_err(|e| {
-            HolonError::FailedToAcquireLock(format!(
-                "Failed to acquire read lock on staged OwnedBy collection: {}",
-                e
-            ))
-        })?;
-        !collection.get_members().is_empty()
-    };
-    if already_owned {
-        return Ok(());
-    }
-
     // No space means nothing to belong to. Reachable only while the local space anchor is itself
-    // being established, which is the one lineage that must not be a member of its own space.
+    // being established — the one lineage that must not be a member of its own space.
     let Some(space_reference) = context.get_space_holon()? else {
         return Ok(());
     };
 
-    let space_key = space_reference.key()?.ok_or_else(|| {
-        HolonError::EmptyField("HolonSpace holon has no key; cannot record membership".to_string())
+    // Membership in a non-local space would need cross-space inverse propagation, which belongs
+    // at the outbound-proxy seam rather than here (the same boundary `require_local_target` draws).
+    let HolonId::Local(space_id) = space_reference.holon_id()? else {
+        return Ok(());
+    };
+
+    persist_smartlink(PreparedSmartLink {
+        source_id: source_id.clone(),
+        target_id: HolonId::Local(space_id.clone()),
+        relationship_name: CoreRelationshipTypeName::OwnedBy.as_relationship_name(),
+        canonical_key: canonical_key_from_optional_key(space_reference.key()?)?,
+        occurrence_id: None,
+        relationship_property_values: PropertyMap::new(),
+        target_property_cache_candidates: Vec::new(),
     })?;
 
-    // Carry the key on the reference: relationship commit fail-fasts on a target whose key
-    // cannot be resolved, and this answers it without re-fetching the space per staged holon.
-    let space_target = HolonReference::smart_with_key(
-        TransactionContextHandle::new(Arc::clone(context)),
-        space_reference.holon_id()?,
-        space_key.clone(),
-    );
-
-    staged_holon.add_related_holons_with_keys(owned_by, vec![(space_target, Some(space_key))])?;
+    persist_smartlink(PreparedSmartLink {
+        source_id: space_id,
+        target_id: HolonId::Local(source_id.clone()),
+        relationship_name: CoreRelationshipTypeName::Owns.as_relationship_name(),
+        canonical_key: canonical_key_from_optional_key(source_key)?,
+        occurrence_id: None,
+        relationship_property_values: PropertyMap::new(),
+        target_property_cache_candidates: Vec::new(),
+    })?;
 
     Ok(())
 }
--- a/happ/crates/holons_guest/src/guest_shared_objects/commit_functions.rs
+++ b/happ/crates/holons_guest/src/guest_shared_objects/commit_functions.rs
@@ -436,6 +436,7 @@ fn commit_holon(
             StagedState::ForCreate => {
                 trace!("StagedState::ForCreate — publishing a new HolonNode lineage");
                 staged_holon.prepare_full_relationship_commit_scope()?;
+                stage_space_membership(staged_holon, context)?;
                 let stored = persist_holon(HolonWriteRequest::PublishRoot {
                     holon_node: staged_holon.into_node_model(),
                 })?;
@@ -467,7 +468,10 @@ fn commit_holon(
             StagedState::ForUpdateNewVersion => {
                 trace!("StagedState::ForUpdateNewVersion — publishing the next HolonNode version");
                 let predecessor_id = staged_holon.get_versioned_source_id()?;
-                staged_holon.prepare_full_relationship_commit_scope()?;
+                // Full replay, except space membership: the lineage already belongs to the
+                // space, and re-emitting `OwnedBy` here would make the space `Owns` both the
+                // lineage root and every version published after it.
+                staged_holon.prepare_new_version_relationship_commit_scope()?;
                 // Storage decides how a version is written: it resolves the lineage this
                 // predecessor belongs to and roots the new version there. Immediate-predecessor
                 // ordering is not carried by the node — it is staged below as a SmartLink.
@@ -520,6 +524,61 @@ fn commit_holon(
             holon_write
         )))
     }
+}
+
+/// Stamps `OwnedBy → current HolonSpace` on a lineage-creating commit.
+///
+/// Space membership is infrastructure-supplied, and this is the one place it can be supplied
+/// reliably. Staging can happen in a client transaction that has not yet been told which space
+/// it is bound to, whereas by the time a commit reaches the guest the local space anchor has
+/// always been resolved. Stamping here also keeps membership and publication together — the same
+/// coupling the global index it replaces got from being written during `PublishRoot`.
+///
+/// Only lineage-creating commits stamp: a new version inherits the membership its lineage root
+/// already holds, and re-emitting it would make the space own every version separately.
+///
+/// Idempotent, so a lineage that was already stamped upstream keeps its existing owner.
+fn stage_space_membership(
+    staged_holon: &mut StagedHolon,
+    context: &Arc<TransactionContext>,
+) -> Result<(), HolonError> {
+    let owned_by = CoreRelationshipTypeName::OwnedBy.as_relationship_name();
+
+    let already_owned = {
+        let collection_rc = staged_holon.get_staged_relationship(&owned_by)?;
+        let collection = collection_rc.read().map_err(|e| {
+            HolonError::FailedToAcquireLock(format!(
+                "Failed to acquire read lock on staged OwnedBy collection: {}",
+                e
+            ))
+        })?;
+        !collection.get_members().is_empty()
+    };
+    if already_owned {
+        return Ok(());
+    }
+
+    // No space means nothing to belong to. Reachable only while the local space anchor is itself
+    // being established, which is the one lineage that must not be a member of its own space.
+    let Some(space_reference) = context.get_space_holon()? else {
+        return Ok(());
+    };
+
+    let space_key = space_reference.key()?.ok_or_else(|| {
+        HolonError::EmptyField("HolonSpace holon has no key; cannot record membership".to_string())
+    })?;
+
+    // Carry the key on the reference: relationship commit fail-fasts on a target whose key
+    // cannot be resolved, and this answers it without re-fetching the space per staged holon.
+    let space_target = HolonReference::smart_with_key(
+        TransactionContextHandle::new(Arc::clone(context)),
+        space_reference.holon_id()?,
+        space_key.clone(),
+    );
+
+    staged_holon.add_related_holons_with_keys(owned_by, vec![(space_target, Some(space_key))])?;
+
+    Ok(())
 }
 
 fn stage_predecessor_relationship(
--- a/happ/crates/holons_guest/src/guest_shared_objects/commit_functions.rs
+++ b/happ/crates/holons_guest/src/guest_shared_objects/commit_functions.rs
@@ -545,9 +545,16 @@ fn persist_space_membership(
     source_id: &LocalId,
     source_key: Option<MapString>,
 ) -> Result<(), HolonError> {
-    // No space means nothing to belong to. Reachable only while the local space anchor is itself
-    // being established — the one lineage that must not be a member of its own space.
+    // Every guest dance context is anchored to a persisted space before its body runs
+    // (`dance_adapter::init_guest_context` calls `ensure_local_holon_space`), and the anchor itself
+    // never reaches commit, so this branch should be unreachable. Skipping quietly would leave a
+    // published holon permanently undiscoverable with nothing to show for it, so say so.
     let Some(space_reference) = context.get_space_holon()? else {
+        warn!(
+            "No space holon bound to this transaction; {} was published without space membership \
+             and will not appear in whole-space discovery",
+            short_hex(source_id, 8)
+        );
         return Ok(());
     };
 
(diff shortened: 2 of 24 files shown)


evomimic:
# Comments on Draft PR

I know you have not submitted this PR for review yet, but I wanted to provide feedback now because I think the implementation is drifting into unnecessary complexity.

## On the asserted need to populate `OwnedBy` / `Owns` directly during pass 1

**Status: Disagree.**

I do not think the asserted gap exists under the current transaction lifecycle.

A pass-1 failure produces an `Incomplete` commit response. It does not complete the transaction or clear its nursery. A holon successfully persisted in pass 1 is retained as a staged holon in `Committed(saved_id)` state, so its persisted identity remains available through the transaction’s staged references.

The relevant sequence is:

    1. Pass 1 saves holon A and records its LocalId on the staged holon.
    2. Pass 1 fails for holon B.
    3. The commit response is Incomplete; pass 2 is skipped.
    4. The transaction remains open and its nursery remains intact.
    5. The caller repairs or abandons B, then commits again.
    6. The subsequent complete commit processes ordinary relationships for A using its retained committed LocalId.
    7. Normal relationship persistence writes A --OwnedBy--> Space and materializes Space --Owns--> A.

Therefore, A is not permanently invisible and does not need a special pass-1 persistence path. Before transaction completion, it is discoverable through the nursery/staged-reference surface, which is the appropriate recovery surface for an incomplete transaction. Whole-space discovery is a finalized persisted-membership view, not the recovery mechanism for partially completed transactions.

### Proposed ownership implementation boundary

I think the required change is deliberately small: implement context-derived space membership entirely in the nursery’s `HolonStagingBehavior`, and keep commit relationship persistence unchanged.

#### `stage_new_holon`

- Require the current `HolonSpace` from the transaction’s space manager.
- Construct a bound `HolonReference` for that space.
- Add `OwnedBy → current HolonSpace` to the new staged lineage.
- If the transaction cannot provide its current space, return an explicit error rather than silently creating an unowned holon.

`LocalHolonSpace` bootstrap remains the explicit exception: it is created through its existing bootstrap path and is not staged as a member of itself.

#### `stage_new_from_clone`

- Clone the ordinary relationship state.
- Inspect the inherited `OwnedBy` relationship.
- If it names a different space, replace it with `OwnedBy → current HolonSpace`.
- If it already names the current space, retain one canonical membership.

An independent clone is a new lineage, so its ownership must be anchored to the current transaction space rather than the source lineage’s space.

#### `stage_new_version`

- Clone the ordinary relationship state.
- Remove `OwnedBy` from the cloned relationship map.

`OwnedBy` belongs only to the lineage root. A version’s owner is resolved by tracing its lineage to the root and following that root’s `OwnedBy`; copying the relationship onto the version would be misleading and risks re-persisting it against the new version ID.

### Commit consequence

Pass 2 remains ordinary relationship persistence:

- New lineages and independently cloned lineages already have staged `OwnedBy`, so normal relationship commit persists it and materializes `Owns`.
- New versions have no staged `OwnedBy`, so there is nothing to suppress or specially filter.
- No direct `persist_space_membership` path, hand-authored inverse SmartLinks, or global `OwnedBy` commit filter is needed.

An incomplete commit remains an open transaction with a retained nursery. Successfully persisted staged holons and their pending relationship work remain available until a later complete commit or rollback; this is not a reason to make membership a special pass-1 persistence action.

The partial-commit regression test should be changed accordingly: assert that the successful holon remains addressable through its staged reference after `Incomplete`, then assert membership appears after the follow-up complete commit.

## Issue #642 contract departures in PR #657

The retrieval direction is correct: `GetAllHolons` traversing the active `HolonSpace`’s `Owns` relationship, excluding the space anchor, is the intended replacement path. The following points are departures from the issue’s specified design or Definition of Done.

1. **Ownership is authored directly at commit rather than remaining staged for ordinary relationship persistence**

   **Status: Disagree.**

   The PR replaces the prior staged-relationship approach with `persist_space_membership` after `PublishRoot` succeeds.

   This changes ownership from staged lifecycle state into a special persistence side effect. Restore the prior approach: infer and add `OwnedBy` to the staged holon, then let ordinary relationship commit materialize `Owns`.

2. **`OwnedBy` is excluded from every ordinary relationship commit scope**

   **Status: Disagree.**

   The PR filters `OwnedBy` out of both full and touched relationship commit scopes. It states that any staged `OwnedBy`, including one inherited during clone, is dropped.

   Handle version semantics in `stage_new_version`: remove `OwnedBy` from the cloned relationship map. New lineages and independent clones retain staged ownership; versions never carry it into pass 2. No global commit-time filter is needed.

3. **The inverse pair is hand-authored instead of using ordinary relationship commit behavior**

   **Status: Disagree.**

   `persist_space_membership` directly constructs and persists both `OwnedBy` and `Owns` SmartLinks.

   Issue #642 calls for ordinary relationship commit behavior to persist `OwnedBy` and materialize `Owns`. The direct pair bypasses the normal relationship-commit and inverse-materialization path, hard-coding relationship names, direction, key selection, properties, and duplicate behavior.

4. **The partial-commit fixture asserts a new, unsupported intermediate-state contract**

   **Status: Disagree.**

   The new fixture requires a holon saved in pass 1 to appear in `GetAllHolons` immediately when another staged holon causes the commit response to be `Incomplete`.

   On `Incomplete`, the transaction remains open; saved staged holons retain their IDs and can be discovered through the nursery. A later complete commit can run ordinary relationship persistence for those committed staged holons. The fixture should instead assert retained recoverability after the incomplete response, then assert `OwnedBy`/`Owns` persistence after repair or abandonment of the failed sibling and a complete commit.

5. **`AllHolonNodes` retirement has not been implemented**

   **Status: Blocker.**

   The PR description states that `AllHolonNodes` is still written and its tests still pass. The changed-file set does not remove the module, write, reader surface, link type, integrity validation, probes, externs, constants, or index-specific tests required by Issue #642.

   Removing the complete `AllHolonNodes` surface is required by Issue #642. It must remain in this PR’s scope before the PR can claim to deliver or close the issue.

6. **Deletion cleanup is implemented through another hand-managed relationship path**

   **Status: Partially agree.**

   The required outcome is correct: deleted holons must not remain discoverable through `Owns`.

   However, `retract_space_membership` repeats the direct ownership-persistence exception and assumes all discovered `OwnedBy` links are space membership managed by this service. Prefer the corresponding lifecycle-aware relationship cleanup, with explicit ownership-scope semantics, rather than a second special direct-link protocol.

7. **Coverage does not demonstrate the intended lifecycle semantics**

   **Status: Partially agree.**

   The PR adds useful coverage for count changes, version non-duplication, deletion, and its proposed partial-commit behavior. It does not demonstrate the intended staged ownership lifecycle, independent-clone re-anchoring, or ordinary inverse materialization.

   Replace or supplement the partial-commit test with tests that pin the intended lifecycle:

   - A new holon receives context-derived `OwnedBy` in staged state.
   - An independent clone is re-anchored to the current transaction space.
   - A version omits `OwnedBy` from its cloned staged relationship map.
   - An incomplete commit retains saved staged identity without requiring finalized `Owns` visibility.
   - A subsequent complete commit materializes the normal forward and inverse relationship pair.

## Changes consistent with Issue #642

- `GetAllHolons` now resolves the current space and traverses `Owns`.
- The public command, wire, host dispatch, and SDK surface remain in place.
- The `LocalHolonSpace` anchor is excluded from `GetAllHolons`.
- Version-producing updates do not add membership for each historical version.
- Deleted holons are intended not to remain discoverable.
- No QueryExpression, QueryDance, or new public query API is introduced.

## Recommended revision

- Retain the retrieval migration and anchor contract.
- Revert `persist_space_membership` and restore staged ownership lifecycle behavior in the nursery.
- Handle clone re-anchoring and version ownership removal in their respective staging methods.
- Treat incomplete commit as an open transaction with recoverable nursery state, not as a completed whole-space-discovery state.
- Complete the `AllHolonNodes` retirement surface before claiming Issue #642.

nphias:
moved work to sub issues

PR #685, opened 2026-09-05
678 val pre dependency safe validation outcomes and shared objects default completion

### LINKED ISSUE #678: VAL-PRE: dependency-safe validation outcomes and shared-objects default completion
### 1. Summary (Required)

Establish the two foundations that descriptor-aware Commit Validation needs before any validator exists:

1. **Dependency-safe validation outcomes** — serializable, identity-only finding types in `core_types`, a findings collection on `StagedHolon`, a controlled operation that replaces validation state and findings together, and the matching wire projection.
2. **Construction completion at the Shared Objects layer** — invoke `populate_defaults()` from the transient and nursery create/clone paths once a governing descriptor is resolvable, and make an unresolved descriptor a non-fatal deferred outcome instead of a construction failure.

This issue adds no validation logic and changes no Commit decision. It is the data model and the completion guarantee that Capability 1 builds on.

Four concrete gaps block Capability 1:

**No place to record a semantic finding.** `StagedHolon` (`staged.rs:28`) carries `validation_state` and `errors: Vec<HolonError>`. `errors` is for operational commit failures, appended through `add_error()` (`staged.rs:289`). There is no collection for semantic findings, and mixing them into `errors` would conflate "this holon violated its contract" with "persistence failed" — a distinction the Commit response surface has to preserve.

**No way to set validation state.** `validation_state()` (`staged.rs:225`) is a getter. The only path that sets the field is `from_parts()` (`staged.rs:117`), used by wire binding. Commit validation needs to replace state and findings *together* so a subject can never be observed with new findings under a stale state.

**`HolonError` cannot carry validation vocabulary.** It is defined in `integrity_core_types/src/holon_error.rs:8`, the crate that must stay descriptor-independent for PVL. Descriptor-aware outcome types cannot live there, and a new `HolonError` variant is not an acceptable way to express a completion outcome.

**Completion effectively never runs.** `WritableHolon::populate_defaults()` (`writable_holon.rs:23`) exists and works, but its only caller is the loader (`holons_loader/src/controller.rs:323`). Holons created through `TransientHolonManager` or staged through `Nursery` are never completed. Two consequences:

- Commit requires completed explicit state and never supplies defaults, so any producer other than the loader presents incomplete holons to Capability 1's required-property check.
- `populate_defaults()` resolves the contract through `available_properties()` (`readable_holon.rs:209`) → `holon_descriptor()` (`readable_holon.rs:236`), which returns `Err(MissingDescribedBy)` when no descriptor is attached. Today that aborts the loader's whole completion pass into `LoadCommitStatus::Skipped`. During bootstrap, an unresolved descriptor is normal, not an error.

The scale is not marginal. `IsAbstractType` (DS §1.4) and `DefinesInstanceTypeKind` (DS §1.5) are both *required with default `false`*, and the corpus authors them only where they are `true`. `generated/json-imports/core/root.json` holds 38 holons against 12 `IsAbstractType` and 6 `DefinesInstanceTypeKind` mentions — several of which are the property declarations themselves. Without completion, most descriptor holons will fail `DS-PROP-001` the moment Capability 1 gates Commit.

- **Depends on VAL-RETIRE.** Default materialization exercises the value path, and `ValueDescriptor::is_valid()` currently errors against the real corpus: `StringValueDescriptor::is_valid()` → `resolve_string_constraints()` cannot classify `Length16k.StringLengthConstraint` under the retired one-bound model and returns `unsupported_constraint`. Debugging bootstrap convergence through that is avoidable by sequencing.
- **Blocks CAP1-A** and, transitively, CAP1-B.
- No dependency on VAL0 schema work beyond what VAL-RETIRE already lands.

Two independent workstreams, sequenced as separate commits.

#### 4.1 Dependency-safe validation outcomes

Add to `shared_crates/type_system/core_types/` — correct home because it already depends on `integrity_core_types` and sits below `holons_core`:

- `CommitValidationViolation { kind, rule_key, severity, subject, descriptor_identity, message }`
- `CommitValidationViolationKind`, `ValidationSubjectPath`, `ValidationSeverity`

All fields serializable and identity-only. No bound runtime reference may appear in a finding. `ValidationSeverity` is an independent dependency-safe mirror of the Core schema enum, not generated from it.  `CommitValidationReport` is deliberately not defined here. It never crosses a boundary — the host reads `CommitResponse` as a holon by property name (`map_commands_runtime/src/tests.rs:51`), and per-holon findings reach clients through the staged pool — so it needs neither serialization nor a dependency-direction home below `holons_core`. It is defined in `holons_validation` by VAL-C1a, alongside the validator that returns it.

Add to `holons_core`:

- a `validation_findings` collection on `StagedHolon`, distinct from `errors`
- a single controlled operation that replaces `validation_state` and `validation_findings` together and leaves `errors` untouched; extend `from_parts()` accordingly

Add to `holons_boundary`:

- a `validation_findings` field on `StagedHolonWire` (`src/context_binding/staged_wire.rs:12`), carried through `bind`, `rebind`, and `From<&StagedHolon>` alongside the existing `validation_state` and `errors`

Nothing reads these yet. The commit is purely additive.

#### 4.2 Construction completion at the Shared Objects layer

Change `populate_defaults()` to return an explicit outcome rather than treating an unresolved descriptor as failure. Suggested shape, keeping `integrity_core_types` untouched:

```rust
pub enum CompletionOutcome {
    Completed,
    DeferredNoDescriptor,
}

fn populate_defaults(&mut self) -> Result<CompletionOutcome, HolonError>
```

`HolonError::MissingDescribedBy` becomes `Ok(DeferredNoDescriptor)`; every other error still propagates. `MultipleDescribedBy` stays fatal — that is a real defect, not a deferral. There is precedent for this tolerance pattern in `definitional_equivalence.rs:305-312`.

Invoke completion where a descriptor can actually be resolved. Note that `TransientHolonManager::create_empty` (`transient_holon_manager.rs:130`) and `create_empty_without_key` (`:149`) construct holons with no `DescribedBy` at all — the descriptor is attached afterward — so those are *not* viable hook points. The viable ones are:

- `TransientHolonManager::new_from_clone_model` (`transient_holon_manager.rs:160`)
- `Nursery::stage_new_holon` (`nursery.rs:187`)
- `Nursery::stage_new_from_clone` (`nursery.rs:199`)
- `Nursery::stage_new_version` (`nursery.rs:224`)

`populate_default_if_required_and_absent` (`property_descriptor.rs:52`) is already idempotent — it skips populated properties and optional ones — so re-running completion is safe and the loader's existing pass becomes a bootstrap backstop rather than the only path.

Make the loader retry completion over its resolved staged import set once reference resolution has run. The retry must converge over a self-describing corpus: `MetaHolonType.MetaTypeDescriptor` is its own describing type (`schema-src/core/root.tdl:122`).

| Area                             | Change                                                                                                             |
| -------------------------------- | ------------------------------------------------------------------------------------------------------------------ |
| `core_types`                     | New violation types                                                                                         |
| `holons_core`                    | `StagedHolon` findings + controlled replacement; `populate_defaults` signature; transient/nursery completion hooks |
| `holons_boundary`                | `StagedHolonWire` findings field                                                                                   |
| `holons_loader`                  | Completion retry over the resolved staged set; outcome handling at `controller.rs:323`                             |
| Callers of `populate_defaults()` | Signature change ripples to the loader and its tests                                                               |

**Explicitly out of scope**, to keep the CAP1-A boundary unambiguous:

- `effective_relationship_targets()`, `effective_constraints()`, `effective_validation_bindings()`
- Making `equals_or_extends` / `effective_relationship_members` public — both are `pub(crate)` in `descriptors/inheritance.rs` and neither is re-exported from `descriptors/mod.rs`. They become public in CAP1-A, when there is a consumer.
- The `holons_validation` crate, any validator, any rule or constraint handler
- Any change to Commit's decision, ordering, or response surface
- Untangling `ValueDescriptor::is_valid()` — that is VAL-RETIRE

No Commit behavior changes here. Holons will carry more populated properties after staging, which is the intended effect.

Existing coverage is insufficient — `populate_defaults` has unit coverage only for its two failure modes (`property_descriptor.rs:183`, `:198`) and nothing for the success path outside the loader.

New unit tests in `holons_core`:

- completion populates a required-with-default property and leaves an optional one absent
- completion is idempotent: a second run over a completed holon changes nothing
- completion never overwrites an explicitly authored value
- an unattached descriptor yields `Ok(DeferredNoDescriptor)`, not an error
- multiple `DescribedBy` targets remain a hard error
- the controlled replacement operation replaces state and findings atomically and leaves `errors` untouched
- **clone coverage:** completion fills only omissions in a newly created independent staged clone and never retroactively alters persisted historical state

Round-trip tests in `holons_boundary`: `StagedHolon` → `StagedHolonWire` → `StagedHolon` preserves findings, keeps them distinct from `errors`, and tolerates the field's absence on deserialization.

Sweetests:

- **Bootstrap convergence** is the critical one. `load_core_schema_fixture`, `load_generated_core_schema_fixture`, and `load_generated_validation_schema_fixture` must complete with every staged descriptor holon carrying explicit `IsAbstractType` and `DefinesInstanceTypeKind` values, and the retry must terminate on the self-describing corpus.
- `load_holons_internal_fixture.rs` metrics assertions will need review if deferral changes `LoadCommitStatus` or holon counts.

- [ ] `CommitValidationViolation`, its kind/subject/severity types are defined in `core_types`, serializable, and free of bound runtime references.
- [ ] `StagedHolon` carries a `validation_findings` collection distinct from `errors`, with a single controlled operation that replaces state and findings together.
- [ ] `StagedHolonWire` carries findings through `bind`, `rebind`, and `From<&StagedHolon>`, and round-trips them.
- [ ] `integrity_core_types` is unchanged; no descriptor-aware or validation vocabulary was added to it.
- [ ] `populate_defaults()` returns an explicit completion outcome; an unresolved descriptor is non-fatal and multiple descriptors remain fatal.
- [ ] Completion runs on every shared-objects construction and independent-clone path. It is invoked directly from `TransientHolonManager::new_from_clone_model`, `Nursery::stage_new_holon`, and `SmartReference::clone_holon_impl`; `Nursery::stage_new_from_clone` and `Nursery::stage_new_version` inherit it through those calls rather than completing a second time.
- [ ] Saved clone models carry no relationships, so the saved-clone path completes only after the relationship map — including `DescribedBy` — has been attached. A completion write on that path is classified as an ordinary content mutation, so `Nursery::stage_new_version` escalates to `ForUpdateNewVersion` rather than letting a graph-only commit discard the materialized defaults.
- [ ] The loader no longer aborts to `LoadCommitStatus::Skipped` on a deferred descriptor.
- [ ] **Convergence criterion:** loading the canonical Core and Validation corpora leaves every staged descriptor holon with explicit `IsAbstractType` and `DefinesInstanceTypeKind` values, and the retry terminates on the self-describing corpus.
- [ ] Clone coverage proves completion fills only omissions in a new independent staged clone.
- [ ] `npm run check`, `npm run fmt:check`, `npm run test:unit`, and the schema-loading sweetests pass.

<details>

**Adding a `DefaultsDeferredNoDescriptor` variant to `HolonError`.** Rejected: `HolonError` lives in `integrity_core_types`, which must stay descriptor-independent for PVL. A completion outcome is also not an error, and modelling it as one keeps every caller treating a normal bootstrap state as a failure.

**Putting findings in the existing `errors: Vec<HolonError>`.** Rejected: the Commit response surface must distinguish semantic rejection from operational failure, and `add_error()` appends rather than replaces, which is wrong for a per-pass finding set.

**Completing at `create_empty` time.** Not viable — those constructors produce holons with no `DescribedBy` target, so there is no contract to complete against.

**Leaving completion to the loader only.** Rejected: it would make the loader a second acceptance gate in practice, since every non-loader producer would fail Capability 1's required-property check.

### 9. Risks or Concerns

**Bootstrap convergence is the real risk.** The corpus is self-describing, and completion depends on descriptor resolution that itself depends on holons in the same load set. If the retry does not converge, the loader either loops or leaves holons incomplete — and incomplete holons become Commit rejections the moment Capability 1 lands. This is why convergence is a named DoD criterion rather than an implicit expectation.

**Signature churn.** Changing `populate_defaults()`'s return type touches every caller. The blast radius is small today (one production caller plus tests), which argues for doing it now rather than after Capability 1 adds more.

**Ordering with VAL-RETIRE.** If VAL-PRE lands first, completion work will intermittently hit the broken `ValueDescriptor::is_valid()` path and burn time on a bug this issue does not own.

### 10. Additional Context

- Commit Validation Implementation Plan — VAL-PRE precursor
- Commit Validation Design Specification — §5.1 violation and report types, §5.4 state transition ownership
- Descriptor-Kernel Semantic Rules — §1.4, §1.5, §6.2 default materialization boundary
- Validation Architecture — §3.2 construction completeness

</details>

owleyeview:
Closes #678.

Adds the foundations for descriptor-aware Commit Validation: identity-only validation findings and default completion across shared-object construction and clone paths. No validator or new Commit acceptance gate is introduced.

- Defines serializable finding types in `core_types`, leaving `integrity_core_types` unchanged.
- Adds staged findings distinct from operational errors, with controlled replacement of validation state and findings together. Carries findings through wire binding, serialization, and the SDK contract.
- Makes missing descriptors a non-fatal completion deferral, preserves partial progress and authored values, and keeps multiple descriptors fatal.
- Uses the loader’s single post-resolution completion pass as the bootstrap retry. Undescribed nodes can now commit; relationship persistence still reports errors when required descriptor metadata is unavailable.

Corrects the issue’s assumption about saved clones: their clone models contain no relationships. `SmartReference` therefore completes defaults after attaching the relationship map, including `DescribedBy`. Version staging classifies any materialized defaults as content changes so a graph-only commit cannot discard them. Persisted historical state remains unchanged.

Coverage includes deferred retries, idempotency, clone isolation, lifecycle classification, findings round-trips, and explicit descriptor flags across the canonical Core and Validation bootstrap corpus.

Validation: `npm run check`, `npm run fmt:check`, 300 `holons_core` unit tests, and `npm run test` passed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1032,6 +1032,7 @@ dependencies = [
  "base_types",
  "integrity_core_types",
  "serde",
+ "serde_json",
  "uuid 1.21.0",
 ]
 
--- a/happ/crates/holons_loader/src/controller.rs
+++ b/happ/crates/holons_loader/src/controller.rs
@@ -315,14 +315,27 @@ impl HolonLoaderController {
         }
 
         // ─────────────────────────────────────────────────────────────────────
-        // DEFAULT POPULATION: complete only the exact nursery-staged set.
+        // DEFAULT POPULATION: retry construction completion after Pass-2 resolution
+        // over the exact nursery-staged set.
         // ─────────────────────────────────────────────────────────────────────
         let mut population_errors = Vec::new();
+        let mut deferred_count = 0;
         for mut staged_holon in context.staged_references()? {
             let source_loader_key = staged_holon.key()?;
-            if let Err(error) = staged_holon.populate_defaults() {
-                population_errors.push(ErrorWithContext { error, source_loader_key });
+            match staged_holon.populate_defaults() {
+                Ok(CompletionOutcome::Completed) => {}
+                Ok(CompletionOutcome::DeferredNoDescriptor) => deferred_count += 1,
+                Err(error) => {
+                    population_errors.push(ErrorWithContext { error, source_loader_key });
+                }
             }
+        }
+
+        if deferred_count > 0 {
+            warn!(
+                "Default population deferred for {} staged holon(s) with unresolved descriptors",
+                deferred_count
+            );
         }
 
         if !population_errors.is_empty() {
--- a/host/map-sdk/src/internal/wire-types/references.ts
+++ b/host/map-sdk/src/internal/wire-types/references.ts
@@ -130,6 +130,33 @@ export type ValidationState =
   | 'Validated'
   | 'Invalid';
 
+/** Identity-only semantic findings, matching Rust's default serde enum encoding. */
+export type CommitValidationViolationKindWire =
+  | 'NoDescriptor'
+  | 'UnsupportedValidationRule'
+  | { UnsupportedConstraintType: { constraint_identity: string; constraint_type_identity: string } }
+  | { RuleViolation: { code: string } }
+  | 'UnresolvedLocalDependency'
+  | 'RelationshipCoordinationRequired';
+
+export type ValidationSubjectPathWire =
+  | { Holon: { holon_identity: string } }
+  | { Property: { holon_identity: string; name: string } }
+  | { Value: { holon_identity: string; property: string } }
+  | { Relationship: { source_identity: string; name: string; target_identity: string } }
+  | 'Transaction';
+
+export type ValidationSeverityWire = 'Info' | 'Warning' | 'Error';
+
+export interface CommitValidationViolationWire {
+  kind: CommitValidationViolationKindWire;
+  rule_key: string | null;
+  severity: ValidationSeverityWire;
+  subject: ValidationSubjectPathWire;
+  descriptor_identity: string | null;
+  message: string;
+}
+
 // `Committed` is the only tuple-like staged payload variant used here.
 export type StagedState =
   | 'Abandoned'
@@ -161,6 +188,7 @@ export interface StagedHolonWire {
   holon_state: HolonState;
   staged_state: StagedState;
   validation_state: ValidationState;
+  validation_findings?: CommitValidationViolationWire[];
   property_map: PropertyMap;
   staged_relationships: StagedRelationshipMapWire;
   original_id: LocalId | null;
@@ -1158,6 +1186,9 @@ export function isStagedHolonWire(value: unknown): value is StagedHolonWire {
     isHolonState(value['holon_state']) &&
     isStagedState(value['staged_state']) &&
     isValidationState(value['validation_state']) &&
+    (value['validation_findings'] === undefined ||
+      (Array.isArray(value['validation_findings']) &&
+        value['validation_findings'].every(isCommitValidationViolationWire))) &&
     isPropertyMap(value['property_map']) &&
     isStagedRelationshipMapWire(value['staged_relationships']) &&
     isNullable(value['original_id'], isLocalId) &&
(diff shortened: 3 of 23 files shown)


evomimic (approved):
## Approve

Approved for Issue #678.

This PR adds the intended dependency-safe validation-finding surface and default-completion behavior without introducing validation logic or changing Commit acceptance. Findings remain separate from operational errors, round-trip through the staged wire and SDK contract, and construction completion correctly defers when descriptors are unresolved while preserving fatal multiple-descriptor errors.

The saved-clone path attaches relationships before completion and classifies materialized defaults as content changes, preserving historical state and avoiding graph-only loss. Loader coverage verifies canonical Core and Validation descriptor defaults converge.

CI is green on the latest head (`5243147`). Manual verification also passed: `build:happ`, `build:host`, `npm test`, and `npm start`.

Issue #719, opened 2026-09-16
Complete declared-only Commit verification and schema alignment after #720


owleyeview:
### 1. Summary

The primary behavior originally proposed here shipped in **#717 / merged PR #720**, merge commit [`90ec58bf2e18f65b913cd2badea2c5aa4fcacb8c`](https://github.com/evomimic/map-holons/commit/90ec58bf2e18f65b913cd2badea2c5aa4fcacb8c), including the final staged-update correction changes at `0d0b2a84`.

Keep #719 open for bounded verification and schema/documentation alignment. Do not reimplement the delivered enforcement, cloning, correction, or wire-scope mechanisms.

This body supersedes the original implementation proposal and the earlier scope comments. In particular, staged-update correction is no longer an outstanding implementation gap.

### 2. Delivered by PR #720

- **Common pre-write enforcement:** Commit assesses nonempty authored relationship collections on every live candidate, independently of touched-name persistence scope. Each name must resolve through the source descriptor's effective declared relationship contract. Inverse and unknown names are rejected before candidate persistence.
- **Finding contract:** `RuleViolation { code: "UndeclaredRelationship" }` uses the existing relationship subject with source identity, relationship name, and target identity. Findings combine with conformance results before outcome installation. Missing/multiple descriptors retain `NoDescriptor`; operational assessment failures do not install partial assessment outcomes.
- **Loader behavior:** ungoverned assembly preserves input while the graph is incomplete. Semantic invalidity is reported by common Commit as `Rejected`, rather than a loader-specific inverse-resolution `Skipped` failure. Operational loader and persistence failures remain separate.
- **Phase-specific clones:** destination `TransactionContext::clone_holon(&source)` owns transient allocation; `HolonReference` delegates clone-model construction to its variants. Saved sources require a descriptor and copy effective declared relationships, both definitional and non-definitional, excluding inverses. Transient/staged clones preserve current input without descriptor-dependent filtering. Their reads use the local relationship map, without saved-inverse fallback.
- **Retry scope transport:** required `relationship_commit_scope` carries `Full` or `TouchedOnly` through staged projection, binding/rebinding, and runtime restoration. Missing scope is not defaulted. SDK types and guards require the field.
- **Governed correction:** invalid local relationship input can be removed from creates and updates without permitting undeclared additions or introducing persisted inverse deletion. `RelationshipMutationEffect` makes lifecycle effects explicit; cleanup preserves graph-only and version-producing update state.
[...]

Issue #41, opened 2024-02-21
Introduce SmartReferences


evomimic:
SmartReferences are read-only references to cached holons.

```
pub struct SmartReference {
    holon_space_id: Option<SpaceId>    
    holon_id: HolonId,
    key : Option<MapString>,
    rc_holon: Option<Rc<Holon>>,
    smart_property_values: Option<PropertyMap>,
}

```

SmartReference should implement the following method:
`ensure_rc(&mut self, context: &HolonsContext)->Result<(), HolonError> `
 This is a private function that attempts to ensure that the SmartReference contains a populated `rc_holon` field.
If `rc_holon` is already populated, it simply returns Ok(self)
Otherwise, invoke `get_rc_holon `on the `cache_manager` (found in the context) to get a reference to the cached holon, update self's `rc_holon` field with the returned Rc and return Ok(self)
Else return HolonNotFound error
    

SmartReference should implement the following trait:

```
pub trait HolonFieldGettable {
    fn get_property_value(&self, context: HolonsContext, property_name: &PropertyName) -> Result<PropertyValue, HolonError>;
    fn get_property_names(&self, context: HolonsContext) -> Result<Vec<&PropertyName>, HolonError>;
    fn get_relationship_map(&self, context: HolonsContext) -> Result<&RelationshipMap, HolonError>;
    fn get_key(&self, context: HolonsContext)->Result<Option<MapString>,HolonError>;
}

```

### Implementations

` fn get_property_value(&self, context: HolonsContext, property_name: &PropertyName) -> Result<&PropertyValue, HolonError>;`

- First check `smart_property_values` to see if a value for the specified property_name is available. If so, return that value.
- Otherwise, invoke `ensure_rc` on self to be sure the rc_holon field is populated. Then use the rc_holon field to get the requested property value by property_name

`fn get_relationship_map(&self, context: HolonsContext) -> Result<&RelationshipMap, HolonError>;`
- invoke `ensure_rc` on self to be sure the `rc_holon` field is populated. Then use the `rc_holon` field to get the `relationship_map`

`fn get_key(&self, context: HolonsContext)->Result<Option<MapString>,HolonError>;`
- invoke `ensure_rc` on self to be sure the `rc_holon` field is populated. Then use the `rc_holon` field to get the holon's key

PR #720, opened 2026-09-17
Make SmartReferences space-bound instead of transaction-bound

### LINKED ISSUE #717: Make SmartReferences Space-Bound Instead of Transaction-Bound
### 1. Summary (Required)

**What is the enhancement?**  
Make `SmartReference` a space-bound, self-resolving reference to immutable saved holons rather than a transaction-bound reference. Introduce a narrow read capability for the owning `HolonSpaceManager`, preserve `HolonCollection` as a collection of bound references, and move clone allocation from `ReadableHolon` to the destination transaction.

This is a foundational Reference Layer enhancement. It corrects the ownership boundary exposed when the per-space `RelationshipCache` returned `SmartReference` values originally minted for a different transaction, producing `CrossTransactionReference` during Canvas initialization.

---

**Why is this needed?**  
Saved holons and their saved-to-saved relationship topology are immutable. `HolonSpaceManager` owns one `HolonCacheManager` and therefore one `RelationshipCache` per space. However, `SmartReference` currently retains a `TransactionContextHandle`, and `SmartReferenceWire` serializes a `TxId`. Consequently, a cached `RelationshipMap` / `HolonCollection` can retain Smart references bound to the transaction that first populated it and later return them to another transaction.

The resulting `CrossTransactionReference` is not a cache-keying problem: it reveals that saved-reference scope is modeled incorrectly. Transaction ownership is semantically necessary for transient and staged references because they resolve through a transaction's `TransientHolonManager` or `Nursery`. It is not necessary for read-only saved references, whose backing cache/router and immutable saved relationship data are space-scoped.

The existing Reference Layer is intended to give ordinary callers one self-resolving `HolonReference` surface while hiding whether a holon is transient, staged, or saved. Binding a saved reference to the space preserves that goal; requiring callers to supply a transaction to ordinary saved reads would not.

---

- This issue is an architectural prerequisite for #706, **Properties Pane and Descriptor-Driven Value Presentation**, because the current cross-transaction relationship-cache failure prevents reliable Canvas initialization.
- #716, **Extract DAHN selection into a host-native peer crate**, is already implemented as a checkpoint and is independent of this Reference Layer work.
- This issue is newly grounded architectural work; it is not presently a separately named PR unit in `docs/roadmap/desc-driven-impl-plan.md`. Its design authority is the Reference Layer / Shared Objects architecture and saved-holon immutability rules described below.

---

**How would you solve it?**  

1. Define a narrow, runtime-only, opaque `SpaceReadHandle` owned and minted by `HolonSpaceManager`. It carries a stable runtime read-scope identity plus only the saved-state services required by `SmartReference`. It is not a transaction, a distributed snapshot, or a mutable `HolonSpaceManager` facade.
2. Give `SpaceReadHandle` the crate-internal saved-read surface:
   - `read_saved_holon(holon_id)`;
   - `read_saved_relationships(source_holon_id, relationship_name)`; and
   - `read_all_saved_relationships(source_holon_id)`.
   These operations route through the space cache/router and saved-holon service. The named and all-relationship operations must reuse/populate one coherent per-space relationship cache rather than returning independently fetched maps.
3. Replace `SmartReference`'s `TransactionContextHandle` with `SpaceReadHandle`. Saved property, model, access, and relationship reads remain parameterless and self-resolving. Smart-reference identity/equality becomes `(space_read_handle_identity, holon_id)`; cached SmartLink property hints remain non-identity metadata.
4. Retain `TransactionContextHandle` exclusively for `TransientReference` and `StagedReference` resolution and mutation.
5. Update Smart-reference factories, client and guest holon-service materialization, context binding, equality/identity, and `SmartReferenceWire` so a saved reference no longer carries or validates transaction identity. A Smart wire is an unbound saved-reference locator containing only holon identity and optional SmartLink/property hints; ingress binds it to the receiving space. A local ID has no originating space qualifier; projection does not detect foreign-space mismatches and is not a cross-space transport mechanism. Remove the transaction `rebind` workaround.
6. Preserve the existing per-space `RelationshipCache` representation: `RelationshipCache` → `RelationshipMap` → `HolonCollection` → `SmartReference`. Do not introduce a parallel `CachedRelationshipMembers` or SmartReference-like projection. On cache insertion, validate Smart-only membership and seal collections as `CollectionState::Saved`, so the existing collection API rejects later mutation.
7. Remove `clone_holon()` from the general `ReadableHolon` contract. Cloning creates a transient that is owned by the destination `TransactionContext` (for example, `TransactionContext::clone_holon(&source)`). The supplied `HolonReference` is responsible for resolving its own source state and producing a normalized `HolonCloneModel`; source and destination may belong to different transactions. The clone operation imposes no space-equality check, but this does not establish supported cross-space transport or direct foreign-space references. The resulting transient belongs exclusively to the destination transaction. Migrate command, nursery, loader, descriptor, and test callers. Source accessibility and phase-specific clone-model requirements apply; source/destination transaction equality is not required. Multi-space operation is not currently supported. Under the speculative mirror design, externally owned holons would be represented by local mirrors in the receiving space; cloning mirrors remains unspecified.
8. Lift transaction-local descriptor interpretation out of `SmartReference`. Ordinary Smart reads resolve immutable saved state through `SpaceReadHandle`; descriptor/validation operations that intentionally use a transaction's staged schema overlay must be explicitly anchored in `TransactionContext` or a transaction-owned descriptor facade. Replace the present universal `HolonReference::tx_id()` / `bound_context()` assumptions with phase-aware execution-scope compatibility checks.
9. Preserve the distinction between governed authoring and raw staging. Ordinary programmatic `add_related_holons` authoring uses the source's effective declared relationship contract and rejects undeclared or inverse relationship writes. Loader assembly may use ungoverned staged writes while the import's descriptor graph is incomplete; this preserves imported input for validation rather than authorizing inverse persistence.
10. Before any candidate node or SmartLink is written, Commit validates the complete live candidate set against the completed descriptor graph. Every nonempty authored relationship collection must be licensed by the source's effective declared relationship contract. Inverse and unknown relationship names produce blocking `RuleViolation { code: "UndeclaredRelationship" }` findings, with relationship subjects identifying the source, relationship name, and target. No target-contract lookup or inverse-index inference is used to authorize authored occurrences. Missing or ambiguous descriptors retain the `NoDescriptor` finding; empty collections are ignored.
11. Semantic rejection preserves live staged candidates for correction and retry. The offending candidate is `Invalid`, unaffected candidates are assessed independently, and actionable findings survive guest → wire → client projection. Semantic rejection is distinct from operational loader or persistence failure; it does not install persistence errors. `IndependentlyAuthoredInverseRelationship` is superseded by the source-contract-based `UndeclaredRelationship` rule.

Non-goals:

- Do not split or redesign `HolonCollection` / `HolonCollectionApi` in this issue. A collection's effective scope follows its contained references; the collection itself does not retain a transaction handle.
- Do not introduce a saved-relationship topology DTO or a second cache representation.
- Do not change the saved-relationship rule: saved holons may relate only to saved holons. Transient and staged relationship rules remain transaction-local.
- Do not add a space shutdown lifecycle. A future lifecycle design may define `SpaceClosed`, but no such operation is currently exposed.
- Do not promise globally serializable DHT snapshot semantics. Direct reads of a saved `HolonId` and its saved relationship topology are stable because they are immutable; changing logical-head or query selection is a separate observed-view/query concern.

---

**What does this impact?**  

- `shared_crates/holons_core` Reference Layer: `SmartReference`, `HolonReference`, `ReadableHolon`, clone implementation dispatch, factories, and identity semantics.
- Space/read infrastructure: `HolonSpaceManager`, cache access/routing, and the new narrow read handle. The handle type may be passed across crate boundaries for binding, but its fields, constructor, and raw `read_saved_*` methods remain opaque/crate-internal; callers use `SmartReference` rather than raw cache access.
- Relationship access: `RelationshipCache` must safely share cached `RelationshipMap` / `HolonCollection` values across transactions once their Smart-reference members are space-bound.
- Boundary binding: `shared_crates/holons_boundary/src/context_binding/smart_reference_wire.rs` and all ingress/egress consumers.
- Host and guest holon services that currently mint Smart references from a transaction context, including `holons_client`, `holochain_receptor`, and `holons_guest`.
- Clone consumers, including the Commands contract/runtime, Nursery, loader, descriptor helpers, fixtures, and integration tests.
- Query's planned bound-first posture: the roadmap explicitly identifies `HolonReference`, `BoundHolonCollection`, and contract-significant `SmartReference` as primary runtime shared types. This change aligns saved-reference ownership with that posture without changing query result contracts in this issue.

---

- Add focused Rust tests proving that a `RelationshipCache` populated through transaction A can be read through transaction B in the same space without `CrossTransactionReference`.
- Verify cached relationship members remain ordered and preserve keys and SmartLink-derived property hints.
- Verify saved-reference reads resolve through the owning space and cannot resolve through an unrelated space.
- Verify transient and staged references remain bound to, and reject use outside of, their owning transaction.
- Add coverage for Smart-reference wire projection/binding without a serialized `TxId` and for expected ingress binding to the receiving space.
- Add clone tests for destination-owned allocation across transactions within the supported single-space execution model. Sources resolve through their own bound handles; no source/destination transaction equality is required. The resulting clone remains transient and belongs to the destination transaction. Ordinary transient/staged resolution and mutation remain bound to their owning transaction.
- Verify phase-specific clone construction: `HolonReference` delegates to its variants; transient/staged sources preserve their current authored relationship state without requiring a descriptor; saved sources require a descriptor and include both definitional and non-definitional effective declared relationships, excluding inverses.
- Verify governed authoring rejects undeclared/inverse writes while raw loader assembly can stage such input for Commit validation.
- Verify inverse-authored and unknown-relationship imports return semantic rejection before any node or SmartLink writes. Assert the offending source remains live and `Invalid`, carries `RuleViolation { code: "UndeclaredRelationship" }`, and retains exact source/name/target identities through guest → wire → client projection. Assert unaffected candidates remain `Validated` without findings when otherwise valid.
- Preserve separate operational loader/persistence failure coverage and verify corrected staged input can be retried successfully.
- Add descriptor tests proving that a staged schema override affects only validation explicitly performed through that transaction, not ordinary reads from a Smart reference that happened to be minted there.
- Run affected shared-crate, host, hApp/WASM, and Sweettest suites, including Canvas/startup coverage that previously surfaced the launcher failure.

---

- [ ] A narrow, opaque `SpaceReadHandle` exists, is minted only by `HolonSpaceManager`, and is the sole bound runtime dependency of `SmartReference`; it exposes only crate-internal saved-holon and saved-relationship reads, not mutable manager lifecycle or transaction-management operations.
- [ ] `SmartReference` no longer stores, compares, serializes, validates, or dereferences through a `TransactionContextHandle` / `TxId`; equality is same-space plus same saved holon identity.
- [ ] `SmartReferenceWire` is bound at the space ingress boundary and no longer uses transaction identity to reject an otherwise valid saved reference.
- [ ] Saved-holon property/model/relationship reads remain parameterless on `SmartReference` and resolve against its owning space.
- [ ] `RelationshipCache` safely reuses existing cached `RelationshipMap` / `HolonCollection` values across transactions in one space, has a regression test covering the former cross-transaction failure, and uses one coherent cache for named and all-relationship reads.
- [ ] The cache validates Smart-only members and seals cached relationship collections as `CollectionState::Saved`, preventing collection mutation through its existing API.
- [ ] `ReadableHolon` no longer offers clone allocation; clone creation is transaction-owned and all production callers are migrated.
- [ ] Descriptor/validation behavior that requires a staged schema overlay obtains it explicitly from its transaction; `SmartReference` no longer supplies a transaction context, and `HolonReference` execution-scope checks are phase-aware.
- [ ] Transient and staged reference transaction isolation remains covered by tests and unchanged in behavior.
- [ ] Raw loader staging preserves incomplete authored input; ordinary governed authoring still rejects undeclared/inverse relationship writes.
- [ ] Commit rejects the entire candidate set before node or SmartLink persistence when populated authored relationships are absent from the source's effective declared contract, using `RuleViolation { code: "UndeclaredRelationship" }` without target-contract inference.
- [ ] Semantic rejection preserves staged lifecycle and projected actionable findings with exact relationship source/name/target identities, assesses unaffected candidates independently, and supports correction and retry. Operational failures remain separately covered.
- [ ] Relevant formatting, type/build, unit, host, hApp/WASM, and integration checks pass.

<details>

- **Key the relationship cache by `(TxId, HolonId)`: rejected.** This would mask the scope error, duplicate immutable saved topology per transaction, and defeat the intended per-space cache ownership.
- **Cache only target ids/hints in a new `CachedRelationshipMembers` projection: rejected for this issue.** Once Smart references are space-bound, the existing `RelationshipMap` and `HolonCollection` are already safe to cache and preserve ordering, keys, and hints without a duplicate representation.
- **Keep Smart references transaction-bound and require explicit transaction context for reads: rejected.** It weakens the self-resolving Reference Layer and wrongly assigns transaction ownership to immutable saved state.
- **Refactor `HolonCollectionApi` now: deferred.** Its relationship/transient conflation is known design debt, but it is not required to make a collection of space-bound Smart references safe in the relationship cache.

---

### 9. Risks or Concerns

**What could go wrong?**

- This changes a foundational runtime reference contract and has host, guest, wire, command, loader, descriptor, and test ripple effects.
- The implementation must verify that `TransactionContext` does not supply unmodeled authorization semantics for saved reads. It cannot by itself provide a globally serializable Holochain DHT snapshot; any future query/read-view consistency contract must be explicit rather than retained implicitly as Smart-reference binding.
- Moving clone allocation must preserve source-bound resolution and destination-owned allocation without reintroducing context parameters on ordinary reference operations. Absence of a space-equality check must not be presented as support for cross-space transport, foreign-local wire locators, or mirror cloning.
- This issue must preserve the Reference Layer's invariant that runtime reference handles are not serialized directly; only wire data crosses transport boundaries and is bound at ingress.

---

### 10. Additional Context

**Authoritative design/context sources**

- `map-dev-docs/docs/core/architecture/holon-layered-representation-design-spec.md` — Reference Layer owns phase-aware, uniform access handles; Shared Objects owns lifecycle distinctions.
- `map-dev-docs/docs/core/rust-api.md` — Smart references are read-only views of saved/cache-backed holons; transient and staged references are the writable phases.
- `map-dev-docs/docs/core/transactions/relationship-persistence-design-spec.md` — relationship persistence and lifecycle constraints.
- `map-dev-docs/docs/core/transactions/transactions-design-spec.md` — reference capabilities, phase-specific cloning, raw staging, Commit candidate validation, and cache-read lifecycle.
- `map-dev-docs/docs/roadmap/desc-driven-impl-plan.md` — Wave 4 identifies the primary bound posture around `HolonReference`, `BoundHolonCollection`, and contract-significant `SmartReference`.

**Repository findings**

- `shared_crates/holons_core/src/core_shared_objects/space_manager.rs` constructs one `HolonCacheManager` / cache router per `HolonSpaceManager`.
- `shared_crates/holons_core/src/core_shared_objects/relationship_cache.rs` caches `RelationshipMap` values by source `HolonId` and currently relies on saved-holon immutability.
- `shared_crates/holons_core/src/reference_layer/smart_reference.rs` currently stores `TransactionContextHandle` and uses it for cache access, relationship reads, and transient clone allocation.
- `shared_crates/holons_boundary/src/context_binding/smart_reference_wire.rs` currently embeds `TxId`, validates it on bind, and offers transaction rebinding as a workaround for stale Smart references.

</details>

evomimic:
## Summary

Implements #717 by moving saved SmartReferences from transaction binding to an immutable SpaceReadHandle, while retaining transaction ownership for transient and staged references.

## Changes

- Adds the space-bound read handle and makes SmartReferences resolve saved holons and relationships through it.
- Moves clone creation to TransactionContext, preserving transaction-local transient ownership.
- Limits host cross-transaction relationship caching to declared, definitional source relationships; inverse and non-definitional memberships remain fresh.
- Establishes the source-side relationship contract needed by the cache policy and clarifies read-navigation versus authored-relationship semantics.
- Implements common pre-write Commit rejection of populated undeclared relationships, phase-specific cloning, and loader/projected-finding coverage from #719 under the revised source-contract-based design.
- Updates wire binding, host/guest services, SDK fixtures, and Sweettest coverage for the changed reference model.

## Validation

- npm run build:happ
- npm run build:host
- npm test
- npm run start:loader
- npm start
- Pre-push formatting checks

## Follow-up

#719 remains open for remaining correction, verification, and documentation work described in its scope-reconciliation comment. Target-side occurrence-orientation inference is superseded, not deferred implementation work.

## Clone and transport scope clarification

Clone allocation belongs to the receiving transaction. Cross-transaction cloning is supported. Absence of a source/destination space-equality check does not define cross-space transport or permit direct foreign-space references. Multi-space operation is not currently supported; the speculative mirror design would represent externally owned holons through local mirrors, whose cloning semantics remain unspecified.

Copied relationship references retain their runtime bindings. `SmartReferenceWire` omits that binding and interprets a local ID in the receiving space. It is not a space-qualified locator, and projection currently performs no foreign-space rejection check. These limitations are documented rather than presented as implemented multi-space support.


owleyeview (changes requested):
## Review Findings

Reviewed head `1e7602fdc242b21ef3981a195ff5801136882c1d`. CI is green. The primary #717 direction is sound, and this PR implements substantially more of #719 than the “narrow prerequisite slice” description suggests. I recommend addressing the following before merge.

### 1. [P1] Enforce the same-space boundary for saved clones

`TransactionContext::clone_holon` checks ownership only when `source.transaction_context()` returns `Some`. Smart references return `None`, so a saved source from another space bypasses the check.

#717 explicitly requires a Smart clone source to belong to the destination transaction’s space. Check this before reading the source or allocating the transient clone.

Please add tests proving:

- Same-space, different-transaction saved cloning succeeds.
- Different-space saved cloning is rejected.
- Transient/staged sources retain exact-transaction isolation.

Source: [transaction_context.rs:374](https://github.com/evomimic/map-holons/blob/1e7602fdc242b21ef3981a195ff5801136882c1d/shared_crates/holons_core/src/core_shared_objects/transactions/transaction_context.rs#L374)

### 2. [P2] Restricted read transactions leave accumulating registry entries

`open_restricted_cache_read_transaction` registers a weak reference for every internal context. Dropping the context does not remove that entry; cleanup requires a subsequent lookup of that particular expired transaction ID.

Relationship reads create these contexts even on cache hits, and descriptor traversal creates additional contexts. A long-lived host therefore accumulates dead transaction-registry entries.

Provide lifecycle cleanup for these internal contexts, including descriptor-resolution contexts, and a repeated-read regression proving registry size remains bounded.

Source: [transaction_manager.rs:96](https://github.com/evomimic/map-holons/blob/1e7602fdc242b21ef3981a195ff5801136882c1d/shared_crates/holons_core/src/core_shared_objects/transactions/transaction_manager.rs#L96)

### 3. [P2] All-relationship reads introduce redundant fetches and retain a cache-coherence gap

The host `get_all_related_holons` path reads each declared collection individually, then unconditionally fetches the full relationship map. On a cold cache this is N named service calls plus one full fetch, excluding descriptor-resolution work. Non-definitional collections incur the extra named fetch even on subsequent calls.

Saved cloning uses this path, so the amplification affects staging as well as reads. Prefer one bulk fetch that populates/reuses eligible cached collections while retaining fresh inverse/non-definitional membership.

Separately, the request-local branch returns the raw full fetch without sharing the named-read cache. This does not satisfy #717’s explicit requirement that named and all-relationship reads use one coherent cache.

Add service-call-count and named→all/all→named tests, including saved collection sealing and cross-transaction reuse where applicable.

Source: [holon_cache_manager.rs:205](https://github.com/evomimic/map-holons/blob/1e7602fdc242b21ef3981a195ff5801136882c1d/shared_crates/holons_core/src/core_shared_objects/holon_cache_manager.rs#L205)

### 4. [P2] Restore integration assertions for returned staged findings

The inverse-loader fixture removes its staged `validation_state()` and `validation_findings()` assertions rather than replacing them with the new rejection contract. The inverse branch in `commit_conflict_tests` likewise stops checking validation state.

Response-level `ValidationViolationCount == 1` does not prove that the offending source retains its actionable finding after guest → wire → client projection.

Assert that the offending source remains live, is `Invalid`, and carries `IndependentlyAuthoredInverseRelationship` with a relationship subject identifying the source, name, and target. Check unaffected candidates separately rather than assuming every staged holon is invalid.

Also preserve operational-loader-failure coverage alongside semantic rejection; replacing the operational helper should not erase that distinction.

Source: [inverse loader executor](https://github.com/evomimic/map-holons/blob/1e7602fdc242b21ef3981a195ff5801136882c1d/tests/sweetests/tests/execution_steps/load_book_person_inverse_test_schema_executor.rs)

### Verification

CI passes on the reviewed head. The PR description reports successful manual build/test/start checks; I have not independently rerun those or Sweettests. The findings above are based on source and test-diff inspection.

evomimic:
Thanks for the careful review. The findings helped expose a few places where the implementation had gotten ahead of the settled semantics. Here is the intended disposition of each item.

## 1. Clone source/destination boundary

The proposed-solution text in #717 that required same-space Smart sources and same-transaction staged/transient sources was incorrect and has been corrected. **_The source can therefore be from any space or transaction._** 

A clone is a read of source state plus creation of a new transient in the destination `TransactionContext`. The destination context owns allocation and subsequent lifecycle of that transient; the source reference resolves its own state through its bound mechanism:

- `SmartReference` resolves saved state through its `SpaceReadHandle`.
- `StagedReference` and `TransientReference` resolve transaction-local state through their bound transaction context.

`TransactionContext::clone_holon(&source)` owns destination allocation. `HolonReference` delegates clone-model construction to its variants. Transient and staged references copy their current relationship state, with downstream validation enforcing persistence constraints. `SmartReference` requires a described source and copies only effective declared relationships, excluding inverse collections. Both definitional and non-definitional declared relationships are included. Clone inclusion and cache eligibility are separate policies.

## 2. Restricted read-transaction registry lifecycle

Agreed on the registry-growth concern. 

Addressed by separating public transactions from private restricted read frames. Public transactions remain registered; internal cache-read and descriptor-resolution frames are never registered. Dropping them therefore leaves no registry entry requiring later cleanup.

The regression opens and drops 32 private restricted read frames, verifies they cannot be looked up through the public registry, and confirms the registry remains empty. This tests frame lifecycle directly rather than exercising repeated relationship reads end to end.

## 3. All-relationship reads, redundant retrieval, and cache coherence

Agreed on the problem. 

The updated flow is now:

    get_all_related_holons(source):
      enumerate the source's available outbound relationship names
      call get_related_holons(source, name) for each
      assemble the returned collections into a RelationshipMap

Named reads check the cache before resolving descriptors. A hit returns the sealed collection immediately. On a miss, the collection is fetched and sealed, then its eligibility is determined before insertion. Descriptor-resolution guards do not prevent reuse of existing cache entries.

The policy applies to both request-local and space-scoped caches:

- Definitional declared relationships are eligible for reuse.
- Non-definitional declared relationships are read fresh.
- Inverse relationships are read fresh.

Regression coverage checks named/all ordering, cross-transaction reuse, fresh non-definitional and inverse membership, collection sealing, known-empty collection reuse, and classification-free cache hits.


## 4. Returned staged findings for directly authored inverse relationships

The integration coverage now checks projected staged findings in `commit_conflict_tests`: inverse-authored input returns `Rejected`, retains live staged candidates, marks the offending candidate `Invalid`, and leaves unaffected candidates `Validated`. It also verifies rejection occurs before any node or SmartLink writes.

The finding contract has changed deliberately. `IndependentlyAuthoredInverseRelationship` was removed; authored relationships absent from the source’s effective declared contract now produce `RuleViolation { code: "UndeclaredRelationship" }`. This rejects inverse authoring without target-contract inference.

Operational-failure coverage remains separate from semantic rejection, including persistence failure/retry coverage and duplicate-key loader failure coverage.

owleyeview (dismissed):
## Review Findings

Reviewed head `027e89d3191de0ae43f48296d3a767cc2ebd03ae`. CI is green, and startup verification is understood to have been completed.

The revision resolves the private-read registry leak, unifies named/all relationship cache behavior, and restores integration assertions for projected validation findings. I have two remaining correctness concerns.

### 1. [P2] Rejected update candidates cannot remove the offending relationship

`classify_relationship_removal` permits descriptor-independent removal only for `ForCreate`. Every update state still requires the relationship being removed to resolve as a declared relationship:

```rust
match effective_relationship_declaration(&source_ref, relationship_name.clone()) {
    Ok(relationship_descriptor) => Ok(Some(relationship_descriptor.is_definitional()?)),
    Err(error @ HolonError::DescriptorDeclarationNotFound { .. }) => Err(error),
    Err(error) => Err(error),
}
```

Consequently, an update containing undeclared input through ungoverned staging or wire restoration is rejected by Commit, but `remove_related_holons` cannot discard that same input. It returns `DescriptorDeclarationNotFound` before mutation.

The new correction test exercises `ForCreate`, so it does not cover this case.

Please provide a governed path for discarding invalid uncommitted relationship input on updates, without permitting new undeclared additions or introducing persisted inverse-deletion semantics. Cover:

- Graph-only update: reject → remove offending occurrence → successful retry.
- Version-producing update: the same sequence, without demoting its lifecycle state.
- Removing the last member leaves no Pass 2 descriptor-resolution failure.

Source: [staged_reference.rs:409](https://github.com/evomimic/map-holons/blob/027e89d3191de0ae43f48296d3a767cc2ebd03ae/shared_crates/holons_core/src/reference_layer/staged_reference.rs#L409)

### 2. [P2] Cross-space clone targets need a fail-closed transport boundary

Cross-space cloning is now explicitly supported by the revised contract. The remaining problem is preserving the identity of copied relationship targets across transport.

A saved clone can retain a Smart relationship target bound to space A with `HolonId::Local(id)`. Projection writes only that local ID and property hints. Binding the clone in destination space B attaches B’s read handle to the same ID, losing the target’s originating space. This also applies to a copied `DescribedBy` reference.

The existing `require_local_target` check does not catch this: it rejects `External` IDs, but the incorrectly scoped ID remains `Local`.

This does not require implementing multi-space persistence now. Please fail closed when projection cannot safely represent a foreign local reference, or preserve a space-qualified locator. The check must occur while the originating scope is still available; the current wire payload does not contain enough information to detect the mismatch during binding.

A unit regression using two space managers is sufficient: verify that a foreign local target cannot silently become a destination-local target during projection/binding. No multi-space conductor harness is required.

Sources: [saved clone construction](https://github.com/evomimic/map-holons/blob/027e89d3191de0ae43f48296d3a767cc2ebd03ae/shared_crates/holons_core/src/reference_layer/smart_reference.rs#L387), [Smart wire projection/binding](https://github.com/evomimic/map-holons/blob/027e89d3191de0ae43f48296d3a767cc2ebd03ae/shared_crates/holons_boundary/src/context_binding/smart_reference_wire.rs#L22)

These findings are based on source-path inspection; I have not independently run local builds or Sweettests.

evomimic:
### Clarification of clone scope and finding 2

Our earlier statement that a clone source can be from “any space or transaction” overstated current support. Cross-transaction cloning and destination-owned allocation are supported. The absence of a space-equality check in `clone_holon` does not establish direct foreign-space reference transport.

Multi-space operation is not currently supported. Under the speculative mirror design, an externally owned holon would be represented in the receiving space by a local mirror; cloning such mirrors remains unspecified.

The wire limitation identified in finding 2 is real: projection omits the runtime space binding, and a local ID is interpreted in the receiving space. We have not added a foreign-space rejection check or a space-qualified locator. We are clarifying the contract and code documentation rather than claiming that cross-space transport is supported or that this boundary has been hardened. A supported application path carrying a foreign-local reference across that boundary has not been established in this review.

The issue and PR descriptions now reflect this scope. Matching transaction-spec edits and API doc comments are prepared locally. The finding 1 update-cleanup fix and its integration cases are also prepared locally; the full rebuild/test/start sequence is pending.

evomimic:
Thanks for the follow-up review. Commit `0d0b2a84` is pushed for re-review.

## 1. Rejected update candidates cannot remove the offending relationship

Fixed. Governed removal now permits discarding undeclared local relationship input when the source’s effective declared contract resolves successfully but does not contain that name. Other descriptor-resolution errors still propagate. Undeclared additions remain prohibited.

The documented internal `RelationshipMutationEffect` enum replaces the previous `Option<bool>` classification. Discarding invalid input preserves the existing update lifecycle; it does not authorize persisted inverse deletion or introduce saved-state fallback.

Regression coverage exercises:

- Graph-only update: reject → remove offending occurrences → successful retry without writing a new node.
- Version-producing update: the same sequence, preserving its lifecycle and writing a new node.
- Removal of the last invalid members, leaving empty collections that do not fail Pass 2 descriptor resolution.
- Continued rejection of undeclared additions and propagation of missing-descriptor errors.

## 2. Cross-space clone targets and transport

Our earlier wording overstated current support. Cross-transaction cloning and destination-owned allocation are supported. The absence of a space-equality check in `clone_holon` does not establish support for direct foreign-space references or cross-space transport.

Multi-space operation is not currently supported. Under the speculative mirror design, externally owned holons would be represented in the receiving space by local mirrors; cloning those mirrors remains unspecified.

The wire limitation you identified is real: projection omits the originating runtime space binding, and binding interprets a local ID in the receiving space. We have not added a foreign-space rejection check or a space-qualified locator. This is a contract clarification, not transport hardening.

The API comments, #717, and this PR’s description now state that limitation explicitly. The matching transaction-spec clarification has also landed on the docs repository’s `main` through [map-dev-docs PR #41](https://github.com/memetic-activation-platform/map-dev-docs/pull/41).

Given the clarified supported scope, we propose treating cross-space boundary hardening as future work rather than a merge blocker for this PR.

## Verification

Following a clean rebuild, hApp and host builds passed, `npm test` was nominal, and both application and loader startup checks were nominal. All 331 focused core tests passed, the integration changes compiled, and pre-push formatting checks passed.

